PR #1, opened 2022-02-04
Artnaseef

artnaseef:
Add DB setup + migrator (liquibase).  Add docker-it.


artnaseef:
I added a DEBUG step to the circleci build to list the contents of the .m2 folder during the test setup.

Note the change to the surefire plugin config in the parent pom (see the commented-out section).

platform/db/schema/pom.xml.old
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
j-white, line 1: Can delete

platform/db/schema/o
```diff
@@ -0,0 +1,18 @@
+[INFO] Scanning for projects...
```
j-white, line 1: Can delete

platform/pom.xml
```diff
@@ -129,23 +138,24 @@
...
-                            <excludes>
-                                <exclude>none</exclude>
-                            </excludes>
-                            <includes>
-                                <include>**/*IT.java</include>
-                            </includes>
-                        </configuration>
-                    </execution>
-                </executions>
+                <!-- TODO: errors using surefire plugin skip the post-integration-test phase -->
```
j-white, line 141: Should we change this to use failsafe by default for the ITs instead?

platform/db/schema-liquibase/pom.xml.old
```diff
@@ -0,0 +1,71 @@
...
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.opennms.core</groupId>
+      <artifactId>org.opennms.core.schema</artifactId>
+    </dependency>
+  </dependencies>
+
+  <repositories>
```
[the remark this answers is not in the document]
artnaseef: Oh pom.xml.old wasn't intended to be committed - my bad.

platform/pom.xml
artnaseef: Yeah, I think that is the best approach.  Failsafe does require a little more config to use.  Shall I look into a parent-pom config so projects with *IT.java just "work"?

artnaseef: OK, I added it to the parent pom and tested.  Looks good - all of the modules run the failsafe plugin, but those without tests just otherwise ignore it.

Changing docker-it to use the parent config looks good.

PR #3, opened 2022-02-09
Add jaas auth

artnaseef:
Adds JAAS Authentication and Roles validation (`@RolesAllowed`).
Adds core/rest bundle that adds both to the default Whiteboard.
Maps CXF Authentication failure to 403 instead of 500.

### DIFF
--- a/platform/alarms/itests/src/test/java/org/opennms/horizon/alarms/itests/CoreContainer.java
+++ b/platform/alarms/itests/src/test/java/org/opennms/horizon/alarms/itests/CoreContainer.java
@@ -60,7 +60,7 @@ public class CoreContainer extends GenericContainer<CoreContainer> implements Te
                 .withCommand(containerCommand)
                 .withFileSystemBind("target/test-classes/users.properties", "/opt/horizon-stream/etc/users.properties", BindMode.READ_ONLY)
                 .waitingFor(new HttpWaitStrategy()
-                        .forPath("/cxf/alarms/list")
+                        .forPath("/alarms/list")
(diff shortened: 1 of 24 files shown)


platform/docker-it/src/test/resources-filtered/karaf/etc/org.apache.cxf.osgi.cfg
```diff
@@ -0,0 +1 @@
+org.apache.cxf.servlet.context=/api
```
j-white, line 1: Let's copy this into the custom distribution as well for consistency

platform/alarms/itests/src/test/java/org/opennms/horizon/alarms/itests/CoreContainer.java
```diff
@@ -57,9 +58,11 @@ protected void configure() {
                 .withNetwork(Network.SHARED)
                 .withNetworkAliases(ALIAS)
                 .withCommand(containerCommand)
+                .withFileSystemBind("target/test-classes/users.properties", "/opt/horizon-stream/etc/users.properties", BindMode.READ_ONLY)
                 .waitingFor(new HttpWaitStrategy()
-                        .forPath("/alarms/list")
+                        .forPath("/cxf/alarms/list")
```
j-white, line 63: Can we make the same change to the container distribution?

artnaseef: Done.  Commit is imminent.

PR #12, opened 2022-04-12
Spring Boot Rest server implementation compare

yangli2014:
- The spring-rest branch has NodeDto contain LocationDto 
- The spring-rest branch using Hibernate saveOrUpdate to handle saving/merging parent entity: it looks simpler but has the potential risk of overwriting some existing values in the database?

- The spring-jpa branch uses the pure Spring JPA repositories and light-weight DTO
- It needs more code in the service layer, however, there is no risk of overwriting the data and more flexibility when dealing with more complex entity relationship mapping.


yangli2014:
I need feedback from our team.

yangli2014:
After rethinking that two branches and discussed with Art, I think we should go with the following approach
1. Adding saveOrUpdate with hibernate session into the JpaRepository
2. Using light-weight DTO to avoid heavy relational references in memory. This will simplify the UI end payload data structure (only reference entity ID).

yangli2014:
close this PR since we already decide using the Spring Boot with JPA by adding saveOrUpdate option

PR #23, opened 2022-05-03
the local kubernetes environment for REST API server

yangli2014:
## JIRA Link
https://issues.opennms.org/browse/HS-xx

## Checklist
* [ ] JIRA has been updated to 'in review'.
* [ ] Unit tests have been added as part of the PR.

### DIFF
--- a/local-docker-compose/kubernetes.kafka.yaml
+++ b/local-docker-compose/kubernetes.kafka.yaml
@@ -205,7 +205,7 @@ spec:
         - name: postgres-volume
         - name: postgres-initial-script
           configMap:
-            name: alarmd-camel-config
+            name: postgres-initial-sql
       containers:
         - name: postgres
           image: postgres:13.3-alpine
@@ -332,14 +332,12 @@ spec:
           image: opennms/horizon-stream-api:local
           imagePullPolicy: Never
           env:
-            - name: JAVA_OPTS
-              value: "-Djava.security.egd=file:/dev/./urandom -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005"
-            - name: DATASOURCE_URL
-              value: "jdbc:postgresql://${POSTGRES_SERVICE_HOST}:5432/horizon_stream"
-            - name: POSTGRES_SERVER
-              value: "postgres"
-            - name: KEYCLOAK_BASE_URL
+            - name: SPRING_DATASOURCE_URL
+              value: "jdbc:postgresql://postgres:5432/horizon_stream"
+            - name: KEYCLOAK_AUTH-SERVER-URL
               value: "http://keycloak:8080"
+            - name: KEYCLOAK_REALM
+              value: "opennms"
             - name: KEYCLOAK_ADMIN_USERNAME
               value: "keycloak-admin"
             - name: KEYCLOAK_ADMIN_PASSWORD
@@ -356,6 +354,12 @@ data:
     useSeda=false
     useKafka=true
     useActiveMq=false
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-initial-sql
+data:
   postgres.initial.script.sql: |
     CREATE USER opennms with password 'opennms';
     CREATE DATABASE horizon_stream;
--- a/local-docker-compose/kubernetes.kafka.yaml
+++ b/local-docker-compose/kubernetes.kafka.yaml
@@ -28,7 +28,7 @@ metadata:
 spec:
   type: ClusterIP
   ports:
-    - port: 8080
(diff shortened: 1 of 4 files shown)


local-docker-compose/kubernetes.kafka.yaml
```diff
@@ -326,14 +332,20 @@ spec:
           image: opennms/horizon-stream-api:local
           imagePullPolicy: Never
           env:
+            - name: JAVA_OPTS
+              value: "-Djava.security.egd=file:/dev/./urandom -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005"
```
geraldhumphries, line 336: We can probably get rid of JAVA_OPTS, that was added while we were debugging your container

local-docker-compose/kubernetes.kafka.yaml
```diff
@@ -326,14 +332,20 @@ spec:
...
+            - name: DATASOURCE_URL
+              value: "jdbc:postgresql://${POSTGRES_SERVICE_HOST}:5432/horizon_stream"
+            - name: POSTGRES_SERVER
+              value: "postgres"
```
geraldhumphries, line 340: DATASOURCE_URL should be SPRING_DATASOURCE_URL, with the value `jdbc:postgresql://postgres:5432/horizon_stream`. I don't think "POSTGRES_SERVER" is used.

rest-server/src/main/resources/application.yml
```diff
@@ -2,7 +2,7 @@ server:
   port: 9090
 spring:
   datasource:
-    url: jdbc:postgresql://localhost:5432/horizon_stream
+    url: jdbc:postgresql://${POSTGRES_SERVICE_HOST:localhost}:5432/horizon_stream
```
geraldhumphries, line 5: Rather than adding a new env var, just use `SPRING_DATASOURCE_URL=jdbc:postgresql://postgres:5432/horizon_stream` in the kubernetes manifest. It can stay as `url: jdbc:postgresql://localhost:5432/horizon_stream` in this file.

rest-server/src/main/resources/application.yml
```diff
@@ -30,8 +30,8 @@ graphql:
 # keycloak
 keycloak:
   realm: opennms
-  auth-server-url: http://localhost:8080
-  resource: login-app
+  auth-server-url: ${KEYCLOAK_BASE_URL:"http://localhost:8080"}
```
geraldhumphries, line 33: Rather than a new env var, just use `KEYCLOAK_AUTH-SERVER-URL=http://keycloak:8080` in the kubernetes manifest. It can stay as `auth-server-url: http://localhost:8080` in this file.

rest-server/src/main/docker/Dockerfile
```diff
@@ -3,4 +3,6 @@ FROM eclipse-temurin:11
 #USER spring:spring
 ARG JAR_FILE=target/*.jar
 COPY ${JAR_FILE} app.jar
+RUN apt-get update && apt-get install -y iputils-ping
+CMD bash
```
geraldhumphries, line 7: We can probably remove this now.

local-docker-compose/kubernetes.kafka.yaml
```diff
@@ -344,3 +356,7 @@ data:
     useSeda=false
     useKafka=true
     useActiveMq=false
+  postgres.initial.script.sql: |
```
geraldhumphries, line 359: It might be better to use a new ConfigMap, or rename this one to something other than `alarmd-camel-config`. @jaymberry What do you think?

geraldhumphries (changes requested):

local-docker-compose/kubernetes.kafka.yaml
yangli2014: good idea

rest-server/src/main/resources/application.yml
yangli2014: if the application.yml file keeps unchanged, how can the property value be populated with the correct value at the runtime?

geraldhumphries: It's one of the ways Spring Boot does configuration. You can use environment variables to override any of the properties in its configuration. This property's name is `keycloak.auth-server-url`, so you can use the env var `KEYCLOAK_AUTH-SERVER-URL` to override it without needing to insert your own env vars into the config files.

Spring Boot converts the environment variable names from uppercase/underscores to lowercase/periods. I am pretty sure the dashes stay the same, but I could be wrong.

yangli2014:
I pushed the change

geraldhumphries (approved):

PR #96, opened 2022-06-03
HS-137: Remove documentation that was moved to the Wiki

geraldhumphries:
- `design/PRINCIPLES.md` moved to https://github.com/OpenNMS/horizon-stream/wiki/Design-Principles
- `dev/README.md` moved to https://github.com/OpenNMS/horizon-stream/wiki/Getting-Started#local-development-with-skaffold-recommended
- `howto/KEYCLOAK-DEMO` was considered unnecessary, so it was removed.

Other files to look at:
- `platform/README.md` contains a lot of out-of-date info, or info that could be moved to the wiki (TBD), and should be cleaned up
- `external-it/external-horizon-stream-it/README.md` might need to be looked at

I've left the main README files for individual services.


dev/README.md
```diff
@@ -1,82 +0,0 @@
-# Local Development with Skaffold
```
artnaseef, line 1: Thinking this file may be better served to stay in the project, as the contents can change with commits and versions over time, and the wiki's versioning is distinct from the project's versioning.  If a developer were to pull up an older version (e.g. to see how something used-to-work, or verify whether a bug existed in an older version of the code), having this file in the source tree means that developer can find the right version of these instructions.

artnaseef (commented):
See the comments on the dev/README.md file.  Other than that comment, looks good!

dev/README.md
yangli2014: Even if it is in the project, we can't guarantee it will constantly be updated when the code changed. For example the following script 

`for i in $(docker images | grep **skaffold** | awk '{print $1":"$2}'); do docker rmi $i; done; docker images`

geraldhumphries:
I am really back and forth on what I want to do with the development/getting started guide, and if it should be in the codebase. Brought it back in the merge, will be updating the wiki with new things soon. Will bring it up at the dev meeting today.

geraldhumphries:
All documentation additions have been moved to the wiki.

cgorantla (approved):

PR #107, opened 2022-06-10
HS-61: Standard Flow Keycloak Login Templates

mikewrosey:
## JIRA Link
https://issues.opennms.org/browse/HS-61

## Relates to
https://issues.opennms.org/browse/HS-178

## Details of the change
Moves from password grand to standard flow.
Adds KC login templates w/ some feather styles.

screenshot-localhost_28080-2022 06 10-10_45_57
screenshot-localhost_28080-2022 06 10-10_46_39


dev/kubernetes.kafka.yaml
```diff
@@ -294,7 +294,7 @@ spec:
     spec:
       containers:
         - name: keycloak
-          image: quay.io/keycloak/keycloak:17.0.1 #After testing, version 18 and 17.0.1 works on M1 MacBook, for more information https://github.com/keycloak/keycloak/issues/8825
+          image: opennms/horizon-stream-keycloak
           args: [ "start-dev" ]
```
artnaseef, line 298: As an alternative to a custom docker image, could we use a configMap with the additional files?  That would simplify release and deployment.

keycloak-ui/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# After testing, version 18 and 17.0.1 works on M1 MacBook, for more information https://github.com/keycloak/keycloak/issues/8825
+FROM quay.io/keycloak/keycloak:18.0.0 as builder
```
artnaseef, line 2: Just a note here, in case it's not already clear - the `as builder` syntax isn't needed here.  It's only necessary with a multi-stage dockerfile (one that has multiple FROMs).

I felt like mentioning it because at first I thought it was multi-stage on seeing that, and it tripped me up, but that was my mistake.

artnaseef (commented):
Two questions coming to mind on the TTF files - first, is the license for those files compatible with including them in our source repo (and do we need to mention the license for them in a NOTICE file or the like)?

Second, do we want to store asset files like these in the source repo?  I'm wondering if the size of the repo could get rather large, impacting anyone that wants to develop (think slow/large `git clone`).  Also, is there any concern that updates to those assets require a rebuild and re-release?  Perhaps it's fine because they are fonts and only around 100kb each?

Note that I did review the javascript, but am not very good with it.  And I only quickly scanned the CSS.

Hope this helps!

mikewrosey:
> Two questions coming to mind on the TTF files - first, is the license for those files compatible with including them in our source repo (and do we need to mention the license for them in a NOTICE file or the like)?
> 
> Second, do we want to store asset files like these in the source repo? I'm wondering if the size of the repo could get rather large, impacting anyone that wants to develop (think slow/large `git clone`). Also, is there any concern that updates to those assets require a rebuild and re-release? Perhaps it's fine because they are fonts and only around 100kb each?
> 
> Note that I did review the javascript, but am not very good with it. And I only quickly scanned the CSS.
> 
> Hope this helps!

The font is licensed under OFL (open font license), which essentially means we can do anything with it except try to sell the font itself (no NOTICE required).

There are tradeoffs to storing assets vs getting them from a CDN. The primary reason for packaging is to guarantee that the user can access them from anywhere. I don't think size is an issue, but yes updates would require a rebuild. Versioning is a bit wild, different locations (Google fonts vs NPM, vs homebrew), could all be different. The idea with the KC templates here is not to use the latest version but the same version as FeatherDS, (which includes these assets in their repo).

Hope this provides more context..

mikewrosey:
> Two questions coming to mind on the TTF files - first, is the license for those files compatible with including them in our source repo (and do we need to mention the license for them in a NOTICE file or the like)?
> 
> Second, do we want to store asset files like these in the source repo? I'm wondering if the size of the repo could get rather large, impacting anyone that wants to develop (think slow/large `git clone`). Also, is there any concern that updates to those assets require a rebuild and re-release? Perhaps it's fine because they are fonts and only around 100kb each?
> 
> Note that I did review the javascript, but am not very good with it. And I only quickly scanned the CSS.
> 
> Hope this helps!

The font is licensed under OFL (open font license), which essentially means we can do anything with it except try to sell the font itself (no NOTICE required).

There are tradeoffs to storing assets vs getting them from a CDN. The primary reason for packaging is to guarantee that the user can access them from anywhere. I don't think size is an issue, but yes updates would require a rebuild. Versioning is a bit wild, different locations (Google fonts vs NPM, vs homebrew), could all be different. The idea with the KC templates here is not to use the latest version but the same version as FeatherDS, (which includes these assets in their repo).

Hope this provides more context..

dev/kubernetes.kafka.yaml
geraldhumphries: The most straightforward solution for deploying Keycloak with custom themes would be just mounting a volume with the files. We wouldn't need a custom image at all, and ConfigMaps aren't designed for this type of work.

This custom image isn't actually intended to be used in production, it's for development only. It was built this way to enable integration into Skaffold and the dev loop - Skaffold's file sync is used to keep the container up to date with local changes.

This is another instance where there are significant differences between the development and production images. Another example is the Vue app, where the dev version of the image runs a Vite dev server. I'm going to push a change that renames the image to `opennms/horizon-stream-keycloak-dev` to make that more clear and match the Vue image's convention.

keycloak-ui/Dockerfile
geraldhumphries: Good catch, I'll remove that and follow that convention in the future.

geraldhumphries:
A lot of the CSS in this PR looks fairly complicated, like it was generated from SCSS. An example is `keycloak-ui/themes/horizon-stream/common/resources/css/btn.css`. Was it copied from FeatherDS? If so, is there a better way to retrieve and refer to those files that would be easier to maintain and prevent developers from editing them? e.g. using NPM.

mikewrosey:
> A lot of the CSS in this PR looks fairly complicated, like it was generated from SCSS. An example is `keycloak-ui/themes/horizon-stream/common/resources/css/btn.css`. Was it copied from FeatherDS? If so, is there a better way to retrieve and refer to those files that would be easier to maintain and prevent developers from editing them? e.g. using NPM. It would also enable easier upgrades if we could do it that way.

Yep, some CSS (like theme/btn files), was generated by SCSS and pulled from feather. I thought about using NPM but I wasn't sure if it was worth the extra deployment work or pulling entire feather packages with all the JS/Vue code, just for the 3-4 style files. The other reason was that these templates are feather-esque, but won't be identical without using the real component, so minor version changes probably won't make a difference here, but if there was a theme overhaul, then yes, we would have to add that manually. What do you think ?

mikewrosey:
> Looks good, the FeatherDS dependencies will need some more discussion, and another issue will be created (@mikewrosey please link it here).

https://issues.opennms.org/browse/HS-178

PR #109, opened 2022-06-10
HS-158: Simple Helm charts for supporting services

jkapponms:
## JIRA Link
https://issues.opennms.org/browse/HS-158

Helm charts for the Keycloak and Strimzi Kafka operators


jkapponms:
Parsing the gargantuan config files in the operator and then having it apply them is more work than this was worth. 

Instead, I wrote simple Helm charts wrapping the config and checked in here. This way the existing Helm chart support the operator has can be reused.

jaymberry (commented):
Good to go.

PR #118, opened 2022-06-16
HS-156 inventory and scheduling

mfrazier63:
## JIRA Link
https://issues.opennms.org/browse/HS-156

## Checklist
* [x] JIRA has been updated to 'in review'.
* [x] Unit tests have been added as part of the PR.

## Details of the change

This adds in processing of the requisition data to generate whatever inventory records (nodes, interfaces, monitored services, etc) can be derived.

It also introduces a simple scheduling approach for repetitive scanning using the dynamic Camel routes using the quartz component, which utilized CRON syntax to specify the scanning interval.

For now, scanning timers are created when a manual scan is kicked off through the REST api. They automatically live forever until the server is stopped. Note that the size of the quartz thread pool is configurable through the provision service etc/cfg file in Karaf.


platform/provision/service/src/main/java/org/opennms/netmgt/provision/scan/NodeScanner.java
```diff
@@ -12,85 +11,80 @@
...
+    public void scanNode(RequisitionNodeDTO node) throws Exception {
+            context.addRoutes(createScheduledRoute(node));
     }
 
-    private RouteBuilder createRoutes() {
+    private RouteBuilder createScheduledRoute(RequisitionNodeDTO node) {
+        ForeignSourceDTO foreignSourceDTO = foreignSourceRepository.getForeignSource(node.getForeignId());
+
+        //TODO: do we need to be more sophisticated on building the cron string?
+        String routeOrigin = String.format("customQuartz://%s?cron=0 0/%d * 1/1 * ? *&triggerStartDelay=20",node.getNodeLabel(), foreignSourceDTO.getScanInterval().getStandardMinutes());
```
artnaseef, line 36: Looking at this again, I realized... there is a problem converting period-in-minutes to cron schedule. Cron is  based on the calendar, so if the period is not an even mulitple of minutes, hours, days, weeks, or months we can't specify the schedule with a single cron entry.

For example, if the schedule is defined with a period of 70 minutes starting at midnight on jan 1st, we would need the following cron entries and more:

    1 10 * * * ? * ? * ...
    2 20 * * * ? * ? * ...
    3 30 * * * ? * ? * ...
    ...

Let's follow-up after this PR.

artnaseef (approved):
See comment for future followup regarding cron.

Otherwise, looking really good.  I appreciate the PR even though we know there's more to be done here.

platform/provision/service/src/main/java/org/opennms/netmgt/provision/scan/NodeScanner.java
artnaseef: If the input is restricted, that's good.  I'm only seeing it at this level with the `getScanInterval().getStandardMinutes()` so it wasn't clear.

PR #136, opened 2022-06-29
HS-175 Convert REST API server to BFF component with GraphQL support

yangli2014:
## JIRA Link
https://issues.opennms.org/browse/HS-175
https://issues.opennms.org/browse/HS-209
https://issues.opennms.org/browse/HS-210
## Checklist
* [ ] JIRA has been updated to 'in review'.
* [ ] Unit tests have been added as part of the PR.

## Details of the change

- The Alarms and Events Rest API endpoints still exist. The API doc can be accessed at http://localhost:9090/swagger-ui/index.html
- The GraphQL schema and playground can be accessed at http://localhost:9090/gui
- The Nginx API gateway is configured to point to the API server.


yangli2014:
> 

As we discussed I will address it in HS-210

platform/events/api/src/main/java/org/opennms/horizon/events/xml/DateTimeAdapter.java
```diff
@@ -45,6 +45,12 @@ public String marshal(final Date date) throws Exception {
     /** {@inheritDoc} */
     @Override
     public Date unmarshal(final String string) throws Exception {
+        try{
+            Long time = Long.valueOf(string);
+            return new Date(time);
+        } catch (NumberFormatException e){
+
```
geraldhumphries, line 52: We should at least log something on exception here.

geraldhumphries (changes requested):
Feels really good seeing a PR that removes so much code :)

platform/events/api/src/main/java/org/opennms/horizon/events/xml/DateTimeAdapter.java
yangli2014: Sure. The code here is handle the date in timestamp or ISO string format.

yangli2014: After I cleaned up, this change is not needed anymore.

geraldhumphries (approved):

mfrazier63 (approved):
Looks much better!

PR #137, opened 2022-07-05
HS-115: merge the operator into the HS repo

jkapponms:
## JIRA Link
https://issues.opennms.org/browse/HS-115

## Details of the change
Merging the Operator's code into this repository


charts/opennms-operator/templates/cert/cert-cluster-issuer.yaml
```diff
@@ -0,0 +1,18 @@
+{{ if .Values.TLS.enabled }}
```
jkapponms, line 1: `TLS.Enabled`

PR #154, opened 2022-07-14
HS-206 Integrate minions table

chinh-cle:
## JIRA Link
https://issues.opennms.org/browse/HS-207-minions-table

## Details of the change
Integration using mock server


ui/mock-graphql/src/schema.ts
```diff
@@ -1,11 +1,14 @@
 import { makeExecutableSchema } from '@graphql-tools/schema'
-import { device, listDevices } from './store/Queries/devices'
-import typeDefs from './typeDefs'
+import typeDefs from './typeDefs.js'
```
mikewrosey, line 2: can we keep the .js off this one to indicate ts

ui/src/types/device.ts
```diff
@@ -0,0 +1,29 @@
...
+        name
+        icmp_latency
+        snmp_uptime
+      }
+      count
+      totalCount
+      offset
+    }
+  }
+`
```
mikewrosey, line 29: Don't think we need anything in this file and can be removed

ui/src/types/minions.ts
```diff
@@ -0,0 +1,29 @@
...
+        label
+        status
+        location
+      }
+      count
+      totalCount
+      offset
+    }
+  }
+`
```
mikewrosey, line 29: same with this one, we can use the generated types

ui/src/graphql/Queries/minionsQuery.graphql
```diff
@@ -0,0 +1,13 @@
...
+      id
+      label
+      status
+      location
+    }
+    count
+    totalCount
+    offset
+  }
+}
```
mikewrosey, line 13: Can we delete leave all the mock queries just in the mock folder for now? these aren't really used yet and will likely change once the endpoint is available

ui/package.json
```diff
@@ -3,10 +3,6 @@
   "version": "0.0.0",
   "license": "AGPL v3",
   "type": "module",
-  "private": true,
-  "workspaces": [
-    "mock-graphql"
-  ],
```
mikewrosey, line 9: If we keep the workspace, yarn should install both project pckgs at once, but on the other hand it's just extra build time for skaffold... Not sure if you had other reasons for removing it? but I'm good either way

PR #156, opened 2022-07-14
HS-148: Fixing the Kubernetes Replicator

jkapponms:
## JIRA Link
https://issues.opennms.org/browse/HS-148

## Details of the change
Fixed the kubernetes-replicator not being able to read anything from the cluster


jkapponms:
> I noticed the `packaged` folder and `.tgz` files were committed. That kind of looks like committing build artifacts, do you know if that's good practice with this stuff?

They're packaged Helm charts. Ideally we'd have some other place to use as our Helm repository, but for the time being they're being hosted as a Gist on this repository.

jkapponms:
> I noticed the `packaged` folder and `.tgz` files were committed. That kind of looks like committing build artifacts, do you know if that's good practice with this stuff?

To explain a bit more, when you `helm install` a publically hosted Helm chart, tarballs like the one I've commited are what `helm` is downloading

geraldhumphries:
Makes sense, sounds like they should stay in the repo until we have a better place for them. Just thought I'd flag that in case they were committed accidentally.

PR #165, opened 2022-07-15
HS-241: Update test for Minion registration

cgorantla:
## Description
Update test with awaitility

## Jira link(s)
- https://issues.opennms.org/browse/HS-241


yangli2014 (approved):
It looks good to me.

cgorantla:
Couple of things are blocking this:

1.  On the local-sample cluster that is running with external-it,  any rest query is going to ui service instead of core service.
2.  We don't have minion in the local-sample cluster yet.

external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/HorizonStreamTestSteps.java
```diff
@@ -135,12 +138,18 @@ public void sendGETRequestToHorizonStreamAtPath(String path) throws MalformedURL
 
     @Then("verify HTTP response code = {int}")
     public void verifyHTTPResponseCode(int expectedResponseCode) {
-        assertEquals(expectedResponseCode, restResponse.getStatusCode());
+        // Wait till Minion is registered to Horizon Core
+        await().atMost(180, TimeUnit.SECONDS).pollDelay(0, TimeUnit.SECONDS)
```
geraldhumphries, line 142: Awaitility looks really nice - we're using `org.opennms.horizon.RetryUtils` in other ITs to do something similar. Maybe it would be worth refactoring those other pieces to use Awaitility?

geraldhumphries (approved):
Looks good!

external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/HorizonStreamTestSteps.java
cgorantla: yes, `Awaitility` is good for  these types of checks.

geraldhumphries (approved):

cgorantla:
Created https://issues.opennms.org/browse/HS-376 to re-enable external-it

For now, merging this although this will not be executed.

PR #171, opened 2022-07-19
Jira/hs 190 195 graphql for devices

yangli2014:
## Jira link(s)
- https://issues.opennms.org/browse/HS-190
- https://issues.opennms.org/browse/HS-195

## Checklist

##Test
Test in GraphQL UI playground http://localhost:9090/gui


shared-lib/dto-model/pom.xml
```diff
@@ -15,6 +15,7 @@
     <properties>
         <lombok.version>1.18.24</lombok.version>
         <lombok.binding.version>0.2.0</lombok.binding.version>
+        <graphql-sqpr.version>0.11.2</graphql-sqpr.version>
```
cgorantla, line 18: Do we still need this here ?

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/service/DeviceService.java
```diff
@@ -52,6 +52,8 @@ public Integer createDevice(DeviceDTO newDevice) {
     OnmsNode node = mapper.fromDto(newDevice);
     if(node.getLocation() == null) {
       node.setLocation(sessionUtils.withReadOnlyTransaction(() -> locationDao.getDefaultLocation()));
```
cgorantla, line 54: i think we need to create `OnmsMonitoringLocation` even for `Default` location since it may not exist in DB yet.

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/service/DeviceService.java
```diff
@@ -52,6 +52,8 @@ public Integer createDevice(DeviceDTO newDevice) {
...
+    } else {
+        sessionUtils.withTransaction(() ->locationDao.saveOrUpdate(node.getLocation()));
```
cgorantla, line 56: Are we creating `OnmsMonitoringLocation` somewhere in the mapping ?

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/rest/api/DeviceRestAPI.java
```diff
@@ -28,6 +28,7 @@
 
 package org.opennms.horizon.inventory.device.rest.api;
```
cgorantla, line 29: I forgot to add this comment in previous PR,  Can we always return `javax.ws.rs.core.Response` even for GET methods ?

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/rest/api/DeviceRestAPI.java
```diff
@@ -55,5 +56,6 @@ public interface DeviceRestAPI {
   @POST
   @Produces(MediaType.APPLICATION_JSON)
   @Consumes(MediaType.APPLICATION_JSON)
+  @RolesAllowed("admin")
```
cgorantla, line 59: Do we need `@RolesAllowed("admin", "user")`   on `getById` and `findAll` as well ?

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/service/DeviceService.java
yangli2014: yes. Inside the DeviceMapper there is a LocationMapper

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/rest/api/DeviceRestAPI.java
yangli2014: user will have the lowest privilege (read-only). So I think we are good, also if we use key-cloak composite role mapping then the admin should include all the user privileges.

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/service/DeviceService.java
```diff
@@ -48,12 +51,16 @@ public DeviceCollectionDTO searchDevices() {
     return new DeviceCollectionDTO(deviceDTOS);
   }
 
-  public Integer createDevice(DeviceDTO newDevice) {
+  public void createDevice(DeviceDTO newDevice) {
     OnmsNode node = mapper.fromDto(newDevice);
     if(node.getLocation() == null) {
-      node.setLocation(sessionUtils.withReadOnlyTransaction(() -> locationDao.getDefaultLocation()));
+        OnmsMonitoringLocation location = new OnmsMonitoringLocation();
+        location.setLocationName(DEFAULT_LOCATION);
```
cgorantla, line 58: For both of these you can use `MonitoringLocationDao.DEFAULT_MONITORING_LOCATION_ID`

cgorantla (commented):
Can we rename `DeviceRestAPI` -> `DeviceRestService` and `DeviceRestAPIImpl`  ->  `DeviceRestServiceImpl`

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/rest/impl/DeviceRestAPIImpl.java
```diff
@@ -42,18 +41,23 @@ public void setService(DeviceService service) {
...
+  public Response findAll() {
+    return Response.ok(service.searchDevices()).build();
   }
 
   @Override
   public Response createDevice(final DeviceDTO device) {
     try {
-      return Response.ok(service.createDevice(device)).build();
+        service.createDevice(device);
+      return Response.ok("created").build();
```
cgorantla, line 60: Can we return created entity or path to created entity.  If the resource doesn't exist yet, we should return with Response.accepted().  Let me know what do you think

platform/db/dao-api/src/main/java/org/opennms/horizon/db/dao/api/MonitoringLocationDao.java
```diff
@@ -40,6 +40,7 @@
 public interface MonitoringLocationDao extends OnmsDao<OnmsMonitoringLocation, String> {
 
 	String DEFAULT_MONITORING_LOCATION_ID = "Default";
+    String DEFAULT_MONITORING_AREA = "localhost";
```
cgorantla, line 43: can use `DEFAULT_MONITORING_LOCATION_ID` for ` monitoringarea` as well.  This may be irrelevant field that can be removed later.

platform/inventory/device/src/main/java/org/opennms/horizon/inventory/device/rest/impl/DeviceRestAPIImpl.java
yangli2014: Good catch. I will update it

yangli2014: I suggest we update OnmsDao.save(T entity) return saved entity instead of K. It seems being used anywhere.

cgorantla: I think `K` is enough to send path for the resource.

yangli2014: Then I will just return the ID for the new device. The path in GraphQL result from the BFF end might not very usefull

cgorantla (approved):
LGTM!

PR #188, opened 2022-07-29
HS-193: TSDB DevOps

yangli2014:
## Description
Under the operator folder run the following commands
```
make copy-charts dependencies build
./operator
```
Wait for a while and run `kubectl get all -n local-instance` to verify the prometheus and pushgateway pods are created 

## Jira link(s)
- https://issues.opennms.org/browse/HS-193

## Checklist

### DIFF
--- a/operator/internal/reconciler/init.go
+++ b/operator/internal/reconciler/init.go
@@ -32,12 +32,12 @@ func (r *OpenNMSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 func (r *OpenNMSReconciler) InitServiceHandlers() {
 	r.Handlers = []handlers.ServiceHandler{
+		&handlers.PrometheusHandler{},
 		&handlers.BaseHandler{},
 		&handlers.PostgresHandler{},
 		&handlers.KeycloakHandler{},
-		&handlers.OpenNMSHandler{},
 		&handlers.IngressHandler{},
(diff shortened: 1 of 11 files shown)


geraldhumphries (approved):
Looks good.

Just a note: while we were debugging this, we renamed `Pushgateway` vars to `PushGateway`. It turns out that the official name for pushgateway does not use a capitalized letter G. Ideally we should go back to using a lowercase G, but I don't think it matters too much - the presence of pushgateway in the project is likely going to be temporary.

operator/internal/reconciler/init.go
```diff
@@ -32,11 +32,12 @@ func (r *OpenNMSReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 func (r *OpenNMSReconciler) InitServiceHandlers() {
 	r.Handlers = []handlers.ServiceHandler{
+		&handlers.PrometheusHandler{},
 		&handlers.BaseHandler{},
 		&handlers.PostgresHandler{},
 		&handlers.KeycloakHandler{},
-		&handlers.OpenNMSHandler{},
 		&handlers.IngressHandler{},
 		&handlers.ElasticsearchHandler{},
+		&handlers.OpenNMSHandler{},
```
jkapponms, line 41: Ingress should still be the last thing being deployed

PR #198, opened 2022-08-04
HS-265 Geomap page

chinh-cle:
## Description
- Add map to nav rail
- Convert vuex to pinia

## Jira link(s)
- https://issues.opennms.org/browse/HS-265-geomap-page


ui/src/components/Map/LeafletMap.vue
```diff
@@ -152,7 +153,6 @@ const iconCreateFunction = (cluster: Cluster) => {
   return divIcon({ html: `<span class=${highestSeverity}>` + cluster.getChildCount() + '</span>' })
 }
 
-const setIcon = (node: Node) => setMarkerColor(nodeLabelAlarmServerityMap.value[node.label])
```
mikewrosey, line 155: Probably keep this line, or comment it if needed. We might need it later to use the different color icons, which we haven't brought over from the core assets folder yet, but I think we should

ui/src/components/Map/LeafletMap.vue
```diff
@@ -221,7 +221,7 @@ const onLeafletReady = async () => {
     await nextTick()
 
     // save the bounds to state
-    store.dispatch('mapModule/setMapBounds', leafletObject.value.getBounds())
+    // store.dispatch('mapModule/setMapBounds', leafletObject.value.getBounds())
```
mikewrosey, line 224: will `mapStore.mapBounds = leafletObject.value.getBounds()` work ?

ui/src/store/Views/mapStore.ts
```diff
@@ -0,0 +1,151 @@
...
+      this.selectedSeverity = selectedSeverity
+    },
+    setSearchedNodeLabels(nodeLabels: string[]){
+      this.searchedNodeLabels = nodeLabels
+    },
+    setNodeSortObject(sortObj: FeatherSortObject) {
+      this.nodeSortObject = sortObj
+    },
+    setAlarmSortObject(sortObj: FeatherSortObject) {
+      this.alarmSortObject = sortObj
```
mikewrosey, line 89: For any action that's just assigning a state var, like `setNodeSortObject`, `setAlarmSortObject` etc. we should be able to get rid of it and just use `mapStore.nodeSortObject = someValue` in the component. Not a big deal, but it does mean we can get rid of some code

ui/src/store/Views/appStore.ts
```diff
@@ -0,0 +1,18 @@
...
+    <State>{
+      navRailIsOpen: false
+    },
+  actions: {
+    setNavRailOpen(navRailOpen: boolean) {
+      this.navRailIsOpen = navRailOpen
+    }
+  }
+})
+
```
mikewrosey, line 18: I think this is a near duplicate of the `layoutStore` and can probably use that one instead

ui/src/containers/Appliances.vue
```diff
@@ -38,7 +38,7 @@ const gapWidth = computed<string>(() => appliancesStore.minionsTableOpen ? '20px
     font-weight: bold;
   }
   .btns {
-    display: flex;
+    display: flex !important;
```
mikewrosey, line 41: instead of !important here, I think we can remove this below, line 64

```
    .btns {
      display: block;
    }
```

mikewrosey (approved):

PR #201, opened 2022-08-05
Feature/minion ng poms

splatch:
Fixes issues with maven poms.


mfrazier63 (commented):
I think I just want to reject this PR as I've already made those on the main branch (but didn't push it as I have more to do). But let's leave it here so I can double check that I got everything.

splatch:
Discarding it in favor of Mark changes. I will do spotless in a separate PR.

PR #209, opened 2022-08-10
Asn/grpc shared lib

artnaseef:
Add GRPC from POC into shared_lib.  Also adds a parent pom which still needs to be fit into platform and others.


mfrazier63 (approved):
LGTM.

Once nit.. rename module to "shared" or maybe "common". :-)

PR #242, opened 2022-08-17
HS-304 Display list in device table

chinh-cle:
## Description
- Highlight selected device
- Remove tab focus on mounted
- Filter tab list when device out of bounds
- Renaming geomap to map
- Use types from generated file
- Fix build errors
- Highlight device selected row
- Fix error when device has no location

## Jira link(s)
- https://issues.opennms.org/browse/HS-304-geomap-device-list

## Checklist


ui/src/components/Map/GridTabs.vue
```diff
@@ -26,12 +26,10 @@ const goToAlarms = () => router.push(`/map${route.query.nodeid ? '?nodeid=' + ro
 
 const goToNodes = () => router.push('/map/nodes')
 
-onActivated(() => {
-  if (router.currentRoute.value.name === 'MapAlarms') {
-    alarmTab.value.tab.click()
-  } else {
-    nodesTab.value.tab.click()
-  }
```
chinh-cle, line 34: not sure if this has anything todo with `MapKeepAlive`, it seems the hook is not being called anyhow.

mikewrosey: On horizon core, the keepAlive is used in App.vue like so:
```
      <router-view v-slot="{ Component }">
        <keep-alive include="MapKeepAlive">
          <component :is="Component" />
        </keep-alive>
      </router-view>
```

With this, onMounted is not called again and instead we rely on onActivated. This was done because there were bugs in the MarkerCluster package that occurred when remounting the map. With the limited geomap, we haven't run into those bugs yet, but it's possible they will arise once we add more functionality. We can keep using onMounted for now, until that happens.

ui/src/store/Views/mapStore.ts
```diff
@@ -1,59 +1,53 @@
...
-          (node: Node) =>
-            !(node.assetRecord.latitude == null || node.assetRecord.latitude.length === 0) &&
-            !(node.assetRecord.longitude == null || node.assetRecord.longitude.length === 0)
-        )
-
-        this.nodesWithCoordinates = nodes
+    async fetchDevices() {
+      const { data: devices } = await useQuery({
+        query: ListDevicesForMapDocument
+      })
```
mikewrosey, line 41: One pattern we might want to stick to is keeping all the `useQuery` calls for a particular feature inside their own query stores. So this would be in something like `mapQueries.ts`. That way we have a store that's just for the data, which *should* be easier to test. It also means we'll never have to do something like `fetchDevices()` inside an onMounted component hook. All other non-api related state can go inside the View folder stores. If needed, we can call useWhateverQuery from inside other stores, not just components. Let me know what you think. Doesn't have to be modified now, but going forward it's something that would make things consistent.

ui/src/components/Map/MapAlarmsGrid.vue
```diff
@@ -100,13 +100,15 @@
...
+import { AlarmDto } from '@/types/graphql'
+import { FeatherSortObject } from '@/types'
+import { AlarmQueryParameters } from '@/types/map'
 import { FeatherSelect } from '@featherds/select'
 import { FeatherCheckbox } from '@featherds/checkbox'
 import { FeatherSortHeader, SORT } from '@featherds/table'
 
 const mapStore = useMapStore()
-const alarms = computed<Alarm[]>(() => mapStore.fetchAlarms())
+const alarms = computed<AlarmDto[]>(() => mapStore.fetchAlarms())
```
mikewrosey, line 111: Looks like this one is causing the build issues

ui/src/store/Views/mapStore.ts
chinh-cle: @mikewrosey Let me know what you think of this 
https://github.com/OpenNMS/horizon-stream/pull/242/commits/59048412bdf276f1aee37b4f35ee585c1a9b01c5

ui/src/graphql/Views/mapQueries.graphql
```diff
@@ -1,20 +1,34 @@
...
+      sysOid
+    }
+  }
+}
+
+fragment LocationPartsForMap on Query {
+  listLocations {
+    locations {
+      latitude
+      longitude
```
mikewrosey, line 26: Do we still need this fragment on the map query ? Doesn't look like it's being used in mapQueries.ts

ui/src/store/Views/mapStore.ts
mikewrosey: Yea, that looks good, thanks for taking care of that

PR #252, opened 2022-08-24
HS-201: Call PagerDuty on alarm create

james8hut:
## Description
Call PagerDuty on alarm create

## Jira link(s)
- https://issues.opennms.org/browse/HS-201

## Flagged for review
Should AlarmSeverity class move into the shared-lib? There are also classes called Severity and OnmsSeverity which are identical as far as I can see.

## Checklist


notifications/src/main/java/org/opennms/horizon/notifications/api/dto/AlarmSeverity.java
```diff
@@ -0,0 +1,188 @@
...
+    WARNING(4, "Warning", "cyan"),
+    MINOR(5, "Minor", "yellow"),
+    MAJOR(6, "Major", "orange"),
+    CRITICAL(7, "Critical", "red");
+
+    private static final Map<Integer, AlarmSeverity> m_idMap;
+
+    private int m_id;
+    private String m_label;
+    private String m_color;
```
geraldhumphries, line 56: We should avoid introducing this style of Hungarian notation into the notification codebase. I know this was probably a copy paste from Classic, but it's a good idea to refactor it.

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/AlarmSeverity.java
```diff
@@ -0,0 +1,188 @@
...
+
+    static {
+        m_idMap = new HashMap<Integer, AlarmSeverity>(values().length);
+        for (final AlarmSeverity severity : values()) {
+            m_idMap.put(severity.getId(), severity);
+        }
+    }
+
+    private AlarmSeverity(final int id, final String label, final String color) {
```
geraldhumphries, line 65: I think the current preference for new code in Classic is to avoid the final keyword in method args. We should probably follow that as well.

geraldhumphries (changes requested):

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -135,36 +137,65 @@ private String getAuthToken() throws NotificationConfigUninitializedException {
...
-        payload.setSource("Horizon Stream");
-        payload.setComponent("todo");
+        AlarmSeverity alarmSeverity = AlarmSeverity.get(alarm.getSeverity());
+        payload.setSeverity(PagerDutySeverity.fromAlarmSeverity(alarmSeverity).toString());
+
+        if (alarm.getNodeLabel() != null) {
+            payload.setSource(alarm.getNodeLabel());
+        } else {
+            payload.setSource("unknown");
+        }
```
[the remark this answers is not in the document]
james8hut: Pro:
Once a field is marked as Optional, developer really can’t forget to handle the null case.

Yet, my personal opinion is that I don’t like it.

Con:
Code feels more complicated to look at. (Maybe with familiarity that feeling would reduce.) It also feels that the focus of the logic has shifted from the contents of the string, and what that means from a business perspective, to the fact that the string can be Null, and now the most important is handling the potential for Null.

I think it’s the wrapping that I don’t like. Feels like one extra thing to store in your head while trying to read and understand the code.

I’m aware that my dislike of this, is down to feeling rather than actual science or anything like that.

My feeling is that an attribute is a better feeling for how to describe nullable data.

e.g
@Nullable
String managedObjectType

I’m not sure that actually provides much value though.

My own thoughts are that if a String is null, and (a) it shouldn’t be, or (b) we’ve forgotten to do a check, then it’s ok to blow up with a NullPointerException as quickly as possible, so that we can fix the bug.

I should also point out I’m happy to use Optional if that’s what is decided is best across the entire project.

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -135,36 +137,65 @@ private String getAuthToken() throws NotificationConfigUninitializedException {
...
-        event.setDedup_key(notification.getDedupKey());
-        event.setEvent_action("trigger");
+        event.setDedup_key(alarm.getReductionKey());
+
+        if (AlarmSeverity.CLEARED.equals(alarmSeverity) || AlarmType.RESOLUTION.equals(alarm.getType())) {
+            event.setEvent_action("resolve");
+        } else if (alarm.getAckUser() != null && alarm.getAckUser().length() > 0) {
+            event.setEvent_action("acknowledge");
+        } else {
+            event.setEvent_action("trigger");
```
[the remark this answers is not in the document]
james8hut: done, and done

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -135,36 +137,65 @@ private String getAuthToken() throws NotificationConfigUninitializedException {
...
+        }
+
+        //TODO: Add in alarm id into url
         event.setClient("OpenNMS");
         event.setClient_url("http://opennms.com");
```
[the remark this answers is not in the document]
james8hut: will move

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/AlarmSeverity.java
james8hut: done

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/AlarmSeverity.java
james8hut: done. I must admit that I very rarely use the final keyword for anything other than for String constants defined in a ConstantsClass or something like that.
Out of curiosity, do you know why final is frowned upon? Is it simply as code noise?

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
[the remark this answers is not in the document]
james8hut: Having read up on people not liking final (see below), I think Optional hits the same flag for me, and feels like code noise.

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/AlarmSeverity.java
geraldhumphries: Honestly, for OpenNMS, not sure. I've never used it myself aside from constants and injected dependencies.

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/PagerDutyEventDTO.java
```diff
@@ -28,16 +28,26 @@
...
-    String event_action;
+
+    @JsonProperty("routing_key")
+    String routingKey;
+
+    @JsonProperty("dedup_key")
+    String dedupKey;
+
+    @JsonProperty("event_action")
+    String eventAction;
```
geraldhumphries, line 47: We should be able to use the `PagerDutyEventAction` enum here instead of a String. Adding `@JsonValue` to the `PagerDutyEventAction.getVal()` method should work for deserialization. 

Does this also need to be serialized, e.g. do we need to handle responses from PagerDuty that contain `PagerDutyEventAction`?

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -135,36 +138,65 @@ private String getAuthToken() throws NotificationConfigUninitializedException {
...
-        customDetails.setPingTime("1500ms");
-
-        payload.setSummary(notification.getMessage());
+        payload.setSummary(alarm.getLogMessage().trim());
         payload.setTimestamp(now.toString());
-        payload.setSeverity("critical");
-        payload.setSource("Horizon Stream");
-        payload.setComponent("todo");
+        AlarmSeverity alarmSeverity = AlarmSeverity.get(alarm.getSeverity());
+        payload.setSeverity(PagerDutySeverity.fromAlarmSeverity(alarmSeverity).toString());
```
geraldhumphries, line 151: Similar to `PagerDutyEventDTO`, can we use the severity enum in `PagerDutyPayloadDTO` directly?

geraldhumphries (changes requested):

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/PagerDutyEventDTO.java
james8hut: Currently only sending things one way to PagerDuty. Have implemented in line with the classic plugin.

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
james8hut: Have implemented in line with the classic plugin.

notifications/src/main/java/org/opennms/horizon/notifications/api/dto/PagerDutyEventDTO.java
james8hut: re-done using JsonValue

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
james8hut: re-done using JsonValue

geraldhumphries (approved):

PR #254, opened 2022-08-24
HS-333: config updates

jkapponms:


operator/scripts/local-onms-instance.yaml
```diff
@@ -21,3 +21,6 @@ spec:
     mem: 100Mi
   postgres:
     image: postgres:13.3-alpine
+  credentials:
+    adminPassword: admin
+    userPassword: passw0rd
```
jkapponms, line 26: passwords set on the CRD like this, the operator will defer to.

geraldhumphries (approved):

PR #258, opened 2022-08-25
Handling of application level reconnection.

splatch:
When GRPC channel leaves off the READY state, the stream observers become invalid.
While they do not throw any exception, it is not possible to receive nor publish any update using them.
GRPC api for java does not provide a way to track transition between managed channel states, it allows only to notify about state change.
Yet state change change is not being made unless there is an rpc operation requested.
Its kind of chicken-egg problem as GRPC client code we do, do not attempt to perform operations (where it is aware of channel state) unless channel is ready.
In order to gain a control over channel we need to force a state check. By this we can catch disconnect and connect time.
This commit introduces 30s window to monitor connection state.
In event of communication failure a reconnection attempt shall be made within this window.
Note that some operations (ie. rpc responses) will simply fail as there is no retry logic for them.

### DIFF
--- a/minion/assembly/src/main/resources/etc/org.ops4j.pax.logging.cfg
+++ b/minion/assembly/src/main/resources/etc/org.ops4j.pax.logging.cfg
@@ -26,7 +26,7 @@ color.debug = cyan
 color.trace = cyan
 
 # Common pattern layout for appenders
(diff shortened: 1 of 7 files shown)


cgorantla (approved):
LGTM!

artnaseef (approved):

PR #260, opened 2022-08-26
HS-249: add Grafana into Operator

jkapponms:
## Description
Adding Grafana into the list of deployables for the Operator

## Jira link(s)
- https://issues.opennms.org/browse/HS-249

## Checklist

### DIFF
--- a/charts/opennms/values.yaml
+++ b/charts/opennms/values.yaml
@@ -77,6 +77,7 @@ Postgres:
   OpenNMSPassword: notset
   KeycloakPassword: notset
   NotificationPassword: notset
+  GrafanaPassword: notset
   VolumeSize: 10Gi
 Keycloak:
   Path: /auth
@@ -95,7 +96,19 @@ Keycloak:
     BaseUserId: 00000000-0000-0000-0000-000000000000
     AdminUserId: 00000000-0000-0000-0000-000000000000
 Grafana:
-  Image: placeholder
+  Path: /grafana
+  Port: 3000
+  Replicas: 1
+  ServiceName: grafana
+  Image: opennms/horizon-stream-grafana
+  InitContainerImage: busybox
+  Resources:
+    Limits:
+      Cpu: 500m
+      Memory: 1Gi
+    Requests:
+      Cpu: 100m
+      Memory: 100Mi
 Prometheus:
   Enabled: true
   Server:
--- a/charts/opennms/values.yaml
+++ b/charts/opennms/values.yaml
@@ -60,6 +60,7 @@ OpenNMS:
   Notification: 
     ServiceName: opennms-notifications
     Image: opennms/horizon-stream-notification
+    ImagePullPolicy: IfNotPresent
     Replicas: 1
     Port: 8080
     Resources:
--- a/operator/internal/reconciler/init.go
+++ b/operator/internal/reconciler/init.go
@@ -36,8 +36,9 @@ func (r *OpenNMSReconciler) InitServiceHandlers() {
 		&handlers.PostgresHandler{},
 		&handlers.KeycloakHandler{},
 		&handlers.OpenNMSHandler{},
-		&handlers.ElasticsearchHandler{},
+		//&handlers.ElasticsearchHandler{}, TODO fix/replace
 		&handlers.PrometheusHandler{},
+		&handlers.GrafanaHandler{},
 		&handlers.MailServerHandler{},
 		&handlers.IngressHandler{}, // MUST BE LAST
 	}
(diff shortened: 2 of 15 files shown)


operator/internal/reconciler/init.go
```diff
@@ -36,8 +36,9 @@ func (r *OpenNMSReconciler) InitServiceHandlers() {
 		&handlers.PostgresHandler{},
 		&handlers.KeycloakHandler{},
 		&handlers.OpenNMSHandler{},
-		&handlers.ElasticsearchHandler{},
+		//&handlers.ElasticsearchHandler{}, TODO fix/replace
```
jkapponms, line 39: Disabling ES for now, per @j-white

operator/internal/reconciler/values_updater.go
```diff
@@ -90,24 +90,26 @@ func (r *OpenNMSReconciler) CheckForExistingPostgresCreds(ctx context.Context, v
...
 	v.Values.Postgres.NotificationPassword = notificationPwd
+	v.Values.Postgres.GrafanaPassword = grafanaPwd
 	return v, true
 }
 
 //setCorePasswords - sets randomly generated passwords for the core if not already set
 func setCorePasswords(tv values.TemplateValues, creds v1alpha1.Credentials) values.TemplateValues {
 	if creds.AdminPassword == "" {
-		tv.Values.Keycloak.AdminPassword = security.GeneratePassword(true)
+		tv.Values.Keycloak.AdminPassword = security.GeneratePassword(false)
```
jkapponms, line 107: Toggling off special characters for now. I think we need to audit how they're being used in config files where a password with special chars might accidentally comment itself out.

geraldhumphries (approved):

charts/opennms/values.yaml
```diff
@@ -95,7 +96,19 @@ Keycloak:
     BaseUserId: 00000000-0000-0000-0000-000000000000
     AdminUserId: 00000000-0000-0000-0000-000000000000
 Grafana:
-  Image: placeholder
+  Path: /grafana
+  Port: 3000
+  Replicas: 1
+  ServiceName: grafana
+  Image: opennms/horizon-stream-grafana
+  InitContainerImage: busybox
```
yangli2014, line 104: This value is not used in the deployment. Also the image changed to 
```
image: boky/alpine-bootstrap
          command: ["wait-for-service", "-v", "postgres://grafana@postgres:5432"]
```

yangli2014 (commented):
After running it locally I didn't see the Grafana pod deployed. I am not sure if this story is still ongoing.

operator/internal/reconciler/values_updater.go
geraldhumphries: I think a better solution is to store them as k8s Secrets (which also encodes everything as base64) and configure the deployments to reference the Secrets instead of using string templating. 

This means we will need to refactor how Postgres creates its initial database. However it can run shell scripts on startup just like it currently does with SQL. I think mounting the Secrets as files, reading them into a shell script, then using the values to create the databases/users would be the best option overall.

jkapponms: Postgres and Grafana here are the two using plaintext passwords in config files at the moment. What you described would be the ideal situation yea. 

Where possible, passwords are already being stored in secrets.

jkapponms:
> After running it locally I didn't see the Grafana pod deployed. I am not sure if this story is still ongoing.

It should be there, it was deploying fine when I tested it last before I left. Will sanity check this morning.

yangli2014:
How to test it? cd operator and run ```./scripts/install.sh dev```?

operator/scripts/install-operator-local.sh
```diff
@@ -12,9 +12,16 @@ make local-docker
 if [ $? -ne 0 ]; then exit; fi
 
 echo
-echo _______Pushing Docker Image into Kind Cluster______
+echo ______________Building Service Images______________
+echo
+mvn install -f ../shared-lib 2> /dev/null
+mvn package jib:dockerBuild -Dimage=opennms/horizon-stream-notification -Djib.from.platforms=linux/arm64 -f ../notifications 2> /dev/null
```
geraldhumphries, line 18: Why do we need to build this image in this file? Shouldn't we be pushing it onto Docker Hub and let the cluster pull it from there?

jkapponms: It isn't being uploaded to DockerHub yet is the issue. This is a stop gap.

yangli2014 (commented):
After I run the install script, the Grafana deployed and configured with the initial config database, prometheus data source, and minion dashboard. But there are still some k8s environment issues. I couldn't see any data in Grafana and the UI home page doesn't have any minions show up.

geraldhumphries:
I think we can get this merged in now. There are a couple of issues that will need to be iterated on like Yang's issues, but it is still a step forward. Pushing notification service images was bumped to the next sprint.

geraldhumphries (approved):

yangli2014 (approved):

PR #274, opened 2022-09-06
HS-358: Use Liquibase for notications database

james8hut:
## Description
using Liquibase for the notification database

## Jira link(s)
- https://issues.opennms.org/browse/HS-358

## Flagged for review
Should I refactor the Hungarian notation out of Migrator and MigratorAdminInitialize while I'm in here

## Checklist


platform/db/schema-liquibase/src/main/liquibase/hs-0.1.0/tables/pagerDutyConfig.xml
```diff
@@ -0,0 +1,36 @@
...
+			<!-- Integration key for calling PagerDuty API. -->
+			<column name="integrationkey" type="varchar(256)">
+				<constraints nullable="false" />
+			</column>
+		</createTable>
+	</changeSet>
+
+    <changeSet author="jhutc" id="pager_duty_config_insert">
+        <sql>
+            INSERT INTO pager_duty_config (token, integrationkey) VALUES ('x', 'y');
```
cgorantla, line 32: what are these initial values supposed to be ?

cgorantla (changes requested):
I feel like we should separate notification data source implementation out of existing DB migration.   That would make it easy to maintain and cleaner.

platform/db/schema-liquibase/src/main/liquibase/hs-0.1.0/tables/pagerDutyConfig.xml
james8hut: So the config table just contains a single row containing token and integration_key for calling PagerDuty. User would be required to update these values (via UI) to enable PagerDuty to work. When they do so, the single row is updated with valid values. I'll update the 'x' and 'y' values to 'INVALID' as that makes more sense. If the user attempts to use Pager Duty before setting up the config table, then an exception is thrown from PagerDuty API with Invalid Routing key.

cgorantla: Why inserting in the first place ?

cgorantla: If we don't have sensible defaults, I prefer not to insert anything.

james8hut: Have updated not to insert anything

geraldhumphries: I agree with Chandra, this would probably end up being more difficult to maintain in the long run. It's a little trick we would always need to keep in our heads.

geraldhumphries:
You may be able to reduce a lot of boilerplate with the use of the SpringLiquibase bean. 

Here's an example of it in use: https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/java/io/github/jhipster/sample/config/LiquibaseConfiguration.java#L32

associated SpringLiquibaseUtilClass: https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/springboot/dbmigration/liquibase/main/SpringLiquibaseUtil.java.mustache

And finally, Spring Boot has a LiquibaseAutoConfiguration bean that we may be able to use, but I have no experience with it: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java

cgorantla:
@james8hut Do you still need changes in platform/db/schema ?

james8hut:
> @james8hut Do you still need changes in platform/db/schema ?

@cgorantla In the process of rolling all platform changes back :-)

notifications/pom.xml
```diff
@@ -71,7 +75,11 @@
             <artifactId>postgresql</artifactId>
             <scope>runtime</scope>
         </dependency>
-	</dependencies>
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
```
cgorantla, line 80: Add scope test

geraldhumphries (approved):

notifications/src/main/resources/application.yml
```diff
@@ -2,6 +2,12 @@ spring:
   application:
     name: horizon-stream-notifications
 
+  datasource:
```
cgorantla, line 5: Is this DB already part of skaffold/operator pipeline ?
If not we should create another issue to tackle that.

cgorantla (approved):
LGTM!

notifications/src/main/resources/application.yml
geraldhumphries: Yeah, it is, as part of the postgres-initial-sql scripts.

notifications/pom.xml
geraldhumphries:
```suggestion
            <artifactId>h2</artifactId>
            <scope>test</scope>
```

PR #275, opened 2022-09-06
Bridge ignite-resource concept closer to osgi

splatch:
The ignite ioc api is tied to Spring, however it use is fairly limited. If we stub two methods of Spring framework `ApplicationContext` we are able to take over control over sourcing and lookup performed to bring beans into ignite services.


splatch:
@mfrazier63 @artnaseef I need your suggestion on how to tear down actual use of `OsgiServiceHolder`. It use is in types which are created within a factory type which is controlled by Ignite itself.

PR #282, opened 2022-09-08
HS-337,HS-338: Stats: Create platform module for data choices

thomasbigger584:
## Description
Adding initial datachoices module which will (eventually) send a payload to stats.opennms.com to a new seperate Elastic search cluster and grafana dashboard.
This has been tested with a locally deployed version of https://github.com/OpenNMS/usage-stats-handler/pull/4 which will eventually get deployed to the stats.opennms.com.

This uses the database to store the generated system ID used to identify the installation, and uses the maven project version as the version to be sent to stats.opennms.com. 

The version will need to be updated during the build and deploy pipeline using maven CLI.

Added endpoint to toggle the datachoices (Opt-In/Opt-Out) timer to run. If the datachoices has been enabled, then it will also be enabled after an application restart. The enabled flag is also stored in the database. This has been tested by restarting the k8s pod and seeing the timer running automatically.

Additional work is to follow up this PR to add more metrics into the payload, also to add the relevant endpoint to rest-server for the UI to consume.

## Jira link(s)
- https://issues.opennms.org/browse/HS-337
- https://issues.opennms.org/browse/HS-338

## Checklist

### DIFF
--- a/platform/datachoices/src/main/java/org/opennms/horizon/datachoices/internal/UsageStatisticsReporter.java
+++ b/platform/datachoices/src/main/java/org/opennms/horizon/datachoices/internal/UsageStatisticsReporter.java
@@ -1,0 +1,156 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.horizon.datachoices.internal;
+
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.ContentType;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.opennms.core.web.HttpClientWrapper;
+import org.opennms.horizon.datachoices.dto.UsageStatisticsReportDTO;
+import org.opennms.horizon.datachoices.internal.StateManager.StateChangeHandler;
+import org.opennms.horizon.db.model.OnmsDataChoices;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.Timer;
+import java.util.TimerTask;
+
+public class UsageStatisticsReporter implements StateChangeHandler {
+    private static final Logger LOG = LoggerFactory.getLogger(UsageStatisticsReporter.class);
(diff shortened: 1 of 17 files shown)


platform/datachoices/src/main/java/org/opennms/horizon/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private String url;
+
+    private long interval;
+
+    private Timer timer;
+
+    public synchronized void init() {
+        OnmsDataChoices dataChoices = this.stateManager.getDataChoices();
+        if (Boolean.TRUE.equals(dataChoices.getEnabled())) {
```
james8hut, line 65: Why not simply: 
if (dataChoices.getEnabled())

james8hut (commented):
One minor code quibble, plus a general question. Should we be writing JUnit tests to prove all this stuff works?

platform/db/dao-impl/src/main/java/org/opennms/horizon/db/dao/impl/DataChoicesDaoHibernate.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public OnmsDataChoices find() {
+        Optional<OnmsDataChoices> dataChoicesOpt = findFirst();
+        if (dataChoicesOpt.isPresent()) {
+            return dataChoicesOpt.get();
+        }
+
+        OnmsDataChoices onmsDataChoices = new OnmsDataChoices();
+        onmsDataChoices.setSystemId(UUID.randomUUID().toString());
```
cgorantla, line 57: we could use `SystemInfoUtils.DEFAULT_INSTANCE_ID` for the SystemId.

platform/datachoices/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,46 @@
...
+                https://osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd
+                http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.3.0
+                http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.3.0.xsd
+">
+
+    <cm:property-placeholder persistent-id="org.opennms.horizon.datachoices">
+        <cm:default-properties>
+
+            <!-- TODO: Change this once stats.opennms.com has been updated for the /hs-usage-report endpoint -->
+            <cm:property name="url" value="http://192.168.1.90:3542/"/> <!-- IP of Machine for local testing -->
```
cgorantla, line 15: Just name it as `http://stats.opennms.com` instead of using local url

platform/datachoices/src/main/java/org/opennms/horizon/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -0,0 +1,156 @@
...
+
+        @Override
+        public void run() {
+            UsageStatisticsReportDTO usageStatsReport = generateReport();
+            String usageStatsReportJson = usageStatsReport.toJson();
+
+            try (HttpClientWrapper clientWrapper = HttpClientWrapper.create();
+                 CloseableHttpClient client = clientWrapper.getClient()) {
+
+                HttpPost httpRequest = new HttpPost(url + USAGE_REPORT);
```
cgorantla, line 144: we should also able to configure some default username/password with Basic authentication

platform/db/dao-impl/src/main/java/org/opennms/horizon/db/dao/impl/DataChoicesDaoHibernate.java
thomasbigger584: ` public static final String DEFAULT_INSTANCE_ID = "OpenNMS";` 
   This would not be unique for each system install? Classic also uses a `UUID` which identifies the installation.

cgorantla: yeah, we may want to generate unique ID for stats. Good with what you have there.

thomasbigger584:
> One minor code quibble, plus a general question. Should we be writing JUnit tests to prove all this stuff works?

@geraldhumphries @cgorantla @james8hut 

 I see alot of the modules dont have testing considered. 

What is our strategy for integration testing with karaf and the database?

cgorantla:
> > One minor code quibble, plus a general question. Should we be writing JUnit tests to prove all this stuff works?
> 
> @geraldhumphries @cgorantla @james8hut
> 
> I see alot of the modules dont have testing considered.
> 
> What is our strategy for integration testing with karaf and the database in the platforms folder?

We are only adding unit tests, use mocks as necessary but don't need to wire in any of the dependencies like DB/Kafka.

If you need to test whole End to End tests, https://github.com/OpenNMS/horizon-stream/tree/develop/external-it is a good place to add. It requires cluster to be running.

PR #283, opened 2022-09-08
HS-379: Add support for config updates / event based discovery of devices

cgorantla:
## Description
Add support for sending events for config updates
Add event based discovery for Traps

## Jira link(s)
- https://issues.opennms.org/browse/HS-379
- https://issues.opennms.org/browse/HS-364

## Checklist
* [] Includes [appropriate tests.](https://github.com/OpenNMS/horizon-stream/wiki/Test-Strategy)


cgorantla:
To validate this:

Update config by setting  `newSuspectOnTrap` to true on Core from Rest call : http://localhost:18181/config/snmp-traps

And send a trap, that trap should generate event on Core and also generate another new suspect event.
And that event should add a device as discovered device and start collecting metrics for that new device.

platform/config/service/src/main/java/org/opennms/horizon/config/service/impl/DefaultConfigService.java
```diff
@@ -44,34 +49,48 @@
...
     }
 
 
     @Override
-    public void addConfig(String configName, String jsonConfig) {
+    public void addConfig(String configName, String jsonConfig, String source) {
         addConfigNameToJsonStore(configName);
         try {
+            Optional<String> existingConfig = getConfig(configName);
+            if (existingConfig.isPresent() && existingConfig.get().equals(jsonConfig)) {
```
yangli2014, line 68: We added the configure name to the JSon store at line 65 first then we check if the config exists, it seems will always return true here.

platform/config/service/src/main/java/org/opennms/horizon/config/service/impl/DefaultConfigService.java
```diff
@@ -44,34 +49,48 @@
...
+            LOG.error("Exception while parsing config for {}", configName, e);
             throw new IllegalArgumentException("Invalid json " + jsonConfig);
         }
         jsonStore.put(configName, jsonConfig, CONFIG);
+        sendConfigUpdatedEvent(configName, source);
     }
 
     @Override
-    public void updateConfig(String configName, String jsonConfig) {
+    public void updateConfig(String configName, String jsonConfig, String source) {
```
yangli2014, line 81: This method has exactly the same code as ```addConfig``` above.

yangli2014 (approved):
approved with minor comments

platform/config/service/src/main/java/org/opennms/horizon/config/service/impl/DefaultConfigService.java
cgorantla: Config name is in separate context in json store and Config may not be stored if it is not valid Json.
Config names are stored separately not along with config.

PR #289, opened 2022-09-12
Use PROTOBUF model objects for monitor configuration and results.

artnaseef:
Rip out dynamic plugin config for now; it is currently unused and has a good chance of changing, so not worth maintaining for now.

## Jira link(s)

## Checklist


shared-lib/horizon-echo-contract/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.opennms.horizon.shared</groupId>
+        <artifactId>shared-lib</artifactId>
+        <version>0.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>horizon-echo-contract</artifactId>
```
splatch, line 14: I am not sure if we're in place where we should introduce multiple modules for contracts. I get that its early step to avoid coupling of dependencies, however we are at low risk if we keep grpc-contract just a container for code generated by proto.

shared-lib/horizon-echo-contract/src/main/proto/echo.proto
```diff
@@ -29,35 +29,10 @@ syntax = "proto3";
 
 import "google/protobuf/any.proto";
 
-package workflow;
```
splatch, line 32: At least on diff it seems that git mixed this file with other one.

shared-lib/task-set/src/main/proto/weblisten.proto
```diff
@@ -0,0 +1,39 @@
...
+ *******************************************************************************/
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package opennms.echo;
+option java_multiple_files = true;
+option java_package = "org.opennms.weblisten.contract";
+
+message WebListenRequest {
```
splatch, line 36: Not sure if naming is proper, could you please add here some comment for clarity?

splatch (approved):
Its a good direction. I have some more changes on ping/snmo rpc structures waiting since a while on other feature branches, but I can definitely fit them to this structure.

mfrazier63 (approved):
LGTM

shared-lib/horizon-echo-contract/pom.xml
artnaseef: You're right - I started to do that, and then changed my mind.  That artifact is actually unused, so I'll remove it.  The "real one" is under `shared-lib/task-set/src/main/proto`.  Good catch.

shared-lib/horizon-echo-contract/src/main/proto/echo.proto
artnaseef: Oh, actually I did that - missed the cleanup.

shared-lib/task-set/src/main/proto/weblisten.proto
artnaseef: Sure, I'll add a comment.  LISTENER is a specific type of task in which the service listens for samples (e.g. an SNMP Trap listener).  Note that WebListener is actually a prototype class, so I anticipate we will just remove it - but I left it for now.

PR #296, opened 2022-09-20
HS-394: add alarm to PagerDuty custom details

james8hut:
## Description
Following Dev Jam, we added the alarm details to the PagerDuty custom details in the plugin classic uses. Also, I realised that the token I was using isn't needed. Have also removed the validation of that token also. So duplicating that work in streams.

## Jira link(s)
- https://issues.opennms.org/browse/HS-394

## Flagged for review
none

## Checklist

### DIFF
--- a/notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
+++ b/notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
@@ -149,42 +149,23 @@ public class PagerDutyAPIImpl implements PagerDutyAPI {
 
         payload.setCustomDetails(new HashMap<>());
 
+        // Add all of the event parameters as custom details
         EventDTO lastEvent = alarm.getLastEvent();
-
         if (lastEvent != null) {
             Map<String, Object> customDetails = eparmsToMap(lastEvent.getParameters());
             payload.getCustomDetails().putAll(customDetails);
         }
 
-        // If the event parameters contains a field called 'alarm', then we cannot
-        // add the actual alarm into the map under the same key.
-        String customDetailsAlarmName = getUniqueAlarmNameForCustomDetails("alarm", payload.getCustomDetails());
-
         ObjectMapper mapper = new ObjectMapper();
+
+        // If the event parameters contains a field called 'alarm', then the alarm itself overwrites that (by design).
         JsonNode alarmJson = mapper.convertValue(alarm, JsonNode.class);
-        payload.getCustomDetails().put(customDetailsAlarmName, alarmJson);
+        payload.getCustomDetails().put("alarm", alarmJson);
 
         event.setPayload(payload);
 
         return objectMapper.writeValueAsString(event);
     }
-
-    private String getUniqueAlarmNameForCustomDetails(String customDetailsAlarmName, Map<String, Object> customDetails) {
-        int suffix = 0;
-        boolean loop = true;
-
-        // Need to loop here in case 'alarm_1', 'alarm_2' etc is somehow in custom details as well.
-        while (loop) {
-            if (customDetails.containsKey(customDetailsAlarmName)) {
-                customDetailsAlarmName = "alarm_" + ++suffix;
-            } else {
-                loop = false;
-            }
-        }
-
-        return customDetailsAlarmName;
-    }
-
 
     protected static Map<String, Object> eparmsToMap(List<EventParameterDTO> eparms) {
         final Map<String, Object> map = new LinkedHashMap<>();
--- a/notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
+++ b/notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
@@ -156,10 +156,8 @@ public class PagerDutyAPIImpl implements PagerDutyAPI {
             payload.getCustomDetails().putAll(customDetails);
         }
 
-        ObjectMapper mapper = new ObjectMapper();
-
         // If the event parameters contains a field called 'alarm', then the alarm itself overwrites that (by design).
-        JsonNode alarmJson = mapper.convertValue(alarm, JsonNode.class);
+        JsonNode alarmJson = objectMapper.convertValue(alarm, JsonNode.class);
         payload.getCustomDetails().put("alarm", alarmJson);
 
         event.setPayload(payload);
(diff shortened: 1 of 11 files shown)


notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -64,6 +66,9 @@ public class PagerDutyAPIImpl implements PagerDutyAPI {
     @Autowired
     RestTemplate restTemplate;
 
+    @Autowired
+    ObjectMapper mapper;
```
cgorantla, line 70: Should ObjectMapper should be auto-wired ?

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -179,18 +151,46 @@ private String getEvent(AlarmDTO alarm) throws NotificationConfigUninitializedEx
...
+            loop = false;
+        }
+
+        while (loop) {
+            if (customDetails.get(customDetailsAlarmName) != null) {
+                customDetailsAlarmName = "alarm_" + ++suffix;
+            } else {
+                loop = false;
+            }
+        }
```
cgorantla, line 188: Am I missing something  here , why do we need loop to get something from Map ?

notifications/src/main/resources/db/changelog/hs-0.1.0/tables/pagerDutyConfig.xml
```diff
@@ -14,12 +14,6 @@
 		</preConditions> 
 
 		<createTable tableName="pager_duty_config">
-
```
cgorantla, line 17: Shouldn't this be performed as new changeSet ?

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
james8hut: A good question. If not, then tests will always pick up the concrete implementation, which I guess is what we want. Will change.

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
james8hut: The existing implementation in classic will place all event parameters in a map to be placed in the custom details field.
We are adding to that by adding a value with key called 'alarm' to the map containing all the alarm details. Now consider the case where there is already an event parameter called 'alarm'. You can't have two entries with the same key in a map, so we need to avoid that clash. We do that by adding the alarm into a field called 'alarm_1'.
Now consider the (unlikely) case where there is also an event parameter called 'alarm_1' (and another called 'alarm_2' etc). We need to avoid that clash also. Hence the loop.

I'll add a comment to explain this in the code. I've also just realised I should be using containsKey rather than get with a null check.

notifications/src/main/resources/db/changelog/hs-0.1.0/tables/pagerDutyConfig.xml
james8hut: So this table was created as part of the green thread. As our database isn't persisted yet, I thought it would also work if I just changed the original file, as nobody has a copy of this table yet. I thought it was simpler just to have a single changeset with the correct version of the table in it.

If there is any perceived danger to this approach, I can change and do a new changeSet

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -179,18 +147,45 @@ private String getEvent(AlarmDTO alarm) throws NotificationConfigUninitializedEx
...
+        return objectMapper.writeValueAsString(event);
+    }
+
+    private String getUniqueAlarmNameForCustomDetails(String customDetailsAlarmName, Map<String, Object> customDetails) {
+        int suffix = 0;
+        boolean loop = true;
+
+        // Need to loop here in case 'alarm_1', 'alarm_2' etc is somehow in custom details as well.
+        while (loop) {
+            if (customDetails.containsKey(customDetailsAlarmName)) {
```
cgorantla, line 178: There is no need for iteration for containsKey or get . This will always return alarm_1.  why loop ?

james8hut: Not true. customDetailsAlarmName is altered on line 179. It enters the method as "alarm". If that is present in the customDetails map, it is altered to "alarm_1". If that's present, it is altered to "alarm_2", and so on, until we find a unique name that is not present in the custom details, and then we return that string.

This code was done to prevent an event parameter of alarm being overwritten, by the alarm details. However, I've just spoken to Jesse about this (I was doing the same work for the classic plugin), and he says that actually that's fine, and this method actually isn't needed, so I'll remove it.

notifications/src/main/resources/db/changelog/hs-0.1.0/tables/pagerDutyConfig.xml
cgorantla: that's fine then.

notifications/src/main/java/org/opennms/horizon/notifications/api/PagerDutyAPIImpl.java
```diff
@@ -179,16 +147,24 @@ private String getEvent(AlarmDTO alarm) throws NotificationConfigUninitializedEx
...
 
+        // Add all of the event parameters as custom details
+        EventDTO lastEvent = alarm.getLastEvent();
         if (lastEvent != null) {
             Map<String, Object> customDetails = eparmsToMap(lastEvent.getParameters());
-            payload.setCustomDetails(customDetails);
+            payload.getCustomDetails().putAll(customDetails);
         }
+
+        ObjectMapper mapper = new ObjectMapper();
```
cgorantla, line 159: ObjectMapper could be initialized as member variable as this is getting used in https://github.com/OpenNMS/horizon-stream/pull/296/files#diff-9b7a27f64c50f77415331c380129d9f6d9ae8e79c72031a3ea8e2438076f3abeR110 as well

james8hut: I see that it's actually been used twice in the same method...
Have deleted second copy.
ok to keep a single copy as local variable?

cgorantla: That's also fine!

PR #304, opened 2022-09-22
Removing the KC operator

jkapponms:
## Description
Removing the KC operator from the instances and instead just deploying our own deployment.


geraldhumphries (approved):
LGTM!

Note for others that see the addition of packaged charts here, there's an open issue for getting them out of the repo: https://issues.opennms.org/browse/HS-259

PR #309, opened 2022-09-22
HS-365: Add syslog consumer

cgorantla:
## Description
Add syslog  consumer.

Existing config files from base assembly are loaded in memory and persisted as json.
Any new files with `syslogd-configuration-group` should be uploaded through Config API ( doesn't exist yet)

## Jira link(s)
- https://issues.opennms.org/browse/HS-365

## Checklist


cgorantla:
Verified that this works but syslog messages are dropped as they don't match with the regex. Need to check further

thomasbigger584 (approved):
LGTM, only comment I have is of testing but I see thats being considered after this PR.

PR #310, opened 2022-09-23
HS-354: expose notification endpoint through graphQL

james8hut:
## Description
expose notification endpoint through graphQL

## Jira link(s)
- https://issues.opennms.org/browse/HS-354

## Flagged for review
none

## Checklist


rest-server/src/main/java/org/opennms/horizon/server/service/NotificationsService.java
```diff
@@ -0,0 +1,26 @@
...
+public class NotificationsService {
+    private final NotificationGateway gateway;
+
+    public NotificationsService(NotificationGateway gateway) {
+        this.gateway = gateway;
+    }
+
+    @GraphQLMutation
+    public Mono<Void> savePagerDutyConfig(PagerDutyConfigDTO config, @GraphQLEnvironment ResolutionEnvironment env) {
+        return gateway.post(NotificationGateway.URL_PATH_NOTIFICATIONS_CONFIG, gateway.getAuthHeader(env), config, Void.class);
```
yangli2014, line 23: Since the notification server doesn't have an authentication logic, the keycloak token is not in use. It is OK for now, we might move the authentication and authorization into the BFF server.

yangli2014:
It looks good to me. However when tried to invoke the new GraphQL mutation method I got the following error in notification server 
```
Detail: Failing row contains (null, test-key).; nested exception is org.postgresql.util.PSQLException: ERROR: null value in column "token" of relation "pager_duty_config" violates not-null constraint
```

yangli2014 (approved):

rest-server/src/main/java/org/opennms/horizon/server/service/NotificationsService.java
james8hut: I think we need authentication logic in the notification server. If the logic is just in the BFF server, it might be possible to hit the notification server directly without going through the BFF, and then we would have no authentication protection

PR #337, opened 2022-09-29
[HS-274,HS-398,HS-422,HS-428 - HS-431] Helm chart, operator, and Skaffold updates

geraldhumphries:
## Description

- Helm chart is now deployable on its own, without the operator
- Skaffold now deploys using the Helm chart
- Deploying through Helm now respects the namespace
- Uses multi-arch Kafka image that will start on ARM machines
- Fix Kafka listeners not working
- Fix operator bug causing it to fail finding Secrets/ConfigMaps when the CR name and namespace are different
- Docker image produced by the operator's Makefile is now configurable
- Fix operator hanging on the Keycloak deployment because of missing cert secret
- Remove now-unused kubernetes.kafka.yaml file

## Jira link(s)
- https://issues.opennms.org/browse/HS-274
- https://issues.opennms.org/browse/HS-398
- https://issues.opennms.org/browse/HS-421
- https://issues.opennms.org/browse/HS-422
- https://issues.opennms.org/browse/HS-428
- https://issues.opennms.org/browse/HS-429
- https://issues.opennms.org/browse/HS-430
- https://issues.opennms.org/browse/HS-431

## Checklist


charts/opennms/templates/keycloak/keycloak-deployment.yaml
```diff
@@ -91,28 +91,30 @@ spec:
...
           livenessProbe:
-            exec:
-              command:
-              - curl
-              - --head
-              - --fail
-              - --silent
-              - --insecure
-              - https://127.0.0.1:{{ .Values.Keycloak.HttpsPort }}{{ .Values.Keycloak.Path }}/health/live
+            httpGet:
```
jkapponms, line 94: this change to the probes is also duplicated in your other KC PR

charts/opennms/templates/keycloak/keycloak-realm-configmap.yaml
```diff
@@ -2,11 +2,11 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: keycloak-realm-configmap
-  namespace: {{ .Values.Namespace }}
+  namespace: {{ .Release.Namespace }}
 data:
   onms-realm.json: |
     {
-      "id": "{{ .Values.Keycloak.UUID.RealmId }}",
+      "id": "{{ .Values.Keycloak.UUID.RealmId | default uuidv4 }}",
```
jkapponms, line 9: Yea, KC PR is removing these keys

operator/internal/reconciler/values_updater.go
```diff
@@ -31,9 +31,10 @@ func (r *OpenNMSReconciler) UpdateValues(ctx context.Context, instance v1alpha1.
 		r.ValuesMap = map[string]values.TemplateValues{}
 	}
 
-	namespace := instance.Name
+	name := instance.Name
+	namespace := instance.Spec.Namespace
 
-	templateValues, ok := r.ValuesMap[namespace]
```
jkapponms, line 36: I still prefer that instances be uniquely ID'd by their namespace, since that's enforced by k8s. Two instances could have the same name if their CRDs are installed into different namespaces.

geraldhumphries: Yeah, needs to be either cluster scoped with a namespace in the spec, or namespace scoped. This should work for now, but we might want to think about it more later. The change was to fix a bug rather than change any of the patterns, it should work exactly the same as before.

jaymberry (approved):

PR #369, opened 2022-10-12
Make sure snmp value encoding is part of protobuf contract.

splatch:
## Description
I started to move away from `SnmpValue` byte representation which might soon be unavailable on core side.
This is a contract update which aligns snmp-rpc handler on minion end with snmp client on core side. Client still uses `SnmpValueFactory`, however it can be dropped or replaced with another structure which is independent of actual snmp library.
Marking this PR for now as draft to open discussion on how this part of communication could evolve.

## Jira link(s)

## Checklist

### DIFF
--- a/shared-lib/horizon-ipc/ipc-grpc/ipc-grpc-contract/src/main/proto/opennms/snmp/snmp.proto
+++ b/shared-lib/horizon-ipc/ipc-grpc/ipc-grpc-contract/src/main/proto/opennms/snmp/snmp.proto
@@ -62,7 +62,14 @@ enum SnmpValueType {
 
 message SnmpValue {
   SnmpValueType type = 1;
-  bytes value = 2;
+  oneof value {
+    sint32 sint32 = 2;
+    sint64 sint64 = 3;
+    uint32 uint32 = 4;
+    uint64 uint64 = 5;
+    string string = 6;
+    bytes bytes = 7;
+  };
 }
 
 enum Version {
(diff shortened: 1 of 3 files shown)


mfrazier63 (approved):
LGTM

shared-lib/horizon-ipc/ipc-grpc/ipc-grpc-contract/src/main/proto/opennms/snmp/snmp.proto
```diff
@@ -62,7 +62,14 @@ enum SnmpValueType {
 
 message SnmpValue {
   SnmpValueType type = 1;
-  bytes value = 2;
+  oneof value {
```
cgorantla, line 65: Looks fine to me but still don't see much value in doing this explicitly in contract when this can be left to individual clients. May be I'm not seeing complete picture.

cgorantla: Got it. That makes sense now!

cgorantla (commented):
LGTM!   This may be needed for pushing Snmp metrics into Prometheus.  
Not sure what do with metrics that are non numeric.

splatch:
@cgorantla I think we can close this as you got these changes as part of one of features you worked on, am I right?

cgorantla:
> @cgorantla I think we can close this as you got these changes as part of one of features you worked on, am I right?

Yes, these changes are used as part of Snmp Collector changes.

We can close this.

PR #371, opened 2022-10-13
HS-474: adding an extra listener to kafka for local development

jkapponms:
## Description
Opening up an extra listener on the kafka deployment to allow for port-forwarded local development

## Jira link(s)
- https://issues.opennms.org/browse/HS-474

### DIFF
--- a/charts/opennms/templates/kafka/kafka-deployment.yaml
+++ b/charts/opennms/templates/kafka/kafka-deployment.yaml
@@ -19,15 +19,27 @@ spec:
         - name: {{ .Values.Kafka.ServiceName }}
           image: {{ .Values.Kafka.Image }}
           imagePullPolicy: {{ .Values.Kafka.ImagePullPolicy }}
-          command:
-            - /bin/sh
-            - -c
-            - "export CLUSTER_ID=$(bin/kafka-storage.sh random-uuid) && bin/kafka-storage.sh format -t $CLUSTER_ID -c config/kraft/server.properties && bin/kafka-server-start.sh config/kraft/server.properties --override advertised.listeners=${KAFKA_ADVERTISED_LISTENERS}"
           env:
-            - name: LOG_DIR
-              value: /tmp/logs
+            - name: KAFKA_ENABLE_KRAFT
+              value: "yes"
+            - name: KAFKA_CFG_BROKER_ID
+              value: "1"
+            - name: ALLOW_PLAINTEXT_LISTENER
+              value: "yes"
+            - name: KAFKA_CFG_PROCESS_ROLES
+              value: "broker,controller"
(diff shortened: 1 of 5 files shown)


operator/internal/handlers/keycloak.go
```diff
@@ -15,34 +15,34 @@ limitations under the License.
...
+
+	yaml.LoadYaml(filepath("keycloak/keycloak-initial-cred-secret.yaml"), values, &initialAdminSecret)
+	yaml.LoadYaml(filepath("keycloak/keycloak-realm-configmap.yaml"), values, &realmConfigmap)
+	yaml.LoadYaml(filepath("keycloak/keycloak-deployment.yaml"), values, &deployment)
+	yaml.LoadYaml(filepath("keycloak/keycloak-service.yaml"), values, &service)
+
+	h.Config = []client.Object{
+		&initialAdminSecret,
+		&realmConfigmap,
+		//&deployment,
```
geraldhumphries: Is this intentional?

operator/internal/model/values/kafka_values.go
```diff
@@ -15,5 +15,7 @@ limitations under the License.
 package values
 
 type KafkaValues struct {
-	ServiceValues `yaml:",inline"`
+	ServiceValues  `yaml:",inline"`
+	ControllerPort string `yaml:"ControllerPort"`
+	LocalPort      string `yaml:"LocalPort"`
```
geraldhumphries, line 20: Does it make sense for the operator to handle LocalPort, if it's something that would be used for development/dev mode?

geraldhumphries (approved):

geraldhumphries:
I think we need a strategy for "dev mode" vs "production mode" in the Helm chart. Things like the localhost listener should be disabled in prod, and enabled in dev mode. This also goes for other services, like debug ports on Java services.

charts/opennms/templates/kafka/kafka-deployment.yaml
```diff
@@ -19,15 +19,27 @@ spec:
...
+            - name: KAFKA_CFG_BROKER_ID
+              value: "1"
+            - name: ALLOW_PLAINTEXT_LISTENER
+              value: "yes"
+            - name: KAFKA_CFG_PROCESS_ROLES
+              value: "broker,controller"
+            - name: KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP
+              value: "CLIENT:PLAINTEXT,LOCALCLIENT:PLAINTEXT,CONTROLLER:PLAINTEXT"
+            - name: KAFKA_CFG_CONTROLLER_LISTENER_NAMES
+              value: "CONTROLLER"
```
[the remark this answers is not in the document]
jkapponms: Yes.

PR #376, opened 2022-10-13
Add custom env vars plugin to inject new values into osgi configurations

splatch:
## Description
Default plugin from Karaf can override only existing properties, while this one can inject new values. Properties are guessed from remaining part of env variable after stripping PID part from it.

PR ships it for both - core and minion through bootstrap bundle which should be activated earlier than boot features.

## Jira link(s)

## Checklist

### DIFF
--- a/shared-lib/osgi/osgi-config-plugin/src/main/java/org/opennms/horizon/shared/osgi/config/plugin/internal/EnvVarsConfigurationPlugin.java
+++ b/shared-lib/osgi/osgi-config-plugin/src/main/java/org/opennms/horizon/shared/osgi/config/plugin/internal/EnvVarsConfigurationPlugin.java
@@ -1,0 +1,58 @@
+package org.opennms.horizon.shared.osgi.config.plugin.internal;
+
+import java.util.Dictionary;
+import java.util.Map;
+import java.util.function.Supplier;
+import org.osgi.framework.Constants;
+import org.osgi.framework.ServiceReference;
+import org.osgi.service.cm.ConfigurationPlugin;
+
+/**
+ * Configuration plugin which attempts to map environment variables with configuration properties.
+ *
+ * Matching rules are environment variable of ORG_APACHE_KARAF_FEATURES_BOOT_FEATURE will fit:
+ * - pid org.apache.karaf.features, property boot.feature
+ * - pid org.apache.karaf, property features.boot.feature
+ *
+ * Properties are always lower case with dots as separator.
+ */
+public class EnvVarsConfigurationPlugin implements ConfigurationPlugin {
+
+    public static final int PLUGIN_RANKING = 510;
+    public static final String PLUGIN_ID = "org.opennms.horizon.shared.osgi.config.plugin";
+    private final Map<String, String> environment;
+
+    public EnvVarsConfigurationPlugin() {
+        this(System::getenv);
+    }
+
+    EnvVarsConfigurationPlugin(Supplier<Map<String, String>> environment) {
+        this.environment = environment.get();
+    }
+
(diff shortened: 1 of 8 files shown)


shared-lib/osgi/osgi-config-plugin/src/main/java/org/opennms/horizon/shared/osgi/config/plugin/internal/EnvVarsConfigurationPlugin.java
```diff
@@ -0,0 +1,58 @@
...
+import org.osgi.framework.Constants;
+import org.osgi.framework.ServiceReference;
+import org.osgi.service.cm.ConfigurationPlugin;
+
+/**
+ * Configuration plugin which attempts to map environment variables with configuration properties.
+ *
+ * Matching rules are environment variable of ORG_APACHE_KARAF_FEATURES_BOOT_FEATURE will fit:
+ * - pid org.apache.karaf.features, property boot.feature
+ * - pid org.apache.karaf, property features.boot.feature
```
mfrazier63, line 15: This seems ambiguous and confusing!

mfrazier63 (commented):
For some reason I thought we had decided NOT to do anything custom like this.

shared-lib/osgi/osgi-config-plugin/src/main/java/org/opennms/horizon/shared/osgi/config/plugin/internal/EnvVarsConfigurationPlugin.java
splatch: I know it is a bit of hit or miss, but assuming we have a consistent use of PIDs, and our primary driver is a PID, we can make our own convention over configuration.

geraldhumphries: Can we get rid of the ambiguity and make up some explicit rules about how they're applied?

e.g. `ORG_APACHE_KARAF_FEATURES_BOOT_FEATURE` always applies to the more specific PID, `org.apache.karaf.features, property boot.feature`. It does not affect any less-specific PIDs it would otherwise match, like `pid org.apache.karaf, property features.boot.feature`.

Is there anything we can use as a delimiter? Perhaps a double underscore if we need to be more explicit?

e.g. `ORG_APACHE_KARAF_FEATURES__BOOT_FEATURE` and 
`ORG_APACHE_KARAF__FEATURES_BOOT_FEATURE`.

splatch: Making `__` a separator between PID and property is fine. We might need to replicate a bit of logic from Karaf plugin which do recognize type of property, if it already exist. I will update PR to reflect that.

geraldhumphries:
I have been thinking about this plugin and how we would use env vars.

Once we get rid of `core`, our only Karaf component will be the Minion. The Minion is also the piece our clients will probably be configuring directly the most. We probably want to explicitly define and document possible environment variables for config.

This plugin allows for config of arbitrary settings, based on config PIDs. If we support arbitrary config and we ever need to change how configs are structured or rename PIDs, we risk breaking client deployments and preventing upgrades. Explicitly defining config settings and documenting them allows us to maintain control and prevents much of this risk.

What do you think?

splatch:
Closing it in favor of making it more explicit just for minion in near future.

PR #378, opened 2022-10-14
HS-433,475 Push metrics data to prometheus

yangli2014:
## Description

How to test:
Add a new localhost device from UI, the new device's ICMP metrics will show shortly. 

## Jira link(s)
- https://issues.opennms.org/browse/HS-433
- https://issues.opennms.org/browse/HS-475

## Checklist


charts/opennms/values.yaml
```diff
@@ -33,6 +33,18 @@ OpenNMS:
...
         Memory: 1Gi
+  MetricsProcessor:
+    ServiceName: opennms-metrics-processor
+    Image: opennms/horizon-stream-metrics-processor
+    ImagePullPolicy: IfNotPresent
+    Replicas: 1
+    Resources:
+      Limits:
+        Cpu: "2"
+        Memory: 2Gi
```
j-white, line 44: We can reduce this? 2Gi seems like overkill here.

metrics-processor/pom.xml
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
```
j-white, line 5: Is there a way for us to use the parent pom for new Sprint boot based services?

I believe this is the approach taken here: https://github.com/OpenNMS/horizon-stream/tree/develop/minion-gateway

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
```diff
@@ -0,0 +1,193 @@
...
+            List<TaskResult> resultList = results.getResultsList();
+            for (TaskResult oneResult : resultList) {
+                // TODO: update all returned metrics from the monitor
+                // TODO: support monitor results vs detector results
+                try {
+                    if (oneResult != null) {
+                        if (oneResult.hasMonitorResponse()) {
+                            MonitorResponse monitorResponse = oneResult.getMonitorResponse();
+
+                            switch (monitorResponse.getMonitorType()) {
```
j-white, line 79: I think we should move away from this service having to have specific knowledge of the different monitors and how to handle the time series for each of these.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
```diff
@@ -0,0 +1,193 @@
...
+                                    break;
+                            }
+                        } else if (oneResult.hasDetectorResponse()) {
+                            DetectorResponse detectorResponse = oneResult.getDetectorResponse();
+
+                            // TBD: how to process?
+                            log.info("Have detector response: task-id={}; detected={}", oneResult.getId(), detectorResponse.getDetected());
+                        }
+                    } else {
+                        log.warn("Task result appears to be missing the echo response details");
```
j-white, line 99: Message is not applicable. Would be good to include a copy of the payload in this case.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/ApplicationConfig.java
```diff
@@ -0,0 +1,97 @@
...
+@Configuration
+@EnableKafka
+@EnableKafkaStreams
+public class ApplicationConfig {
+    @Value("${spring.kafka.bootstrap-servers}")
+    private String bootstrapServers;
+    @Value("${spring.kafka.consumer.group-id}")
+    private String groupID;
+    @Value("${spring.kafka.template.default-topic}")
+    private String topic;
```
cgorantla, line 65: we may want to consume from multiple topics

metrics-processor/src/main/java/org/opennms/horizon/tsdata/ApplicationConfig.java
```diff
@@ -0,0 +1,97 @@
...
+        Map<String, Object> props = new HashMap<>();
+        props.put(APPLICATION_ID_CONFIG, "ts-data-stream");
+        props.put(BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        props.put(DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass());
+        props.put(DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass());
+        return new KafkaStreamsConfiguration(props);
+    }
+
+    @Bean
+    public KStream<String, TaskSetResults> initStream(StreamsBuilder streamsBuilder) {
```
cgorantla, line 80: This may be up for debate for I don't see any need for Kafka streams, kafka consumer seems to be sufficient here.  Would like to know your thoughts on this.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
cgorantla: Agree. We can update TaskSetResults to have metrics that match prometheus format.  Can handle this in different issue.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/ApplicationConfig.java
yangli2014: Let's make the TS data available for the UI team first. Then we can improve this.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
yangli2014: I will make the minion metrics work first, then we will have a better idea about how to create a generic message to handle all metrics.

metrics-processor/pom.xml
yangli2014: I will work on it.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/ApplicationConfig.java
yangli2014: I tried Kafka consumer before this, and it works and is simple. However, it is not easy to tell the difference from the performance perspective based on the dev environment.

geraldhumphries: One thing we should consider is that Kafka Streams is not supported by Azure Event Hub. This isn't a dealbreaker right now, we haven't decided for sure if we want to use Event Hub. But it is something to keep in mind.

yangli2014: It is good to know.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
yangli2014: This will be addressed in this story https://issues.opennms.org/browse/HS-487

yangli2014:
I pushed most of the PR comments review fix. The generic message will be handled in another story

cgorantla (approved):
LGTM!

PR #379, opened 2022-10-14
HS-483 - testing runner

jaymberry:
## Jira link(s)

## Checklist


.github/actions/external-it/action.yml
```diff
@@ -19,75 +19,90 @@ runs:
         #kubectl -h
 
         # Install dependencies
-        brew install operator-sdk
-        operator-sdk
-        which operator-sdk
+        #brew install operator-sdk
+        #operator-sdk
+        #which operator-sdk
```
geraldhumphries, line 24: We can remove these.

.github/actions/external-it/action.yml
```diff
@@ -19,75 +19,90 @@ runs:
...
         # Setup localhost DNS
         cat /etc/hosts
         sudo echo "127.0.0.1 $DOMAIN" | sudo tee -a /etc/hosts
 
-        sed -i.bak 's/IMAGE_TAG=latest/IMAGE_TAG=local/g' ./local-sample/config-run
-        sed -i.bak 's/image: opennms\/operator:latest/image: opennms\/operator:local-build/g' charts/opennms-operator/values.yaml
-        ./local-sample/run.sh cicd
+        #sed -i.bak 's/IMAGE_TAG=latest/IMAGE_TAG=local/g' ./local-sample/config-run
+        #sed -i.bak 's/image: opennms\/operator:latest/image: opennms\/operator:local-build/g' charts/opennms-operator/values.yaml
+        #./local-sample/run.sh cicd
```
geraldhumphries, line 40: Similar to above, we can remove commented-out lines.

.github/actions/external-it/action.yml
```diff
@@ -19,75 +19,90 @@ runs:
...
+        kind load docker-image opennms/horizon-stream-minion:local& 
+        kind load docker-image opennms/horizon-stream-minion-gateway:local& 
+        kind load docker-image opennms/horizon-stream-keycloak:local& 
+        kind load docker-image opennms/horizon-stream-grafana:local& 
+        kind load docker-image opennms/horizon-stream-ui:local& 
+        kind load docker-image opennms/horizon-stream-notification:local& 
+        kind load docker-image opennms/horizon-stream-rest-server:local& 
+
+        # Need to wait for the images to be loaded.
+        sleep 120
```
geraldhumphries, line 63: Doesn't `kind load` wait until the image is finished loading to exit? Do we need to sleep?

.github/actions/external-it/action.yml
```diff
@@ -19,75 +19,90 @@ runs:
...
+        kubectl -n local-instance patch deployments opennms-notifications  -p '{"spec": {"template": {"spec":{"containers":[{"name": "opennms-notifications", "image":"opennms/horizon-stream-rest-server:local"}]}}}}'
+        kubectl -n local-instance patch deployments opennms-minion         -p '{"spec": {"template": {"spec":{"containers":[{"name": "opennms-minion",        "image":"opennms/horizon-stream-rest-server:local"}]}}}}'
+        kubectl -n local-instance patch deployments opennms-minion-gateway -p '{"spec": {"template": {"spec":{"containers":[{"name": "opennms-minion-gateway","image":"opennms/horizon-stream-rest-server:local"}]}}}}'
+        kubectl -n local-instance patch deployments onms-keycloak          -p '{"spec": {"template": {"spec":{"containers":[{"name": "keycloak",              "image":"opennms/horizon-stream-keycloak:local"}]}}}}'
+        kubectl -n local-instance patch deployments grafana                -p '{"spec": {"template": {"spec":{"containers":[{"name": "grafana",               "image":"opennms/horizon-stream-grafana:local"}]}}}}'
```
geraldhumphries, line 73: Shouldn't we do this in the CRD? The operator will overwrite these changes otherwise.

install-local.sh
```diff
@@ -1,75 +1,24 @@
...
-if [ $? -ne 0 ]; then exit; fi
-kubectl -n local-instance create secret tls tls-cert-wildcard --cert tmp/server.crt --key tmp/server.key
-if [ $? -ne 0 ]; then exit; fi
+bash scripts/add-local-ssl-cert.sh
 
 kubectl config set-context --current --namespace=local-instance
 
-kubectl get pods -w
+# This is the last pod to run, if ready, then give back the terminal session.
+while [ -v $(kubectl get pods -n local-instance -l=app.kubernetes.io/component='controller-local-instance' -o jsonpath='{.items[*].status.containerStatuses[0].ready}')  ]; do
```
geraldhumphries, line 21: Is it possible to wait until the CRD instance is ready? The operator will mark it as ready when all its resources are ready.

operator/scripts/deploy-horizon-stream.sh
```diff
@@ -11,4 +11,8 @@ bash scripts/add-local-ssl-cert.sh
 
 kubectl config set-context --current --namespace=local-instance
 
-kubectl get pods -w
+# This is the last pod to run, if ready, then give back the terminal session.
+while [ -v $(kubectl get pods -n local-instance -l=app.kubernetes.io/component='controller-local-instance' -o jsonpath='{.items[*].status.containerStatuses[0].ready}')  ]; do
```
geraldhumphries, line 15: Same as above regarding CRD instance state,

.github/actions/external-it/action.yml
```diff
@@ -19,75 +19,90 @@ runs:
...
+        #java -jar "external-horizon-stream-it/target/external-horizon-stream-it-${PROJECT_VERSION}.jar"
+
+        ## If fail, exit which kills the ci-cd workflow.
+        ##for i in $(jq '.[0].elements[].steps[].result.status' external-it/external-horizon-stream-it/cucumber.reports/cucumber-report.json);do
+        ##  if [[ $i != '"passed"' ]];then
+        ##    exit;
+        ##  fi
+        ##done
+
+        kind delete clusters kind
```
geraldhumphries, line 106: Maybe move this to another "cleanup" step?

geraldhumphries (changes requested):

.github/actions/external-it/action.yml
jaymberry: Yes, but I am running them all at once in the background, so need to set a time limit before moving onto next step.

.github/actions/external-it/action.yml
jaymberry: Right, ok, I can re-work this a little, good point.

operator/scripts/deploy-horizon-stream.sh
jaymberry: This is the only way I could accurately wait until it is ready.

.github/actions/external-it/action.yml
jaymberry: Removed

.github/actions/external-it/action.yml
jaymberry: Done

operator/scripts/deploy-horizon-stream.sh
geraldhumphries: If you are using the operator then it should be doing that job for you, is there an issue with it? Note that "deployOnly: true" in the CRD will disable the operator waiting until deployments are ready, and it will mark the instance as ready right away.

jaymberry: No, it does not. It just starts the build and then gives the session back while it builds the local-instance env. We need to add additional steps to check.

geraldhumphries (approved):

PR #388, opened 2022-10-19
HS-485_487 minion metrics to kafka

yangli2014:
## Description

Make sure the minion uptime and response time (not correct data for now) showing in the UI 

## Jira link(s)
- https://issues.opennms.org/browse/HS-485
- https://issues.opennms.org/browse/HS-487

## Checklist

### DIFF
--- a/install-local.sh
+++ b/install-local.sh
@@ -1,7 +1,17 @@
 #!/usr/bin/env bash
 #use this script to install a basic version of OpenNMS Horizon Stream locally
 
-cd operator/
+
+operator_run () {
+  bash scripts/add-local-ssl-cert.sh
+  kubectl config set-context --current --namespace=local-instance
+
+  # This is the last pod to run, if ready, then give back the terminal session.
+  while [ -v $(kubectl get pods -n local-instance -l=app.kubernetes.io/component='controller-local-instance' -o jsonpath='{.items[*].status.containerStatuses[0].ready}')  ]; do 
+    echo "not-ready"
+    sleep 60
+  done
+}
 
 # This determines whether or not to import custom images or not.
 HELP='Need to pass "local" parameter to script'
@@ -10,6 +20,7 @@ if [ -z "$1" ]; then
 else
   if [ $1 == "local" ]; then
 
+    cd operator/
     bash scripts/create-kind-cluster.sh
 
     echo
@@ -19,9 +30,12 @@ else
     if [ $? -ne 0 ]; then exit; fi
 
     bash scripts/create-instance.sh
-  
+ 
+    operator_run
+
   elif [ "$1" == "custom-images" ]; then
 
+    cd operator/
     bash scripts/create-kind-cluster.sh
 
     # Will add a kind-registry here at some point, see .github/ for sample script.
@@ -51,8 +65,12 @@ else
     echo
     kubectl apply -f ../install-local-onms-instance.yaml
     if [ $? -ne 0 ]; then exit; fi
+ 
+    operator_run
 
   elif [ $1 == "existing-k8s" ]; then
+
+    cd operator/
 
     echo
     echo ________________Installing Operator________________
@@ -61,7 +79,16 @@ else
     if [ $? -ne 0 ]; then exit; fi
 
     bash scripts/create-instance.sh
-  
+ 
+    operator_run
+ 
+  elif [ $1 == "existing-k8s-no-op" ]; then
+
+    echo
+    echo ____________Installing HS Instance______________
+    echo
+    helm upgrade -i horizon-stream charts/opennms -f ./install-local-onms-instance.yaml --namespace hs-instance --create-namespace
+ 
   else
     echo "$HELP"
   fi
@@ -82,4 +109,3 @@ while [ -v $(kubectl get pods -n local-instance -l=app.kubernetes.io/component='
   fi
   sleep 60
 done
-


shared-lib/protobuf/src/main/proto/metrics.proto
```diff
@@ -0,0 +1,44 @@
...
+package opennms.metrics;
+option java_multiple_files = true;
+option java_package = "org.opennms.horizon.stream.metrics";
+
+message MetricsMessage {
+  string name = 1;
+  string desc = 2;
+  string unit = 3;
+  map<string, string> labels = 4;
+  int64 value = 5;
```
cgorantla, line 41: we should use `double` for the value

cgorantla (commented):
LGTM overall, need to update metric value to double

shared-lib/protobuf/src/main/proto/metrics.proto
yangli2014: I figured out this issue and will include it in the next commit. I will combine the HS-487 in this PR

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutionResultProcessorImpl.java
```diff
@@ -35,33 +55,61 @@ public TaskExecutionResultProcessorImpl(SyncDispatcher<TaskSetResults> taskSetSi
...
+                .setDesc(METRICS_DESC_SNMP_UP)
+                .setUnit(METRICS_UNIT_SNMP_UP)
+                .setValue(calculateSNMPUpTime(response));
+            listBuild.addMetricsList(messageBuilder.build());
+        }
+        return listBuild.build();
+    }
 
-        return taskSetResults;
+    private Long calculateSNMPUpTime(MonitorResponse response) {
```
cgorantla, line 98: I'm not sure we would eventually need uptime metric for all of the monitors. For now, we can keep this, this may be removed later.

cgorantla (commented):
Now that we are consuming metrics directly from Kafka, we may not need any changes in Core.

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutionResultProcessorImpl.java
yangli2014: if the uptime is not for all monitors we can implement some conditions in the future.

minion/minion-taskset/taskset-ipc/src/main/java/org/opennms/horizon/minion/taskset/ipc/internal/MetricsListSinkModule.java
```diff
@@ -8,10 +8,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class TaskSetResultsSinkModule implements SinkModule<MetricsList, MetricsList> {
+public class MetricsListSinkModule implements SinkModule<MetricsList, MetricsList> {
```
cgorantla, line 11: May be MetricsSinkModule name would be better!

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutionResultProcessorImpl.java
```diff
@@ -35,33 +55,61 @@ public TaskExecutionResultProcessorImpl(SyncDispatcher<TaskSetResults> taskSetSi
...
-                .setMonitorResponse(monitorResponse)
-                .setLocation(identity.getLocation())
-                .setSystemId(identity.getId())
-                .build();
-
-        TaskSetResults taskSetResults =
-            TaskSetResults.newBuilder()
-                .addResults(taskResult)
-                .build();
+    private MetricsList resultToMetrics(String id, ServiceMonitorResponse result) {
```
artnaseef, line 66: Talking about this with Chandra, I have some thoughts:

- Logic in minions will be distributed - across many clients, locations, and instances at a location
- Changes to this logic requires all of those to be updated to take effect
- Exposing metrics internals out feels unnecessary and may lead to unintended problems in the future (e.g. if a third-party updates their minion to produce new metrics, change metric names,, or even just add labels to existing metrics)
- Having the minion send back just the "observed results" of the task set is cleaner in that regard, and enables other processing based on those results
- We could add another processor, running in the cloud, that receives the `ServiceMonitorResponse` and generates the metrics from there
- This conversion appears to lose the Status value (UP, DOWN, etc) from the `ServiceMonitorResponse`

yangli2014: That is a good point: I think minion-gateway will be the place to generate metrics from task results. We can discuss this with the team.

j-white: @artnaseef agreed on your points
@yangli2014  I don't think the `minion-gateway` should be responsible for acting on the task results, probably best to do this elsewhere

up/down is lost currently, the approach with the metrics the way they are now was intended to be temporary in order to show "something" in the UI, we need to rethink how we want this to work

PR #398, opened 2022-10-21
[HS-276] PoC of development environment injection into Helm chart using Tilt

geraldhumphries:
## Description

Quite a lot of code needed to do this. Maybe a Helm-based solution would be better - allow directly passing in configuration of certain ports and env vars.

## Jira link(s)

## Checklist


Tiltfile
```diff
@@ -1,58 +1,16 @@
...
-    )
-
 # Deployment #
-k8s_yaml(
-    helm(
-        'charts/opennms',
-        values=['./skaffold-helm-values.yaml'],
-    )
-)
+decoded_chart = decode_yaml_stream(helm(
```
geraldhumphries, line 10: This decodes the chart into a list containing all of the K8s deployments as dicts. It's pretty important for the Tilt functions that do the injects, but I want to refactor it.

geraldhumphries:
This isn't a good solution or maintainable, I think we should go with a flag in the Helm chart.

PR #407, opened 2022-10-26
HS-514 Convert task result to metric in generic way

yangli2014:
## Description

After the local environment is launched, add an SNMP device on the UI page and then verify ICMP_msec and SNMP_msec metrics data collected in the Prometheus server
## Jira link(s)
- https://issues.opennms.org/browse/HS-514

## Checklist


metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
```diff
@@ -107,87 +90,31 @@ public void consume(byte[] data) {
...
-        // Also update the gauges for additional metrics from the monitor
-        for (Map.Entry<String, Double> oneMetric : metrics.entrySet()) {
-            try {
-                Gauge dynamicMetricGauge = gaugeFactory.lookupGauge(oneMetric.getKey());
-                dynamicMetricGauge.labels(labelValues).set(oneMetric.getValue());
-            } catch (Exception exc) {
-                log.warn("Failed to record metric: metric-name={}; value={}", oneMetric.getKey(), oneMetric.getValue(), exc);
-            }
+        Gauge gauge = Gauge.build()
+            .name(response.getMonitorType().name())
```
cgorantla, line 102: I think we want `response_time` as metric name and monitor type added as label.

cgorantla, line 102: Also as per prometheus metric naming convention, we can append `_msec` to the name.  I'm okay with omitting it.

shared-lib/protobuf/src/main/proto/taskSet.proto
```diff
@@ -73,6 +73,8 @@ message TaskResult {
     MonitorResponse monitor_response = 7;
     DetectorResponse detector_response = 8;
   }
+
+  map<string, string> extra_labels = 9;
```
cgorantla, line 77: I guess we can add when we need this. Don't see any use for it now!

cgorantla (changes requested):

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
yangli2014: it automatically append the _ and unit at the end of the name

yangli2014: I use SNMP_response_time_msec as the metrics name.

j-white: Agreed. I think it makes more sense to have the monitor as a label here, rather than prefixing the metric name.

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
```diff
@@ -107,87 +90,28 @@ public void consume(byte[] data) {
...
+            .help(response.getMonitorType().name() + " round trip response Time")
+            .unit(METRICS_RESPONSE_UNIT)
+            .labelNames(labels.keySet().toArray(new String[0]))
+            .register(collectorRegistry);
+        gauge.labels(labels.values().toArray(new String[0])).set(response.getResponseTimeMs());
+
+        if(response.getMetricsMap()!=null) {
+            response.getMetricsMap().forEach((k, v)-> {
+                Gauge extGauge = Gauge.build()
+                    .name(k)
```
j-white, line 109: Include a prefix here, such as "monitor_" so that these don't collide with other metrics. Similar to above, include the monitor as a label as well, so we can associate this back with the monitor.

yangli2014:
@j-white @cgorantla Let me know if the following change is good enough
Screenshot 2022-10-27 at 10 19 41 AM

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
```diff
@@ -107,87 +98,43 @@ public void consume(byte[] data) {
...
-        for (Map.Entry<String, Double> oneMetric : metrics.entrySet()) {
-            try {
-                Gauge dynamicMetricGauge = gaugeFactory.lookupGauge(oneMetric.getKey());
-                dynamicMetricGauge.labels(labelValues).set(oneMetric.getValue());
-            } catch (Exception exc) {
-                log.warn("Failed to record metric: metric-name={}; value={}", oneMetric.getKey(), oneMetric.getValue(), exc);
-            }
+        if(response.getMetricsMap()!=null) {
+            response.getMetricsMap().forEach((k, v)-> {
+                Gauge extGauge = getGaugeFromName(METRICS_NAME_PREFIX_MONITOR + k, false);
```
cgorantla, line 113: I think we want `monitor_` prefix only for extra metrics that are coming from the response not for the response_time.

Also as you pointed before, we don't need to add unit in the name.

yangli2014: So the type label will be changed to monitor_type="ICMP"?

cgorantla: monitor="ICMP" should be good enough!

yangli2014: updated. thanks

metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java
```diff
@@ -107,87 +98,43 @@ public void consume(byte[] data) {
...
-                .boxed()
-                .collect(Collectors.toMap(i -> GaugeFactory.LABEL_NAMES[i], i -> labelValues[i]));
-
-        pushAdapter.pushMetrics(gaugeFactory.getCollectorRegistry(), groupingKey);
+    private Gauge getGaugeFromName(String name, boolean withDesc) {
+        return gauges.compute(name, (key, gauge) -> {
+            if(gauge != null) {
+                return gauge;
+            }
+            if(withDesc) {
```
cgorantla, line 125: nit: looks like this file is not formatted!

cgorantla (approved):
LGTM!

PR #418, opened 2022-11-02
Use GRPC between core and Minion Gateway to publish task sets.

artnaseef:
## Jira link(s)

## Checklist


cgorantla:
@artnaseef  Do we need to create another gRPC call for  Echo RPC or It will continue to be Ignite based ?

artnaseef:
> @artnaseef Do we need to create another gRPC call for Echo RPC or will it continue to be Ignite based ?

We need another GRPC call.  Working on that separately now.

PR #420, opened 2022-11-03
HS-470: Add Node to Inventory

james8hut:
## Description
Add Node to Inventory.
Also tests for MonitoringServiceTypes which I seem to have missed from last PR.

## Jira link(s)
- https://issues.opennms.org/browse/HS-470

## Flagged for review
n/a

## Checklist

### DIFF
--- a/inventory/src/main/java/org/opennms/horizon/inventory/mapper/NodeMapper.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/mapper/NodeMapper.java
@@ -1,0 +1,12 @@
+package org.opennms.horizon.inventory.mapper;
+
+import org.mapstruct.Mapper;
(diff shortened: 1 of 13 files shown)


cgorantla (commented):
We can move all these simple tests into one IT so that we don't need to pull docker image for each test.

Should these tests be named as IT ?  Are these tests running in build pipeline ?

james8hut:
> We can move all these simple tests into one IT so that we don't need to pull docker image for each test.
Can do. But if we remove REST and use gRPC instead, tests will all change anyway?
Just wanted to put some simple tests in for now to prove Service code down all works.
> 
> Should these tests be named as IT ? 
Can do, with above proviso.
> Are these tests running in build pipeline ?
Not yet. I believe there's another change needed for that.

PR #421, opened 2022-11-04
HS-471 gRPC service in inventory service

yangli2014:
## Description

After inventory starts, make sure the log showing gRPC server started with port 6565 
## Jira link(s)
- https://issues.opennms.org/browse/HS-471

## Checklist


inventory/src/main/java/org/opennms/horizon/inventory/model/MonitoringLocation.java
```diff
@@ -1,13 +1,18 @@
...
 import lombok.RequiredArgsConstructor;
+
 import org.hibernate.annotations.Type;
 
-import javax.persistence.*;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
```
james8hut, line 11: This change was made in my latest PR. So maybe a merge conflict as I think that should be in develop already.
Maybe you want to get latest develop branch?

shared-lib/inventory/src/main/proto/opennms/inventory.proto
```diff
@@ -38,3 +39,13 @@ message MonitoringLocationDTO {
   string tenant_id = 2;
   string location = 3;
 }
+
+service LocationService {
```
james8hut, line 43: Question about the rename. Should all tables appear in the same proto file? I've been adding new tables in different proto files.

shared-lib/inventory/src/main/proto/opennms/inventory.proto
```diff
@@ -38,3 +39,13 @@ message MonitoringLocationDTO {
...
+  rpc createLocation(MonitoringLocationDTO) returns (MonitoringLocationDTO) {};
+  rpc UpdateLocation(MonitoringLocationDTO) returns (MonitoringLocationDTO) {};
```
james8hut, line 45: should UpdateLocation be updateLocation in line with other naming conventions?

james8hut (changes requested):

inventory/src/main/java/org/opennms/horizon/inventory/service/LocationGrpcService.java
```diff
@@ -0,0 +1,127 @@
...
+        MonitoringLocation location = locationRepo.findByLocation(request.getLocation());
+        if(location != null) {
+            Status status = Status.newBuilder()
+                .setCode(Code.ALREADY_EXISTS.getNumber())
+                .setMessage("The location already exist")
+                .build();
+            responseObserver.onError(StatusProto.toStatusRuntimeException(status));
+        } else {
+            //if id exist it will update the location
+            responseObserver.onNext(mapper.modelToDTO(locationRepo.save(mapper.dtoToModel(request))));
```
james8hut, line 72: Query on this. If the id exists, should we throw an error? Just for my own knowledge, does the rest of the application do an update in this scenario or throw an error?

shared-lib/inventory/src/main/proto/opennms/inventory.proto
cgorantla, line 43: We don't need to create all CRUD operations, currently BFF only need all locations and locations by location name, that's all we need to expose.

shared-lib/inventory/src/main/proto/opennms/inventory.proto
```diff
@@ -38,3 +39,13 @@ message MonitoringLocationDTO {
...
+  rpc listLocations(google.protobuf.Empty) returns (stream MonitoringLocationDTO) {};
```
cgorantla, line 46: stream in `grpc` is different, it's continuous flow of data.

inventory/pom.xml
```diff
@@ -91,19 +91,39 @@
             <artifactId>lombok-mapstruct-binding</artifactId>
             <version>${lombok.binding.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-core</artifactId>
+            <version>${grpc.version}</version>
+        </dependency>
+        <dependency>
```
cgorantla, line 99: IMO, we shouldn't bring unofficial spring boot starters.
We can re-use config like `GrpcIpcServerConfig` and `GrpcIpcUtils` for configuring server.

cgorantla (changes requested):

inventory/src/main/java/org/opennms/horizon/inventory/model/MonitoringLocation.java
yangli2014: sure

shared-lib/inventory/src/main/proto/opennms/inventory.proto
yangli2014: I just discussed this with Chandra, we keep two files. One for nodes and another one for monitoring.

shared-lib/inventory/src/main/proto/opennms/inventory.proto
yangli2014: good catch

inventory/src/main/java/org/opennms/horizon/inventory/service/LocationGrpcService.java
yangli2014: In classic, the location name is the primary key if I remember correctly. Also, I am unsure if we have a use case to change the location name.

inventory/src/main/java/org/opennms/horizon/inventory/grpc/GrpcConfig.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final long DEFAULT_MAX_MESSAGE_SIZE = 100 * ( 1024 * 1024 );
+    private static final int DEFAULT_GRPC_PORT = 8990;
+
+    @Value("${grpc.server.port:" + DEFAULT_GRPC_PORT +"}")
+    private int port;
+    @Value("${grpc.server.max.message.size: " + DEFAULT_MAX_MESSAGE_SIZE +"}")
+    private int maxMessageSize;
+    private final MonitoringLocationRepository locationRepo;
+
+    public GrpcConfig(MonitoringLocationRepository locationRepo) {
```
thomasbigger584, line 48: for consistency we can use `@RequiredArgsConstructor` instead of this constructor. This will auto generate with Lombok the constructor if the fields are final, thus giving dependency injection with less code. Same with other classes in the PR.

inventory/src/main/java/org/opennms/horizon/inventory/service/InventoryGrpcService.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.rpc.Status;
+
+import io.grpc.protobuf.StatusProto;
+import io.grpc.stub.StreamObserver;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class InventoryGrpcService extends InventoryServiceGrpc.InventoryServiceImplBase {
+    private final MonitoringLocationRepository locationRepo;
+    private final MonitoringLocationMapper mapper = Mappers.getMapper(MonitoringLocationMapper.class);
```
thomasbigger584, line 50: this can be constructor injected as its a spring bean, instead of statically referencing it. Same with other occurences in the PR.

inventory/src/main/java/org/opennms/horizon/inventory/repository/MonitoringLocationRepository.java
```diff
@@ -2,8 +2,11 @@
 
 import org.opennms.horizon.inventory.model.MonitoringLocation;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface MonitoringLocationRepository extends JpaRepository<MonitoringLocation, Long> {
+    @Query("select location from MonitoringLocation location where location.location =?1")
+    MonitoringLocation findByLocation(String location);
```
thomasbigger584, line 11: If this returns `Optional<MonitoringLocation>` then we can have cleaner null checking in the service with the `.isPresent()` method

inventory/src/main/java/org/opennms/horizon/inventory/service/InventoryGrpcService.java
james8hut: If you merge with develop, you'll see the Mapper has changed slightly. @Mapper(componentModel = "spring"), and you'll also see that Mappers.getMapper is no longer used elsewhere.

inventory/src/main/java/org/opennms/horizon/inventory/grpc/GrpcConfig.java
cgorantla: My 2 cents: I'm not a fan of lombok. When we get to Java 17, we don't need `lombok` in many places.  Records are much better.

inventory/src/main/resources/application.yml
```diff
@@ -14,3 +14,8 @@ spring:
 
   liquibase:
     change-log: db/changelog/changelog.xml
+
+grpc:
+  server:
+    port: 6565
+    max.message.size: 104857600
```
cgorantla, line 21: `max.message.size` may not be needed here. It was needed for IPC but no need for it on API.

shared-lib/inventory/src/main/proto/monitoring.proto
```diff
@@ -38,3 +39,8 @@ message MonitoringLocationDTO {
   string tenant_id = 2;
   string location = 3;
 }
+
+service MonitoringService {
+  rpc listLocations(google.protobuf.Empty) returns (stream MonitoringLocationDTO) {};
```
cgorantla, line 44: we need a list of MonitoringLocationDTO here not a Stream.

inventory/src/main/java/org/opennms/horizon/inventory/grpc/GrpcConfig.java
cgorantla: Once we get rid of Core, we can move to Java 17. So in a month or two.

shared-lib/inventory/src/main/proto/monitoring.proto
yangli2014: If the server uses the stream, the client still can get a list. I will keep it for now. If there is any issue when ingrate with BFF, we can change it.

cgorantla: As we chatted, this needs to be list but I'm okay to change when you update the client.

inventory/src/test/java/org/opennms/horizon/inventory/grpc/GrpcIntegrationTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @AfterEach
+    public void cleanUp(){
+        repo.deleteAll();
+    }
+
+    @Test
+    public void testListLocations () {
+        Iterator<MonitoringLocationDTO> result = serviceStub.listLocations(Empty.newBuilder().build());
```
cgorantla, line 129: This test may be unstable depending on the timing whether Iterator gets 1 or 2 elements, it's better to update API to return list

cgorantla (approved):

PR #426, opened 2022-11-07
HS-470: Add MonitoredService into Inventory

james8hut:
## Description
Add MonitoredService into Inventory

## Jira link(s)
- https://issues.opennms.org/browse/HS-470

## Flagged for review
none

## Checklist

### DIFF
--- a/inventory/src/main/java/org/opennms/horizon/inventory/rest/MonitoredServiceRest.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/rest/MonitoredServiceRest.java
@@ -1,0 +1,62 @@
+package org.opennms.horizon.inventory.rest;
+
+import lombok.RequiredArgsConstructor;
+import org.opennms.horizon.inventory.dto.MonitoredServiceDTO;
+import org.opennms.horizon.inventory.service.MonitoredServiceService;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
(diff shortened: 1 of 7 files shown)


inventory/src/main/java/org/opennms/horizon/inventory/rest/MonitoredServiceRest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+import java.util.Optional;
+
+@RestController
+@RequestMapping("/inventory/services")
```
thomasbigger584, line 20: should `/inventory` be a spring context path instead if its consistently the same for each rest endpoint in the inventory service?

james8hut: Possibly, but the REST layer is probably going away.

james8hut: Hmm, after trying this, I would probably prefer not. If you do it in application.yml, then it needs to be in both src/main/resources and src/test/resources. As such, if there is a different value in main than test, your tests don't really do their job. (Just placing it in main causes the Junit tests to fail.)

https://www.baeldung.com/spring-boot-context-path

Possibly the most workable place to put it is the main method. But that doesn't strike me as a great place to have it. Seems hidden.

thomasbigger584: Another option is to use a test profile in your tests, such that the two application.ymls get merged

PR #435, opened 2022-11-08
[HS-530] Add coverage reports with JaCoCo and enabled SonarCloud coverage

geraldhumphries:
## Description

- Adds a "coverage" profile to Java projects that enables coverage reporting with jacoco-maven-plugin
- Use `verify` phase in pipelines
- Use `working-directory` instead of cd or -f flag
- Slightly refactor minion-gateway so jib goal can be run against root pom

## Jira link(s)
- https://issues.opennms.org/browse/HS-530

## Flagged for review

- Make sure the use of profiles and pluginManagement makes sense
- minion-gateway refactor for the jib plugin. You used to need to do `mvn package && mvn jib:build -f main` to get it working properly so the plugin did not execute on any of the modules except main. Now jib is skipped by default in the parent pom, and reenabled in the main module. `mvn package jib:build` can now be run from the root and works properly.

## Checklist


geraldhumphries:
Looks like SonarCloud is picking up the new reports!

geraldhumphries:
> This enabled integration tests on the inventory service, and now there appears to be a test failure.

Something weird happened with that pipeline run. The test that failed wasn't actually in the code for the commit that was tested. This might actually be a bug in the self-hosted runner or something.

artnaseef (commented):
Looks good to me.  May want to consider using `mvn install` + `mvn -pl... jib...` instead of the jibs plugins with skip.

PR #453, opened 2022-11-10
Adjust keycloak helm chart to avoid troubles with routing.

splatch:
I did verify Keycloak configuration and 19.0.3 (.2 probably too) have a bug related to detection of https from actual configuration. There are few options which are available the most stable seem to be HOSTNAME_URL which allows to pass https, host and port all together. The https detection with hostname alone doesn't work and needs KC_HOSTNAME_STRICT_HTTPS to be set to false, which is undocumented feature.

Shipping it for now, as there is a pending issue in Keycloak itself: keycloak/keycloak#15287, which awaits resolution in release later than we use.

## Jira link(s)

## Checklist

### DIFF
--- a/charts/opennms/templates/keycloak/keycloak-deployment.yaml
+++ b/charts/opennms/templates/keycloak/keycloak-deployment.yaml
@@ -51,12 +51,8 @@ spec:
               value: kubernetes
             - name: KC_CACHE
               value: ispn
-            - name: KC_HOSTNAME
-              value: "{{ .Values.Host }}"
-            {{- if .Values.Keycloak.HostnamePort }}
-            - name: KC_HOSTNAME_PORT
-              value: "{{ .Values.Keycloak.HostnamePort }}"
-            {{- end }}
+            - name: KC_HOSTNAME_URL
+              value: "{{ .Values.Keycloak.HostnameUrl }}"
             {{- if .Values.Keycloak.HostnameAdminUrl }}
             - name: KC_HOSTNAME_ADMIN_URL
               value: "{{ .Values.Keycloak.HostnameAdminUrl }}"
@@ -75,22 +71,26 @@ spec:
                 secretKeyRef:
                   name: postgres
                   key: keycloakPwd
+            {{- if .Values.Keycloak.AdminUsername }}
             - name: KEYCLOAK_ADMIN
               valueFrom:
                 secretKeyRef:
                   key: username
                   name: {{ .Values.Keycloak.ServiceName }}-initial-admin
+            {{- end }}
+            {{- if .Values.Keycloak.AdminPassword }}
             - name: KEYCLOAK_ADMIN_PASSWORD
               valueFrom:
                 secretKeyRef:
                   key: password
                   name: {{ .Values.Keycloak.ServiceName }}-initial-admin
-            - name: KC_HOSTNAME_STRICT_BACKCHANNEL
+            {{- end }}
+            - name: KC_HOSTNAME_STRICT
+              value: "false"
+            - name: KC_HOSTNAME_STRICT_HTTPS
               value: "false"
             - name: KC_HTTP_ENABLED
               value: "true"
-            - name: KC_HTTP_RELATIVE_PATH
-              value: "{{ .Values.Keycloak.Path }}"
             - name: KC_PROXY
               value: passthrough
             - name: jgroups.dns.query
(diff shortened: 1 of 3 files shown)


charts/opennms/values.yaml
```diff
@@ -150,8 +150,8 @@ Keycloak:
   ImagePullPolicy: IfNotPresent
   Port: 8080
   HttpsPort: 8443
-  HostnamePort: ~
-  HostnameAdminUrl: ~
+  HostnameUrl: http://localhost:8123/auth
+  HostnameAdminUrl: http://localhost:8123/auth
```
geraldhumphries, line 154: These values are specific to our local dev setup, and should not be in the chart's defaults. Instead, add them in the `tilt-helm-values.yaml` at the root of the repo.

geraldhumphries (changes requested):

charts/opennms/templates/keycloak/keycloak-deployment.yaml
```diff
@@ -75,21 +71,27 @@ spec:
                 secretKeyRef:
                   name: postgres
                   key: keycloakPwd
+            {{- if .Values.Keycloak.AdminUsername }}
```
geraldhumphries, line 74: What is the impact if AdminUsername/AdminPassword are not set? Are there cases where we do not want them?

charts/opennms/values.yaml
splatch: Roger!

charts/opennms/templates/keycloak/keycloak-deployment.yaml
```diff
@@ -75,21 +71,27 @@ spec:
...
-            - name: KC_HOSTNAME_STRICT_BACKCHANNEL
+            {{- end }}
+            - name: KC_HOSTNAME_STRICT
+              value: "false"
+            - name: KC_HOSTNAME_STRICT_HTTPS
               value: "false"
             - name: KC_HTTP_ENABLED
               value: "true"
-            - name: KC_HTTP_RELATIVE_PATH
+            - name: KC_HOSTNAME_PATH
```
geraldhumphries, line 94: I know that KC_HOSTNAME_URL will override both KC_HOSTNAME and KC_HOSTNAME_PORT, and I was under the impression that KC_HOSTNAME_PATH works the same way - the path in KC_HOSTNAME_URL (e.g. "auth" from `http://localhost:8123/auth`) will take precedence. Since KC_HOSTNAME_URL is always set, can you verify if KC_HOSTNAME_PATH is actually necessary?

I also remember there being a big difference between how KC_HTTP_RELATIVE_PATH and KC_HOSTNAME_PATH behave, and I had picked KC_HTTP_RELATIVE_PATH for a reason. I think it was related to either the admin console or Grafana's SSO. Can you check that the following are all working with these changes through Tilt?

1. Core service's connection to Keycloak, when it checks roles (Keycloak's backchannel communication)
2. Access to Keycloak's admin console at `http://localhost:8123/auth` without incorrect urls for CSS/JS/form submissions (Keycloak's "adminUrl" related settings)
3. Grafana's SSO at `http://localhost:8123/grafana` (Keycloak's frontend URL settings, I believe). This may actually be broken on develop, it is redirecting to port 80 when I tried it!

charts/opennms/templates/keycloak/keycloak-deployment.yaml
splatch: This is a workaround for cases when database has been populated. Historically it was possible to add admin account by 1) environment variables 2) a magical `keycloak-add-user.json` file 3) realm import, although I am not certain if its possible for master realm. It might have changed a bit with quarkus distribution, but its still found in 19.0.x code. Alternative path could be ie. sidecar with json file.

charts/opennms/templates/keycloak/keycloak-deployment.yaml
splatch: My case was impossibility to login into admin console and ui due to port/https clash. Looking at code I think relative path might be still used to link a welcome page to admin page. You're right that we do not need `HOSTNAME_PATH` because it is detectable from entire URL. Its recommended use is only we rewrite ie. `/login` on ingress to `/` on keycloak. I'll remove it.

Going to check grafana and see if it survives.

splatch: Grafana seems to be giving wrong redirect_uri to keycloak. Grafana declares http://localhost/grafana as intended redirect which is wrong. So it is not catching 8123 thing properly as well.

geraldhumphries: Grafana's redirect_uri comes from the `rool_url` option under `[server]`. It is missing the port number. This config worked during local dev.

image

splatch:
Looking at direction in which I began moving I think this PR needs to be closed. The Keycloak URI handling is far from being perfect and issues are crossing several aspects. The ingress blind of port forwards in tilt and Keycloak making its own logic to navigate all this.
I found out that we can get Grafana redirects working locally if we set ` domain = {{ .Values.Host }}:{{ .Values.Ingress.HttpPort}}`, however this forces http which is fine for development but not going to fly with production deployments.

PR #457, opened 2022-11-14
HS-473: Add gRPC service for creating a device

james8hut:
## Description
Add gRPC service for creating a device

This is the first half of the HS-473 story. In a separate story will complete: add a gRPC client in BFF that integrates with addDevice in DeviceService graphQL Service.

## Jira link(s)
- https://issues.opennms.org/browse/HS-473

## Flagged for review
n/a

## Checklist


inventory/src/test/java/org/opennms/horizon/inventory/grpc/DeviceGrpcIntegrationTest.java
```diff
@@ -0,0 +1,147 @@
...
+        .withDatabaseName("inventory").withUsername("inventory")
+        .withPassword("password").withExposedPorts(5432);
+
+    @DynamicPropertySource
+    private static void registerDatasourceProperties(DynamicPropertyRegistry registry) {
+        registry.add("spring.datasource.url",
+            () -> String.format("jdbc:postgresql://localhost:%d/%s", postgres.getFirstMappedPort(), postgres.getDatabaseName()));
+        registry.add("spring.datasource.username", postgres::getUsername);
+        registry.add("spring.datasource.password", postgres::getPassword);
+        registry.add("grpc.server.port", ()->6767);
```
thomasbigger584, line 50: `6767` doesnt look to be a dynamic property, this method is used to set properties at runtime. Should this be in the properties file instead?

james8hut: will move

james8hut: it appears that if this test uses the same port number as the grpc server in GrpcIntegrationTest, then you cannot run both tests at the same time. Using a different port number in this test for now to allow both tests to run one after the other.

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
```diff
@@ -0,0 +1,121 @@
...
+
+            ipInterface.setNode(node);
+            ipInterface.setTenantId(UUID.fromString(request.getTenantId()));
+            ipInterface.setIpAddress(new Inet(request.getManagementIp()));
+
+            ipInterfaceRepository.save(ipInterface);
+        }
+    }
+
+    private MonitoringLocation saveMonitoringLocation(DeviceCreateDTO request) {
```
cgorantla, line 60: we should be able to retrieve existing monitoring locations.

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
```diff
@@ -0,0 +1,121 @@
...
+        saveIpInterfaces(request, node, responseObserver);
+
+        responseObserver.onNext(nodeMapper.modelToDTO(node));
+
+        responseObserver.onCompleted();
+    }
+
+    private void saveIpInterfaces(DeviceCreateDTO request, Node node, StreamObserver<NodeDTO> responseObserver) {
+        if (request.hasManagementIp()) {
+            IpInterface ipInterface = new IpInterface();
```
cgorantla, line 50: we should be able to retrieve existing IP Interfaces on the node and make sure this is not a duplicate

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
```diff
@@ -0,0 +1,121 @@
...
+
+        return nodeRepository.save(node);
+    }
+
+    private void validateInput(DeviceCreateDTO request, StreamObserver<NodeDTO> responseObserver) {
+        boolean validTenant = true;
+        boolean validIp = true;
+
+        try {
+            UUID.fromString(request.getTenantId());
```
cgorantla, line 85: Since we are moving tenant id to string, we can remove checks for validTenant

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
james8hut: ok

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
james8hut: Bit confused here. So, nodeId is a field on the IpInterface table. If the ip address exists on an existing IpInterface, does that mean that the node itself already exists, and therefore we should not create any extra rows?

If my thinking here is correct, if the label and/or location is different, should we then update these fields instead?

inventory/src/main/java/org/opennms/horizon/inventory/service/NodeService.java
```diff
@@ -17,10 +19,16 @@
...
     private final NodeRepository modelRepo;
 
+    private final MonitoringLocationRepository monitoringLocationRepository;
+
     private final NodeMapper mapper;
 
     public NodeDTO saveNode(NodeDTO dto) {
         Node model = mapper.dtoToModel(dto);
+
+        MonitoringLocation monitoringLocation = monitoringLocationRepository.getReferenceById(dto.getMonitoringLocationId());
```
yangli2014, line 29: How about when the location doesn't exist?

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
```diff
@@ -0,0 +1,97 @@
...
+import java.time.LocalDateTime;
+import java.util.Optional;
+import java.util.UUID;
+
+@Slf4j
+@RequiredArgsConstructor
+public class DeviceGrpcService extends DeviceServiceGrpc.DeviceServiceImplBase {
+    private final NodeRepository nodeRepository;
+    private final MonitoringLocationRepository monitoringLocationRepository;
+    private final IpInterfaceRepository ipInterfaceRepository;
```
yangli2014, line 32: It is better to use NodeService with different repositories. In this way, we can put all lookups, create new locations, and IP interfaces business logic inside the service. The gRPC services will work similarly to a controller.

inventory/src/test/java/org/opennms/horizon/inventory/NodeIT.java
```diff
@@ -0,0 +1,270 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+    classes = InventoryApplication.class)
+@Testcontainers
```
yangli2014, line 39: You can merge the develop branch and use PostgresInitializer, only one Postgres container will be created when running multiple test classes.

inventory/src/main/java/org/opennms/horizon/inventory/service/NodeService.java
james8hut: This code is only called from the REST layer, which I'm going to remove in the next story.

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
james8hut: will change

inventory/src/test/java/org/opennms/horizon/inventory/NodeIT.java
james8hut: will merge

inventory/src/main/java/org/opennms/horizon/inventory/service/grpc/DeviceGrpcService.java
cgorantla: yes, A node can have multiple interfaces and an Interface can't exist without a node.  

Here we are dealing with a node with one interface.  There shouldn't be duplicate Interface for a given location. We should throw error so that UI doesn't add new Node with the same interface.

we can check this in a separate issue if needed

cgorantla (approved):
LGTM!

PR #482, opened 2022-11-23
[HS-614] Locally run Minion using Tilt

geraldhumphries:
## Description

Experimental. This could be really helpful but will probably suck to maintain. I am wondering if it's worth integrating local minion runs into Tilt at all.

- Adds a devAssembly profile that produces a 'development' version of the assembly. 
- The dev assembly includes a `users.properties` file that has an `admin/admin` user. The file is filtered with variables that are named to make it clear that they're for development only.
- Adds a resource to Tilt that will compile and run a Minion locally via the Karaf assembly. This is called `minion-local` and is disabled by default.
- Adds a button to Tilt you can click to enable `bundle:watch *`. This was done with an execution on karaf-maven-plugin
- Adds a button to Tilt that lets you send arbitrary commands to Karaf
- Devs can log into the karaf shell with SSH like normal
- Renames the Kubernetes deployment of Minion to `minion-k8s`
- Disables feature verification during the dev loop for Minion and Core for speedier builds
- Requires you to set the `JAVA_HOME` environment variable and point it at JDK 11. This is only necessary for `minion-local`.

## Jira link(s)
- https://issues.opennms.org/browse/HS-614

## Flagged for review

- Previous solution with the 'minion-local' resource was likely to be brittle, so I refactored it into a profile-based solution in Maven: https://github.com/OpenNMS/horizon-stream/pull/482/files/e2978ce8e146235bc593f64b024a069da7dd6789
- The assembly built by minion-local uses a subfolder within `target`. Otherwise the minion-k8s build would interfere with it, as it also builds the assembly.
- The `devAssembly` profile is very verbose, and the concept is something I am very not sure about. Would it be better as a totally separate Maven module?
- Using the assembly directly like this might also suck to maintain. 

## Checklist

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,5 +1,5 @@
 # Tilt config #
-load('ext://uibutton', 'cmd_button')
+load('ext://uibutton', 'cmd_button', 'text_input')
 
 secret_settings(disable_scrub=True)  ## TODO: update secret values so we can reenable scrub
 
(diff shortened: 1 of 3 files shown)


geraldhumphries:
Closing as part of PR cleanup.

PR #483, opened 2022-11-24
HS-603: Add Event gRPC Service to query events

thomasbigger584:
## Description
HS-603: Add Event gRPC Service to query events

## Jira link(s)
- https://issues.opennms.org/browse/HS-603

## Checklist


shared-lib/events/src/main/proto/events.proto
```diff
@@ -67,3 +70,23 @@ message EventParameter {
   string type = 3;
   string encoding = 4;
 }
+
+message EventDTO {
```
cgorantla, line 74: I think we should separate grpc events api from kafka events proto.

thomasbigger584: seperate this into another proto file?

cgorantla: yes, we need to move EventDTO into separate proto file.  don't need to depend on the same EventParameter/EventInfo

thomasbigger584: done

events/persistence/src/main/java/org/opennms/horizon/events/persistence/mapper/EventMapper.java
```diff
@@ -0,0 +1,56 @@
+package org.opennms.horizon.events.persistence.mapper;
```
cgorantla, line 1: missing copyright headers

cgorantla (approved):
lgtm!

PR #484, opened 2022-11-24
DC-429: Add Telemetry Listener to new minion: copy code

patrick-schweizer:
This PR copies the code from OpenNMS Classic (develop branch) to Minion NG.
No code was modified.

## Jira link(s)
- https://issues.opennms.org/browse/DC-429


splatch:
Just a note since its a copied code - we might need to align artifacts and features to layout used in project or discuss and agree how all this should be segregated.

patrick-schweizer:
> Just a note since its a copied code - we might need to align artifacts and features to layout used in project or discuss and agree how all this should be segregated.

@splatch,
how do you think it should look like?

PR #489, opened 2022-11-28
HS-629 add location data loader for GraphQL API

yangli2014:
## Description

Test: GraphQL playground's GraphQL schema for node(s) and minion(s) return type also includes location.

## Jira link(s)
- https://issues.opennms.org/browse/HS-629

## Checklist


cgorantla (approved):
Nice work on data loader.  LGTM!

inventory/src/main/java/org/opennms/horizon/inventory/grpc/InventoryServerInterceptor.java
```diff
@@ -59,18 +59,23 @@ public class InventoryServerInterceptor implements ServerInterceptor {
...
-            return Contexts.interceptCall(context, serverCall, headers, callHandler);
-        } catch (VerificationException e) {
-            log.error("Failed to verify access token", e);
-            serverCall.close(Status.UNAUTHENTICATED.withDescription("Invalid access token"), new Metadata());
-            return new ServerCall.Listener<>() {};
-        }
-        catch (NoSuchElementException e) {
-            serverCall.close(Status.UNAUTHENTICATED.withDescription("Missing tenant id"), new Metadata());
-            return new ServerCall.Listener<>() {};
+        if (!Constants.AUTH_HEADER_SKIP_TOKEN.equals(authHeader)) {
```
j-white, line 62: Seems dangerous - skip auth if a special value is set - is this intended to stay this way?

cgorantla: We are using auth token to retrieve tenant id, since we already have location ids that are specific to tenant,  we don't need to pass tenant id.

If Auth is also primary goal with the Interceptor, probably we shouldn't skip this.

yangli2014: This is for the GraphQL data loader (it seems impossible to use the access token as the data loader key) only for now. All the external APIs still need to pass the authorization. Also, Gerald mentioned we will add an internal BFF to pass tenant id to directly to the inventory (without auth). We might change the logic by that time.

yangli2014: The skipping auth happened in the second round of the call (to get the location for the node) and already passed the auth in the first round call to get the node. Let me try if there is a way to avoid this.

cgorantla (approved):
LGTM!

PR #509, opened 2022-12-07
disable ignite notifier

j-white:
Disable Ignite version lookups at runtime

Working towards reducing the number of external calls made by the services


splatch:
I just come across this PR because I look at reduction of env variables for minion image. I'll pull it in for further improvements.

PR #520, opened 2022-12-09
HS-640: Derive Node Status from Critical service (ICMP) on Management IP

thomasbigger584:
## Description
Request:
```
query {
    nodeStatus(id: 1) {
       id,
       status
    }
}
```

Response:
```
{
    "data": {
        "nodeStatus": {
            "id": 1,
            "status": "UP"
        }
    }
}
```

## Jira link(s)
- https://issues.opennms.org/browse/HS-640

## Checklist
* [] Documentation has been updated as necessary.

### DIFF
--- a/rest-server/src/main/java/org/opennms/horizon/server/service/NodeStatusService.java
+++ b/rest-server/src/main/java/org/opennms/horizon/server/service/NodeStatusService.java
@@ -1,0 +1,75 @@
+package org.opennms.horizon.server.service;
+
+import lombok.RequiredArgsConstructor;
+import org.opennms.horizon.inventory.dto.IpInterfaceDTO;
+import org.opennms.horizon.inventory.dto.NodeDTO;
+import org.opennms.horizon.server.model.TSResult;
+import org.opennms.horizon.server.model.TimeRangeUnit;
+import org.opennms.horizon.server.model.TimeSeriesQueryResult;
+import org.opennms.horizon.server.service.grpc.InventoryClient;
+import org.springframework.stereotype.Service;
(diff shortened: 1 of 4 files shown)


rest-server/src/main/java/org/opennms/horizon/server/service/NodeStatusService.java
```diff
@@ -0,0 +1,80 @@
+package org.opennms.horizon.server.service;
```
cgorantla, line 1: missing copyright headers

rest-server/src/main/java/org/opennms/horizon/server/service/NodeStatusService.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.concurrent.TimeoutException;
+
+import static java.util.Objects.isNull;
+import static org.springframework.util.CollectionUtils.isEmpty;
+
+@Service
+@RequiredArgsConstructor
+public class NodeStatusService {
+    private static final String RESPONSE_TIME_METRIC = "response_time_msec";
+    private static final int TIME_RANGE_IN_MINUTES = 1;
```
cgorantla, line 27: Make this 90 secs ?  That will prevent edge case where gap between metrics is more than 1 minute.  I think we need to consolidate these constants at some point

rest-server/src/main/java/org/opennms/horizon/server/service/NodeStatusService.java
thomasbigger584: should we make sure copyright headers are added to all classes all at once ?

PR #521, opened 2022-12-09
Add initial cucumber tests to alarm service with docker containers

mfrazier63:
## Description
This PR has the initial work to add in cucumber integration tests for the alarm service.

Utilizes docker images in the docker-maven-plugin to spin up the microservice with all it's dependencies (kafka, postgress, etc) so that we can interact with the alarm service through kafka and REST and examine the results.

For now has just one simple test case that sends and event to kafka, then makes sure that an alarm was created and persisted.

## Jira link(s)

## Checklist


Tiltfile
```diff
@@ -126,7 +126,7 @@ jib_project(
     'opennms/horizon-stream-alarm',
     'alarm',
     'opennms-alarm',
-    port_forwards=['29080:9090', '29050:5005', '29065:6565',  '29000:8080'],
+    port_forwards=['31080:9090', '31050:5005', '31065:6565',  '31000:8080'],
```
thomasbigger584, line 129: think we've to change this to 32xxx actually because it looks like the new proxy service took 31xxx

mfrazier63: yeah i just picked that up when back merging develop

geraldhumphries (approved):

alarm/src/main/java/org/opennms/horizon/alarmservice/rest/AlarmRestServiceImpl.java
```diff
@@ -98,32 +96,60 @@ public Response getAlarms(@Context final SecurityContext securityContext, @Conte
...
-            return Response.status(Status.OK).entity(alarmsCollection).build();
+            return ResponseEntity.ok(alarmsCollection);
     }
     
-    @PostMapping(path="{id}/clear", produces = MimeTypeUtils.APPLICATION_JSON_VALUE)
-    public String clearAlarm(@PathVariable Long id) {
+    @PostMapping(path="/clear/{id}", produces = MimeTypeUtils.APPLICATION_JSON_VALUE)
+    public ResponseEntity<AlarmDTO> clearAlarm(@PathVariable Long id) {
+
+        return ResponseEntity.ok(alarmService.clearAlarm(id, new Date()));
```
artnaseef, line 105: Nit here.  I personally prefer to perform the operation on one line, then wrap the result on another, even though Intellij complains the local variable can be removed.  I find it easier to read and debug.

Just felt like sharing.  No need to make any change.

alarm/src/main/resources/org/opennms/horizon/dockerit/alarm.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Alarm Service Basic Functionality
```
artnaseef, line 1: Was this meant to be left commented-out?

Ahh, see the other one in src/tes/resources.  This should just be removed, right?

PR #528, opened 2022-12-09
HS-660 Align kafka processors

splatch:
## Description
A cleanup of earlier changes introduced with HS-660, removing duplicated code and cleaning up third kafka producer we host in minion-gateway.

## Jira link(s)
- https://issues.opennms.org/browse/HS-660

## Checklist


cgorantla (approved):
lgtm!

minion-gateway/main/src/main/java/org/opennms/miniongateway/grpc/server/kafka/KafkaProducer.java
```diff
@@ -0,0 +1,48 @@
+package org.opennms.miniongateway.grpc.server.kafka;
```
artnaseef, line 1: I got really tripped up by the name of this class being the same as the Kafka class name.  How about calling it `TenantAwareKafkaProducer`?

splatch: Renamed it to `KafkaGrpcMessagePublisher`, so without mentioning tenancy context. I stepped back from producer semantics which you found confusing with kafka-producer scripts and tools, let me know if its fine and if we can merge these changes.

splatch:
I had to update this PR as I found one more thing which was happening - we used in all places same configuration key but different value. I also moved log statement down to publisher itself to keep forwarders as simple as it gets. Please re-review changes.

splatch:
I'll be updating this PR, but first have to verify if its still needed!

minion-gateway/main/src/main/java/org/opennms/miniongateway/grpc/server/kafka/KafkaGrpcMessagePublisher.java
```diff
@@ -0,0 +1,83 @@
...
+        List<Header> headers = new LinkedList<>();
+        headers.add(new RecordHeader(TENANT_ID_HEADER_NAME, tenantId.getBytes(StandardCharsets.UTF_8)));
+
+        kafkaTemplate.send(new ProducerRecord<String, byte[]>(
+            topic,
+            null,
+            null,
+            content.toByteArray(),
+            headers
+        ));
```
cgorantla, line 81: No need to fill nulls, this may be better

```suggestion
        var producerRecord = new ProducerRecord<String, byte[]>(kafkaTopic, content.toByteArray());
        // add tenant id to kafka header
        producerRecord.headers().add(new RecordHeader(GrpcConstants.TENANT_ID_KEY,
            tenantId.getBytes(StandardCharsets.UTF_8)));
         kafkaTemplate.send(producerRecord);
```

artnaseef (approved):
Nice reuse

cgorantla (approved):
LGTM!

PR #531, opened 2022-12-12
HS-581 : Generate new event from suspect traps and sent to internal event topic

yangli2014:
## Description

What changed: Create a new event with a suspect Trap message and send it to a new internal event topic. Consume this new event in the inventory and create a new node if the IP interface with the same location and tenant id doesn't exist in the database. 

How to test:
- Wire up the environment with tilt or other k8s tools
- Create a snmp pod with command: _kubectl run snmpd-node --image polinux/snmpd --restart=Never_ 
- Get the minion pod IP address
- Run this command to get into the new created snmp pod _kubectl exec -it pod/snmpd-node -- /bin/sh_
- Run this command inside the snmp pod with minion IP address: _/usr/bin/snmptrap -v 2c -c public <minion IP address>:1162 '' 1.3.6.1.4.1.8072.2.3.0.1 1.3.6.1.4.1.8072.2.3.2.1 i 123456_
- Wait for a few minutes and check the new node created in the Inventory tab on the UI page.

## Jira link(s)
- https://issues.opennms.org/browse/HS-581
- https://issues.opennms.org/browse/HS-653
- https://issues.opennms.org/browse/HS-654

## Checklist


events/traps/src/main/java/org/opennms/horizon/events/traps/EventFactory.java
```diff
@@ -131,6 +131,7 @@ private boolean shouldDiscard(Event econf) {
         return false;
     }
 
+    //TODO: we don't need this here. Get node id in the inventory is much easier.
```
thomasbigger584, line 134: this method is already implemented in another PR

events/traps/src/main/java/org/opennms/horizon/events/traps/TrapsConsumer.java
```diff
@@ -112,11 +114,27 @@ public void consume(@Payload byte[] data, @Headers Map<String, Object> headers)
             // Send them to kafka
             eventForwarder.sendEvents(eventLogProto, tenantId);
 
+            eventLogProto.getEventList().forEach(e ->{
+                    sendNewSuspectEvent(e, tenantId);
```
cgorantla, line 118: we should only send new suspect event if we don't find node id in the event.

yangli2014: I saw the code, we don't have the code to look up node id yet. And looking up the node id is much easier in inventory service. Maybe there will be too many internal event messages without this filer?

cgorantla: yeah, PR #604 will fill up the node details on the event when it can find one.

For those events where we don't find node, we will send new suspect events.

yangli2014: The node id in the event is only for this filter here? If that is the case it might be better to keep the code to look up the node id inside the inventory service.

events/traps/src/main/java/org/opennms/horizon/events/traps/EventFactory.java
yangli2014: I see. Do we really need it here? Lookup node id in the inventory services to filter out the event is much simpler and easier to maintain.

events/traps/src/main/java/org/opennms/horizon/events/traps/TrapsConsumer.java
cgorantla: Having a missing nodeId is a rare case not a regular occurrence.  New suspect event is very rare internal event.

A device continuously sends traps, they are converted to events in the event service itself. Events doesn't to go another service to attach nodeId.

Also, current approach to get nodeId from inventory service will be optimized to internal cache in Events service, then we don't need to query inventory service most of them

yangli2014: sure. I will updated it.

inventory/src/main/java/org/opennms/horizon/inventory/component/InternalEventsConsumer.java
```diff
@@ -0,0 +1,76 @@
...
+@Component
+@PropertySource("classpath:application.yml")
+public class InternalEventsConsumer {
+    private final NodeService nodeService;
+    private final DetectorTaskSetService detectorService;
+
+    @KafkaListener(topics = "${kafka.topics.internal-events}", concurrency = "1")
+    public void receiveTrapEvent(@Payload byte[] data, @Headers Map<String, Object> headers) {
+        try {
+            var event = Event.parseFrom(data);
```
cgorantla, line 62: should filter this based on `EventConstants.NEW_SUSPECT_INTERFACE_EVENT_UEI`

cgorantla (approved):

PR #533, opened 2022-12-12
HS-621 remove core REST API endpoints in BFF 

yangli2014:
## Description

We can shut down the core now. The only API still pointing to the core is the data choice, however, the UI doesn't use it. 
We need to implement gRPC for alarms and the new data choice service.
Another thing is the notification is still using REST API, we need to switch to the gRPC as well.

## Jira link(s)
- https://issues.opennms.org/browse/HS-621

## Checklist


charts/opennms/templates/opennms/minion/minion-configmap.yaml
```diff
@@ -13,3 +13,17 @@ data:
   opennms-minion-identity: |
     id=${env:HOSTNAME:-UNKNOWN}
     location=${env:LOCATION:-Default}
+---
```
jkapponms, line 16: this needs to be in it's own file

artnaseef, line 16: The contents of the `opennms-config` can be merged with `minion-config` above to have a single config.

inventory/pom.xml
```diff
@@ -62,11 +62,6 @@
             <version>${lombok.version}</version>
             <optional>true</optional>
         </dependency>
```
artnaseef, line 64: Did inventory change as part of removing core?

charts/opennms/templates/opennms/minion/minion-configmap.yaml
yangli2014: ok, still in the minion folder?

rest-server/src/main/java/org/opennms/horizon/server/service/AlarmService.java
```diff
@@ -1,63 +0,0 @@
-/*******************************************************************************
```
artnaseef, line 1: Did Alarm service change as a part of removing core?

rest-server/src/main/java/org/opennms/horizon/server/service/DataChoicesService.java
```diff
@@ -1,47 +0,0 @@
-package org.opennms.horizon.server.service;
```
artnaseef, line 1: Is this change related to core?

rest-server/src/main/java/org/opennms/horizon/server/service/DeviceService.java
```diff
@@ -1,66 +0,0 @@
-/*******************************************************************************
```
artnaseef, line 1: Is this related to the removal of core?

rest-server/src/main/java/org/opennms/horizon/server/service/GrpcAlarmService.java
```diff
@@ -28,34 +28,6 @@
 
 package org.opennms.horizon.server.service;
```
artnaseef, line 30: Is this related to the removal of core?

yangli2014: By removing the REST API-based service, it is a place holder for the gRPC service.

rest-server/src/main/java/org/opennms/horizon/server/service/AlarmService.java
yangli2014: yes

rest-server/src/main/java/org/opennms/horizon/server/service/DataChoicesService.java
yangli2014: yes

jkapponms (approved):

geraldhumphries (approved):

PR #556, opened 2022-12-14
Asn/test containers maven lifecycle

artnaseef:
DRAFT - for discussion


j-white:
+1 from me

minion-gateway/docker-it/src/test/resources/org/opennms/horizon/minion-gateway.feature
```diff
@@ -0,0 +1,56 @@
+Feature: Minion Gateway RPC Request Processing
+
+  Background: Common Test Setup
+    Given Internal GRPC Port in system property "application-internal-grpc-port"
```
splatch, line 4: I think for consistency reasons both grpc port and host should be configurable. This way scenario can be adopted to external systems to verify behavior of deployed installations.

splatch (approved):
:+1: LGTM with just one small nitpick

minion-gateway/docker-it/src/test/resources/logback.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
```
splatch, line 2: Looks like we have both log4j and slf4j at the same time - is it intentional?

PR #560, opened 2022-12-16
HS-665 Resend detector tasks for nodes on the inventory restart

yangli2014:
## Description

Make sure the detector tasks for all existing nodes are sent and the minion received all the tasks. 
However, there aren't any metrics for devices coming back after the inventory restarted. It worked once in a while a few days ago. Please refer the issue: https://opennms.atlassian.net/browse/HS-712

## Jira link(s)

## Checklist


inventory/src/main/java/org/opennms/horizon/inventory/service/taskset/DetectorTaskSetService.java
```diff
@@ -69,25 +74,27 @@ public void sendDetectorTasks(Node node) {
         for (MonitorType monitorType : DETECTOR_MONITOR_TYPES) {
             addDetectorTasks(node, ipInterfaces, monitorType);
         }
-
         sendTaskSet(node);
     }
 
+    public void sendDetectorTaskForNodes(Map<String, Map<String, List<NodeDTO>>> nodeListMap) {
```
cgorantla, line 80: I think this still  publishes tasksets for each node sequentially same as before.
Is there advantage doing this way than earlier ?

yangli2014: No, if you add more than one node you try it you will see the difference.

yangli2014: The difference is at line 86 when using tenantId and location to retrieve the Taskset it will send all the tasks for the same tenant and location once. The quick fix will keep the tasks in the hashmap and repeat sending them again and again when looping each node.

yangli2014: The first loop is prepareing task set based on tenant id and location combination. The second loop send the task set

cgorantla: got it, thank you!

cgorantla (approved):
lgtm!

PR #571, opened 2022-12-23
HS-745 Avoid the ventory resend tasks for the same node again

yangli2014:
## Description

Clean the task set cache every time when inventory sends a task set for the tenant id and location combination. 
## Jira link(s)
- https://issues.opennms.org/browse/HS-745

## Checklist


cgorantla:
Current design of the Minion gateway requires resending  whole taskset every time we publish new TaskSet.

yangli2014:
OK, good to know, so when I add multiple nodes, the last node will trigger sending all the task set for all nodes again? Not sure if this is a good design or any technical reason we have to do this.

cgorantla:
> OK, good to know, so when I add multiple nodes, the last node will trigger sending all the task set for all nodes again? Not sure if this is a good design or any technical reason we have to do this.

I think this is limitation of Minion Gateway. Should be resolved soon.

PR #567, opened 2022-12-21
HS-746: Add Cucumber IT for Inventory

cgorantla:
## Description

Spins up Inventory Application and validates following functionality.
 - Sending Minion heartbeat creates location/monitoring system
 - Create new device and validates that device got created
 

This PR depends on #574 for tests to pass
  

## Jira link(s)
- https://issues.opennms.org/browse/HS-746

## Checklist
* [ ] Follows Horizon Stream's [development guidelines.](https://github.com/OpenNMS/horizon-stream/wiki/Development-Guidelines)
* [ ] Appropriate reviewer(s) have been selected.
* [ ] Jira issue(s) have been updated to "In Review".
* [ ] Includes [appropriate tests.](https://github.com/OpenNMS/horizon-stream/wiki/Test-Strategy)
* [ ] Documentation has been updated as necessary.

[HS-746]: https://opennms.atlassian.net/browse/HS-746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #572, opened 2022-12-23
HS-748: Cleanup publishing tasksets

cgorantla:
## Description

Use single API call to publish TaskSets.
Removes TaskSetManager/TaskSetManagerUtil.

Cleanup Tasks generation and publishing

This depends on PR #574 
## Jira link(s)
- https://issues.opennms.org/browse/HS-748

## Checklist

[HS-748]: https://opennms.atlassian.net/browse/HS-748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thomasbigger584 (approved):

shared-lib/task-set-service/ignite-client/src/main/java/org/opennms/taskset/service/igniteclient/impl/TaskSetIgnitePublisherImpl.java
```diff
@@ -29,10 +29,12 @@
...
 import org.apache.ignite.client.IgniteClient;
+import org.opennms.taskset.contract.TaskDefinition;
 import org.opennms.taskset.contract.TaskSet;
 import org.opennms.taskset.service.api.TaskSetPublisher;
 import org.opennms.taskset.service.model.LocatedTaskSet;
 
+import java.util.List;
 import java.util.function.Consumer;
 
 @Deprecated
```
yangli2014, line 40: Are there any places this deprecated class is still in use?

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/TaskSetPublisherImpl.java
```diff
@@ -89,6 +91,11 @@ public void publishTaskSet(String tenantId, String location, TaskSet taskSet) {
         }
     }
 
+    @Override
+    public void publishNewTasks(String tenantId, String location, List<TaskDefinition> taskList) {
```
yangli2014, line 95: Do we plan to add more code here?

shared-lib/task-set-service/api/src/main/java/org/opennms/taskset/service/api/TaskSetPublisher.java
```diff
@@ -28,10 +28,16 @@
...
 
+import java.util.List;
+
 public interface TaskSetPublisher {
+
     String TASK_SET_PUBLISH_SERVICE = "task-set.pub-task";
 
     void publishTaskSet(String tenantId, String location, TaskSet taskSet);
+
+    void publishNewTasks(String tenantId, String location, List<TaskDefinition> taskList);
```
yangli2014, line 42: If this method only is implemented in one class, it is better not put it into the interface here.

yangli2014 (commented):
Great job. After restarting the inventory service there aren't any metrics received from all three nodes I added. This might be related to this PR.

cgorantla:
> Great job. After restarting the inventory service there aren't any metrics received from all three nodes I added. This might be related to this PR.

that doesn't sound like great job.  I'm in process of separating of PR #567 as you requested.  Will evaluate if this is causing any regressions.

cgorantla:
> Great job. After restarting the inventory service there aren't any metrics received from all three nodes I added. This might be related to this PR.

I could reproduce this on develop as well. The problem seems to be that sending detector tasks again to same Minion won't trigger detection since that task is already done by Minion.  Since detection is not done, Monitor/Collector are not getting triggered.  

Ideally we shouldn't need to re-send tasks from Inventory to Minion gateway once it has it's own persistence and can handle partial updates.  Till then there is no easy way to do this reliably.  ( We may send detector tasks as RPC but since detector tasks are going to be replaced by Node Scan, we don't need to change it now)

yangli2014 (approved):
Yes, I can reproduce it on develop branch. But it seems not consistent. Sometimes only part of the nodes' metrics missing.

PR #583, opened 2023-01-06
HS-787: Add deleting functionality in the inventory

yangli2014:
## Jira link(s)
- https://issues.opennms.org/browse/HS-787

## Checklist


inventory/src/main/java/org/opennms/horizon/inventory/grpc/MonitoringSystemGrpcService.java
```diff
@@ -66,15 +67,33 @@ public void getMonitoringSystemById(StringValue systemId, StreamObserver<Monitor
...
+    }
+
+    @Override
+    public void deleteMonitoringSystem(StringValue request, StreamObserver<BoolValue> responseObserver) {
+        Optional<MonitoringSystemDTO> monitoringSystem = tenantLookup
+            .lookupTenantId(Context.current())
+            .map(tenantId -> service.findBySystemId(request.getValue(), tenantId))
+            .orElseThrow();
+        monitoringSystem.ifPresentOrElse(system -> {
+            service.deleteMonitoringSystem(system.getId());
```
cgorantla, line 86: what happens when `deleteMonitoringSystem` throws exception.
It is better to return proper error response instead of throwing exception in this case.

cgorantla (commented):
Need to revisit whether we always need to send proper error response instead of throwing exception to the grpc call.

Otherwise looks good to me.

PR #594, opened 2023-01-09
HS-389: Azure Monitoring (POC)

thomasbigger584:
## Description

IN PROGRESS - Need tidyups and tests, submitting for early review

Adding new Scanner Task to scan for nodes (azure virtual machines)
When we add nodes to database, we send Monitor and Collector tasks to minion to perform the work.

Currently generating dummy IP interface information, we may want to query for this from Azure API.

Using existing ICMP and SNMP MonitorTypes as the UI does queries based on these MonitorTypes. We should eventually create an AZURE MonitorType at some point.

image

## Jira link(s)
- [HS-389](https://opennms.atlassian.net/browse/HS-389)

## Checklist


minion/plugins/azure-plugin/src/main/java/org/opennms/horizon/minion/azure/AzureScanner.java
```diff
@@ -0,0 +1,113 @@
...
+
+            for (AzureValue resourceGroupValue : resourceGroups.getValue()) {
+                String resourceGroup = resourceGroupValue.getName();
+
+                AzureResources resources = client.getResources(token, request.getSubscriptionId(),
+                    resourceGroup, request.getTimeout(), request.getRetries());
+
+                scannedItems.addAll(resources.getValue().stream()
+                    .filter(azureValue -> azureValue.getType()
+                        .equalsIgnoreCase(MICROSOFT_COMPUTE_VIRTUAL_MACHINES))
```
yangli2014, line 87: Only managed Microsoft VM?

yangli2014 (commented):
Great job! We might need to think about reusable code with the standard node scanner.

minion/plugins/azure-plugin/src/main/java/org/opennms/horizon/minion/azure/AzureScanner.java
thomasbigger584: yes, this list contains all resources. we are mapping here VMs as nodes

thomasbigger584:
> Great job! We might need to think about reusable code with the standard node scanner.

Yeah since I cannot yet create a new `MonitorType`, due to the UI querying specifically for ICMP and SNMP i have opted to create a new method there, similar to monitor and collect. Once we can make a new `MonitorType` where the UI also queries for MonitorType.AZURE, then it will be easier i think. This is only done for demo purpose and will likely change

shared-lib/protobuf/src/main/proto/taskSet.proto
```diff
@@ -74,16 +76,22 @@ message TaskResult {
...
+    CollectorResponse collector_response = 7;
   }
 }
 
 message TaskSetResults {
   repeated TaskResult results = 1;
 }
 
+message ScannerResponse {
+  repeated opennms.azure.AzureScanItem results = 1; //todo tbigg: change to any
```
thomasbigger584, line 91: this needs to be more generic with `any`, similar to `CollectorResponse`

shared-lib/protobuf/src/main/proto/taskSet.proto
```diff
@@ -31,18 +31,25 @@ import "google/protobuf/any.proto";
...
 option java_package = "org.opennms.taskset.contract";
 
 enum TaskType {
       UNDEFINED = 0; // a fallback?
-      DETECTOR = 1;
-      MONITOR = 2;
-      LISTENER = 3;
-      CONNECTOR = 4;
-      COLLECTOR = 5;
+      SCANNER = 1;
```
j-white, line 42: What is a "scanner"? Curious on how we define this.

shared-lib/protobuf/src/main/proto/azure.proto
```diff
@@ -0,0 +1,70 @@
...
+
+// todo: create common credential fields into reusable message
+
+message AzureScanRequest {
+  int64 credential_id = 1;
+  string client_id = 2;
+  string client_secret = 3;
+  string subscription_id = 4;
+  string directory_id = 5;
+  int64 timeout = 6;
```
j-white, line 44: For time related field, it's good practice to indicate the units. May `timeout_ms`?

shared-lib/azure/proto/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+            <artifactId>protobuf-java-util</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.10.1</version>
```
j-white, line 38: Can we centrally manage plugin versions?

shared-lib/protobuf/src/main/proto/taskSet.proto
thomasbigger584: This is a task which is used to discover nodes. The `AzureScanner` will fetch for a list of nodes (Azure VMs using the management API) to be sent back and stored in the node table.  Should it be called something else?

Once we store these as nodes, we can send Monitor/Collector tasks to check node status up/down, and collect metrics, similar to how we have it elsewhere, we just dont have a detector to detect services.

thomasbigger584: Currently I am not fetching for IP Addresses of the VM, although this is possible. We are able to get metrics directly from the Resource Name/Resource Group. This can be added if required.

j-white: Got it. Seems like a similar concept to a RequisitionProvider in Horizon: interrogate some external API and then map this to "nodes" for monitoring

Don't need to change anything now, just inquiring.

inventory/src/main/java/org/opennms/horizon/inventory/service/AzureCredentialService.java
```diff
@@ -27,29 +27,106 @@
...
+            .findByLocationAndTenantId(location, tenantId);
+
+        if (locationOp.isPresent()) {
+            return locationOp.get();
+        }
+
+        MonitoringLocation monitoringLocation = new MonitoringLocation();
+        monitoringLocation.setLocation(location);
+        monitoringLocation.setTenantId(tenantId);
+        monitoringLocation = locationRepository.save(monitoringLocation);
```
cgorantla, line 125: what happens this throws exception ?

inventory/src/main/java/org/opennms/horizon/inventory/service/AzureCredentialService.java
```diff
@@ -27,29 +27,106 @@
...
+
         AzureCredential credential = mapper.dtoToModel(request);
         credential.setTenantId(tenantId);
         credential.setCreateTime(LocalDateTime.now());
-        credential = repository.save(credential);
+        credential.setMonitoringLocation(monitoringLocation);
+        final AzureCredential savedCredential = repository.save(credential);
+
+        // Asynchronously send task sets to Minion
+        executorService.execute(() -> scannerTaskSetService.sendAzureScannerTask(savedCredential));
```
cgorantla, line 84: should this be handled in different method than `createCredentials` ?

inventory/src/main/java/org/opennms/horizon/inventory/service/taskset/CollectorTaskSetService.java
```diff
@@ -62,6 +61,14 @@ public void sendCollectorTask(String location, MonitorType monitorType, IpInterf
         }
     }
 
+    public void sendAzureCollectorTasks(AzureCredential credential, AzureScanItem item, String ipAddress, long nodeId) {
```
cgorantla, line 64: I would favor separating out Azure related tasks into different service, good for now. We may want to revisit this later.

cgorantla (commented):
Overall, it looks very good.

Few comments.

inventory/src/main/java/org/opennms/horizon/inventory/service/AzureCredentialService.java
thomasbigger584: sure will add a method for async sending in ScannerTaskSetService

inventory/src/main/java/org/opennms/horizon/inventory/service/AzureCredentialService.java
thomasbigger584: Im only catching InventoryRuntimeException, i can catch a more generic Exception also to handle this

inventory/src/main/java/org/opennms/horizon/inventory/service/taskset/TaskUtils.java
```diff
@@ -5,22 +5,34 @@ public interface TaskUtils {
...
 
     static String identityForIpTask(String ipAddress, String name) {
         return IP_LABEL + ipAddress + "/" + name;
     }
 
+    static String identityForAzureTask(String name) {
+        return AZURE_LABEL + name;
+    }
+
+    interface Azure {
```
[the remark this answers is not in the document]
thomasbigger584: yeah sonarcloud will complain about this, kept it in to keep consistency but can change it now if required

minion/plugins/azure-plugin/src/main/java/org/opennms/horizon/minion/azure/AzureCollector.java
```diff
@@ -0,0 +1,181 @@
...
+            AzureInstanceView instanceView = client.getInstanceView(token, request.getSubscriptionId(),
+                request.getResourceGroup(), request.getResource(), request.getTimeoutMs(), request.getRetries());
+
+            if (instanceView.isUp()) {
+
+                Map<String, Double> collectedData = new HashMap<>();
+
+                collect(request, token, collectedData);
+
+                SnmpResponseMetric results = SnmpResponseMetric.newBuilder()
```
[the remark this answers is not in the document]
thomasbigger584: Yeah for now we are using MonitorType.SNMP so that the values will be displayed in the graphs correctly. We may want to create a new MonitorType.AZURE where we can return a AzureResponseMetric instead but it wouldnt show in graph. 
https://github.com/OpenNMS-Cloud/horizon-stream/blob/develop/metrics-processor/src/main/java/org/opennms/horizon/tsdata/TSDataProcessor.java#L170-L172

https://github.com/OpenNMS-Cloud/horizon-stream/blob/develop/ui/src/containers/Graphs.vue#L69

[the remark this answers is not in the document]
thomasbigger584: created new story, will follow up with this change in another PR

PR #598, opened 2023-01-09
Make sure there is amd64 and armv7/arm64 images for minion

splatch:
## Description
Make sure there is amd64 and armv7/arm64 images so minion deployment on arm is doable.

## Jira link(s)
- https://issues.opennms.org/browse/HS-733

## Flagged for review
Reviewers  selected by awareness of the trouble or ownership of M1 macs.

## Checklist


yangli2014 (approved):
LGTM!

.github/actions/minion/action.yml
```diff
@@ -45,5 +45,10 @@ runs:
 
         mvn verify -DskipTests -Ddocker.image=opennms/horizon-stream-minion:local -Ddocker.skipPush=true
 
+        
+        # Test release
+        docker buildx create --use --driver docker-container --name maven
```
splatch, line 50: If this change stays then I think we can safely remove pom.xml snippet I added as it does not make any profit over explicit pipeline call. We will also consume additional time on build if we go with both buildx calls. Given stability of shell call I'd opt for it over docker plugin used in `pom.xml`.

splatch:
I've removed pom changes which were redundant to workflow. Whole work is done by CI/CD.

jaymberry (approved):

PR #609, opened 2023-01-11
DC-496 Clean up package structure

FrancescoBuc86:
https://opennms.atlassian.net/browse/DC-496


minion/flows/parser/pom.xml
```diff
@@ -44,17 +44,17 @@
       <dependency>
           <groupId>io.github.resilience4j</groupId>
           <artifactId>resilience4j-bulkhead</artifactId>
-          <version>${resilience4j.version}</version>
+          <version>0.17.0</version>
```
patrick-schweizer, line 47: I think we should keep the variable

minion/flows/parser/src/main/java/org/opennms/horizon/minion/flows/shell/SendFlowCmd.java
```diff
@@ -31,11 +31,11 @@ public class SendFlowCmd implements Action {
 
     @Override
     public Object execute() throws Exception {
-        if (file == null || file.isEmpty() || this.getClass().getResource("/flows/" + file) == null) {
+        if (file == null || file.isEmpty() || this.getClass().getResource("/org.opennms.horizon.minion.flows/" + file) == null) {
```
patrick-schweizer, line 34: flows test data stays under /flows, no need to change that

patrick-schweizer (commented):
I wonder if we shouldn't leave the flow test data under /flows as before?

splatch:
Given this cleanup task - there is a naming convention proposed earlier for horizon/minion before separation of all microservices happened: https://github.com/OpenNMS/horizon-stream/wiki/PROPOSED-GROUP-ID---ARTIFACT-ID-NAMING.
Please have a look on it. In short flows module should be named `minion-flows` (to avoid name conflict with possible flows microservice to be created in backend). Sub modules under `minion-flows` should stick with `flows-parser` and so on, without going into additional nesting levels.

minion/flows/parser/pom.xml
FrancescoBuc86: true, forgot to change it back

minion/flows/parser/src/main/java/org/opennms/horizon/minion/flows/parser/transport/Netflow5MessageBuilder.java
```diff
@@ -67,28 +61,28 @@ public FlowMessage.Builder buildMessage(final Iterable<Value<?>> values, final R
         for (Value<?> value : values) {
             switch (value.getName()) {
                 case "@count":
-                    getUInt32Value(value).ifPresent(builder::setNumFlowRecords);
+                    MessageUtils.getUInt32Value(value).ifPresent(builder::setNumFlowRecords);
```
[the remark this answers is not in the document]
FrancescoBuc86: right

FrancescoBuc86:
> I wonder if we shouldn't leave the flow test data under /flows as before?

I changed to be consistent with the flow test classes path, but maybe it's better to keep it under /flows

FrancescoBuc86:
> Given this cleanup task - there is a naming convention proposed earlier for horizon/minion before separation of all microservices happened: https://github.com/OpenNMS/horizon-stream/wiki/PROPOSED-GROUP-ID---ARTIFACT-ID-NAMING. Please have a look on it. In short flows module should be named `minion-flows` (to avoid name conflict with possible flows microservice to be created in backend). Sub modules under `minion-flows` should stick with `flows-parser` and so on, without going into additional nesting levels.

Thanks for the comment, you're right, we'll update the packages / modules name accordingly

patrick-schweizer (approved):
looks good to me

PR #615, opened 2023-01-12
HS-811 - add form for azure

anyarybalova:
## Jira link(s)

## Checklist

### DIFF
--- a/ui/src/components/Discovery/AzureForm.vue
+++ b/ui/src/components/Discovery/AzureForm.vue
@@ -1,0 +1,64 @@
+<template>
+  <div class="azure-container">
+    <div>
+      <div class="location-title">Select a location</div>
+      <DiscoveryLocations />
+    </div>
+    <div class="azure-form">
+      <FeatherInput
+        v-model="store.azure.clientId"
+        label="ClientID"
+      >
+      </FeatherInput>
+      <FeatherInput
+        v-model="store.azure.clientSecret"
+        label="Client Secret"
+      >
+      </FeatherInput>
+      <FeatherInput
+        v-model="store.azure.subscriptionId"
+        label="SubscriptionID"
+      >
+      </FeatherInput>
+      <FeatherInput
+        v-model="store.azure.directoryId"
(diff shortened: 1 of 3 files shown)


ui/src/components/Discovery/AzureForm.vue
```diff
@@ -0,0 +1,64 @@
...
+      >
+      </FeatherInput>
+      <FeatherInput
+        v-model="store.azure.subscriptionId"
+        label="SubscriptionID"
+      >
+      </FeatherInput>
+      <FeatherInput
+        v-model="store.azure.directoryId"
+        label="DirectoryId"
```
mikewrosey, line 25: Maybe `DirectoryID` to keep the same casing as the other ID labels.

ui/src/containers/Discovery.vue
```diff
@@ -40,27 +40,41 @@
...
+        secondary
+        @click="closeModal"
+      >
+        CLOSE
+      </FeatherButton>
+      <FeatherButton
+        v-if="selectedTool === DiscoveryType.Azure"
+        primary
+        @click="saveAzureDiscovery"
+      >
```
mikewrosey, line 65: I see what you mean better about this btn. It would make sense to have it inside the Azure form component, but then it does look out of sync with the close btn. I guess we can leave it this way for now until we find some alternative.

Perhaps we need to remove the footer completely, have the btns inside the content, and just add an `x` to the top right to close the modal

ui/src/containers/Discovery.vue
```diff
@@ -81,7 +98,7 @@ const modalContent = computed(() => {
     case DiscoveryType.ICMP:
       return { component: DiscoveryStepper, props: { callback: closeModal }}
     case DiscoveryType.Azure:
-      // return Azure content
+      return { component: AzureForm, props: { callback: closeModal }}
```
mikewrosey, line 101: If we're leaving the btn here for now, I guess we should remove closeModal from the props and add it to the saveAzureDiscovery fn below

PR #622, opened 2023-01-14
[HS-817] Refactor develop pipeline to use reusable workflows that run in parallel

geraldhumphries:
## Jira link(s)
- https://issues.opennms.org/browse/HS-817

## Checklist


Tiltfile
```diff
@@ -61,7 +61,7 @@ def jib_project(resource_name, image_name, base_path, k8s_resource_name, resourc
 
     custom_build(
         image_name,
-        'mvn jib:dockerBuild -Dimage=$EXPECTED_REF -f {} -Djib.from.platforms=linux/{} {}'.format(base_path, cluster_arch_cmd, submodule_flag),
+        'mvn jib:dockerBuild -Ddocker.image=$EXPECTED_REF -f {} -Djib.from.platforms=linux/{} {}'.format(base_path, cluster_arch_cmd, submodule_flag),
```
artnaseef, line 64: Same concern with `-Ddocker.image` here for projects with more than one image.

install-local/install-local.sh
```diff
@@ -18,9 +18,19 @@ fi
 
 CONTEXT=$1
 DOMAIN=$2
+IMAGE_TAG="local"
+IMAGE_PREFIX="opennms"
 KIND_CLUSTER_NAME=kind-test
 NAMESPACE=hs-instance
 
+if [[ -n "$3" ]]; then
+  IMAGE_TAG=$3
```
artnaseef, line 27: Just a scripting tip - you can use the following syntax as a 1-liner to set the ENV variable and fallback to a default if it is empty:

    IMAGE_TAG=${3:-local}

parent-pom/pom.xml
```diff
@@ -17,9 +17,17 @@
...
         <java.version>17</java.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <sonar.organization>opennms</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
 
+        <!-- Jib properties -->
+        <docker.image>${docker.image.name}:${docker.image.tag}</docker.image>
+        <!--suppress CheckTagEmptyBody -->
+        <docker.image.name></docker.image.name>
```
artnaseef, line 28: HJow about adding a comment to clarify this value MUST be set by the project when using the JIB-maven-plugin?

parent-pom/pom.xml
```diff
@@ -866,6 +875,7 @@
                     </configuration>
                     <executions>
                         <execution>
+                            <id>repackage</id>
```
artnaseef, line 878: This the same `id` as the one provided by the spring-boot-maven parent pom.  I recommend using a different ID - otherwise, if someone manages to get both in the same project, maven will silently merge the two definitions and that can lead to confusion.  Yes, this is speaking-from-experience ;-).

artnaseef (changes requested):
Many comments are intended for follow-up or just FYI.  There are 2 or 3 minor things that would be good to address now.

parent-pom/pom.xml
geraldhumphries: I did that on purpose to match what the spring boot starter parent pom does, but I didn't know about that downside. Will revert!

jaymberry (approved):

.github/actions/jib-maven-sonar/action.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Build Image & Scan
+      env:
+        IMAGE: ${{ inputs.image }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
+        SONAR_TOKEN: ${{ inputs.sonar-token }}
+        SONAR_PROJECT_KEY: ${{ inputs.project-key }}
+      run: |
+        mvn -B verify \
+            jib:build -Ddocker.image=$IMAGE \
```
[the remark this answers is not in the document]
geraldhumphries: The property used by Jib that was causing the problem is actually `image`, `docker.image` here is the workaround. The mock gateway uses `mock-gateway.docker.image`, so setting `docker.image` will only change the one we want.

If you try using `image` you can actually still reproduce the original problem. With the Jib plugin, `-Dimage` will (for some reason) take precedence over the explicitly configured `<image>${docker.image}</image>` config setting in the parent pom. Both plugin executions will pick it up, use the same tag, and I don't think there's any way to prevent that.

parent-pom/pom.xml
geraldhumphries: I was thinking about using the enforcer plugin. Set the value here to something like `INVALID`, then use the enforcer plugin to check for that value and fail the build with a message stating that it must be set in the child pom. That would make it as clear and explicit as possible, so maintainers won't need to dig very deep to know when they need to add that property. It might take some work to make sure it only uses the enforcer plugin when necessary, e.g. we don't want it for shared-lib.

What do you think? I will probably just add a comment if that would take too much time.

.github/workflows/develop.yml
```diff
@@ -1,218 +1,191 @@
...
-          enable-sonar-scan: ${{ github.event.ref == 'refs/heads/develop' }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          sonar-token: ${{ secrets.SONAR_TOKEN }}
+      - name: Create & set up cluster
+        run: |          
+          # Setup localhost DNS
+          sudo echo "127.0.0.1 onmshs" | sudo tee -a /etc/hosts
+          
+          # Build cluster with Horizon Stream installed.
+          ./install-local.sh cicd onmshs ${{ inputs.image-tag || github.sha }} ${{ inputs.image-registry || 'ghcr.io' }}/opennms-cloud
```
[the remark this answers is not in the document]
geraldhumphries: Done! Thanks for the tip, will remember that. There are so many little things to think about with these scripts.

parent-pom/pom.xml
geraldhumphries: I changed the default value to "INVALID" with a comment. If the prop isn't set, the build will fail because docker images can't be tagged with uppercase letters. That's good enough for me.

artnaseef (approved):

PR #625, opened 2023-01-16
DC-456 Modularize metrics-processor

FrancescoBuc86:
- Create two modules, one for flows and one for timeseries, and move the existing classes accordingly

https://opennms.atlassian.net/browse/DC-456


metrics-processor/flows/pom.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+    <parent>
+		<groupId>org.opennms.horizon</groupId>
+		<artifactId>metrics-processor</artifactId>
+		<version>0.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
```
geraldhumphries, line 10: nit: No need for relativePath when following the standard Maven dir structure (parent pom in the parent folder). This is what it defaults to.

metrics-processor/pom.xml
```diff
@@ -1,26 +1,48 @@
...
     <parent>
-		<groupId>org.opennms.horizon</groupId>
-		<artifactId>horizon-parent</artifactId>
-		<version>0.1.0-SNAPSHOT</version>
-        <relativePath>../parent-pom</relativePath>
-	</parent>
+        <groupId>org.opennms.horizon</groupId>
+        <artifactId>horizon-parent</artifactId>
+        <version>0.1.0-SNAPSHOT</version>
+        <relativePath>../parent-pom/pom.xml</relativePath>
```
geraldhumphries, line 10: No need to explicitly point at pom.xml, the folder is enough.

metrics-processor/flows/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+	</parent>
+
+    <name>OpenNMS Horizon Stream :: Flows</name>
```
geraldhumphries, line 13: nit: A name like `OpenNMS Horizon Stream :: Metrics Processor :: Flows` would follow our convention better.

metrics-processor/flows/src/main/java/org/opennms/flows/Main.java
```diff
@@ -0,0 +1,6 @@
+package main.java.org.opennms.flows;
+
+public class Main {
+    public static void main(String[] args) {
+    }
```
geraldhumphries, line 5: We might need a third "main" module to contain the main class/method with the `@SpringBootApplication` and the Jib plugin. We could probably just use `@ComponentScan` to point at the flows/timeseries modules to get the Spring beans.

metrics-processor/timeseries/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+    <parent>
+		<groupId>org.opennms.horizon</groupId>
+		<artifactId>metrics-processor</artifactId>
+		<version>0.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+	</parent>
+
+    <name>OpenNMS Horizon Stream :: Time Series</name>
```
geraldhumphries, line 13: nit: Same comments as above regarding naming conventions and relativePath.

metrics-processor/flows/pom.xml
FrancescoBuc86: Right!

metrics-processor/pom.xml
FrancescoBuc86: true

metrics-processor/flows/pom.xml
FrancescoBuc86: Totally, I'll update it

metrics-processor/flows/src/main/java/org/opennms/flows/Main.java
FrancescoBuc86: Right, I thougt about that too but decided at first to leave it for another PR, but probably you're right, I'll insert it directly in this pr

metrics-processor/timeseries/pom.xml
FrancescoBuc86: Sure

metrics-processor/timeseries/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-core</artifactId>
+            <version>${dropwizard.metrics.version}</version>
+        </dependency>
+    </dependencies>
+
+	<build>
+		<plugins>
```
patrick-schweizer, line 72: I wonder if we can't / should pull the plugin definitions up into the parent pom?

patrick-schweizer (commented):
Lets's finish the PR comments and then take another look..

metrics-processor/timeseries/pom.xml
FrancescoBuc86: yep we could

geraldhumphries: The horizon-parent pom already includes a pluginManagement section with similar config. We can just define the plugin in the individual projects like this:

```
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
            </plugin>
```

metrics-processor/main-metrics/src/main/java/org/opennms/TSDataProcessorApplication.java
```diff
@@ -1,4 +1,4 @@
-package org.opennms.horizon.tsdata;
+package org.opennms;
```
patrick-schweizer, line 1: not sure about the name convention for packages but it seems the other packages seem to start with:
_org.opennms.horizon_

metrics-processor/pom.xml
```diff
@@ -123,27 +56,22 @@
...
+                            <artifactId>lombok</artifactId>
+                        </exclude>
+                    </excludes>
+                </configuration>
             </plugin>
             <plugin>
                 <groupId>kr.motd.maven</groupId>
                 <artifactId>os-maven-plugin</artifactId>
-                <version>${os-maven-plugin.version}</version>
+                <version>1.6.2</version>
```
patrick-schweizer, line 74: do we need explicit definition here?

metrics-processor/pom.xml
```diff
@@ -156,7 +84,7 @@
             <plugin>
                 <groupId>org.xolstice.maven.plugins</groupId>
                 <artifactId>protobuf-maven-plugin</artifactId>
-                <version>${protobuf-maven-plugin.version}</version>
+                <version>0.6.1</version>
```
patrick-schweizer, line 87: same here?

metrics-processor/timeseries/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+    <properties>
+        <java.version>17</java.version>
+    </properties>
+
+	<dependencies>
+        <dependency>
+            <groupId>org.springframework.kafka</groupId>
+            <artifactId>spring-kafka</artifactId>
+        </dependency>
+		<dependency>
```
patrick-schweizer, line 29: formatting
shouldn't lombok move up to parent pom?
I also believe we should set the scope to provide: _<scope>provided</scope>_

patrick-schweizer (approved):
Looks now good to me! :-)

metrics-processor/flows/src/main/java/org/opennms/horizon/flows/Main.java
```diff
@@ -0,0 +1,6 @@
+package org.opennms.horizon.flows;
+
+public class Main {
```
geraldhumphries, line 3: Is this complete?

freddy61025 (changes requested):
Please make those two main class more meaningful, if nothing implemented inside is that possible remove them or at least leave some comment about why it need to be there.
metrics-processor/flows/src/main/java/org/opennms/horizon/flows/Main.java
metrics-processor/timeseries/src/main/java/org/opennms/horizon/tsdata/Main.java

metrics-processor/flows/src/main/java/org/opennms/horizon/flows/Main.java
FrancescoBuc86: removed

FrancescoBuc86:
> Please make those two main class more meaningful, if nothing implemented inside is that possible remove them or at least leave some comment about why it need to be there. metrics-processor/flows/src/main/java/org/opennms/horizon/flows/Main.java metrics-processor/timeseries/src/main/java/org/opennms/horizon/tsdata/Main.java

Yep, removed both of them

metrics-processor/flows/src/main/java/org/opennms/horizon/flows/Main.java
FrancescoBuc86: Actually without it the pr checker will complain that no main class is available, so I'll just leave it there, with comments though

geraldhumphries: If this PR is merged, will the container still build and start properly?

FrancescoBuc86: Hey Gerald, I didn't notice specific issues, but if you do let me know.

PR #633, opened 2023-01-17
HS-853 Tag manager feature

chinh-cle:
## Jira link(s)
- https://issues.opennms.org/browse/HS-853-tag-manager-feature

## Checklist


ui/src/styles/mediaQueries.scss
```diff
@@ -0,0 +1,33 @@
...
+  lg: 992px,
+  xl: 1200px,
+  xxl: 1400px
+) !default;
+
+@use "sass:map";
+
+@mixin screen-xs {
+  @media (max-width: map.get($grid-breakpoints, 'xs')) { @content; }
+}
```
mikewrosey, line 18: is this file a duplicate of mediaQueriesMixins.scss ? looks like it can be removed

ui/src/store/Queries/tagQueries.ts
```diff
@@ -0,0 +1,45 @@
...
+  const tagsSearched = ref([] as Tag[])
+
+  const tagStore = useTagStore()
+
+  const fetchTags = async () => {
+    const { data, execute } = useQuery({
+      query: ListTagsDocument,
+      fetchOnMount: false,
+      cachePolicy: 'network-only'
+    })
```
mikewrosey, line 16: is it possible to move the useQuery hook outside this fn? Seems like it shouldn't need to run every time and that just calling execute like below would work.

ui/src/components/Inventory/InventoryMonitoredNodesTabContent.vue
```diff
@@ -1,74 +1,279 @@
...
+  display: flex;
+  justify-content: end;
+  padding: var(variables.$spacing-l) 0;
+  min-width: vars.$min-width-smallest-screen;
+}
+ul.cards {
   display: flex;
   flex-flow: row wrap;
-  gap: 1rem;
+  justify-content: space-between;
```
mikewrosey, line 247: Might not want `space-between` here, or it could look like this.
screenshot-localhost_3009-2023 03 08-10_13_07

ui/src/components/Inventory/InventoryTagManagerCtrl.vue
```diff
@@ -1,26 +1,40 @@
...
+}
+
+const icon = markRaw({ ExpandLess, ExpandMore })
 </script>
+
+<style lang="scss" scoped>
+@use '@/styles/btns.scss';
+
+.btn-as-input {
+  @include btns.btn-looks-as-select-input;
```
mikewrosey, line 38: Some discrepancy in text here. Should be the same with   @include typography.body-small
screenshot-localhost_3009-2023 03 08-10_35_18

ui/src/components/Common/InputButtonPopover.vue
```diff
@@ -0,0 +1,96 @@
...
+        v-on="on"
+        class="add-btn"
+      >
+        <Icon :icon="icon" />
+      </FeatherButton>
+    </template>
+  </FeatherPopover>
+</template>
+
+<script lang="ts" setup>
```
mikewrosey, line 32: is this a common comp that will be useful later? don't see it used anywhere

mikewrosey:
Some issues with adding tags that I encountered:
1. I have two nodes and two tags that were added through discovery. When I added a tag to one of the nodes it added it to both of the nodes. But since that first time I haven't been able to recreate the problem.
2. When I added a new tag here and then added it to the node, it didn't show up after saving. It only showed up after refreshing the page. Maybe something to do with the order of the callbacks, because it is also difficult to recreate.
3. Was able to add more than one tag with the same name somehow.
4. Sometimes when adding a new tag it would automatically select random tags, as if something from state management was being applied.

screenshot-localhost_3009-2023 03 08-11_11_16


Also, is there any way to remove tags or is that later?

ui/src/components/Inventory/InventoryNodeTagEditOverlay.vue
```diff
@@ -0,0 +1,147 @@
...
+  },
+  isSelected: {
+    type: Boolean,
+    default: false
+  }
+})
+
+const isChecked = ref(false)
+watchEffect(() => {
+  isChecked.value = props.node.isNodeOverlayChecked as boolean
```
mikewrosey, line 62: does isNodeOverlayChecked need to be optional below? seems like it's always t/f and we can probably make it required and remove the `as boolean` on the line above

```
interface NodeContent {
  id: number
  label: string | undefined
  status: string
  metrics: Chip[]
  anchor: Anchor
  isNodeOverlayChecked?: boolean
}
```

ui/src/store/Mutations/tagMutations.ts
```diff
@@ -0,0 +1,17 @@
+import { defineStore } from 'pinia'
+
+export const useTagMutations = defineStore('tagMutations', () => {
+  const editTag = (val: string, toAddTags: boolean) => {
+    if (toAddTags) {
+      // TODO: gql query to add tags to node
```
mikewrosey, line 6: do we still need this store/file if the the add tags to node (and presumably delete when it comes), is in the nodeMutations file?

ui/src/components/Inventory/InventoryFilter.vue
```diff
@@ -164,35 +182,42 @@ const severity: ISelectDropdown = {
...
+  size: 2
 }
-const expandCollapse = ref(computed<IIcon>(() => {
-  expandCollapseBtn.value = inventoryStore.isFilterOpen ? 'collapse-btn' : 'expand-btn'
-  return inventoryStore.isFilterOpen ? collapse : expand
-}))
-
+const expandCollapse = ref(
+  computed<IIcon>(() => {
+    expandCollapseBtn.value = inventoryStore.isFilterOpen ? 'collapse-btn' : 'expand-btn'
```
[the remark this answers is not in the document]
chinh-cle: ya it's used in inventoryFilter test file

mikewrosey (approved):

PR #649, opened 2023-01-20
Jira/hs 872 fix release cicd wrkflw since merg pipeline refactor

jaymberry:
## Description
Since pipeline refactor, the release workflow is in need of work to get back running again. 

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-872

## Checklist


.github/workflows/develop.yml
```diff
@@ -4,13 +4,14 @@ on:
...
-    branches:
-      - 'develop'
+  #pull_request:
+  #  # This workflow is triggered instead of individual workflows that filter based
+  #  # on paths.  If we want to fully separate individual workflows out of this one,
+  #  # we need fully reproducible builds.  Just use this one for now. The downside
+  #  # is it will run redundant workflows for code that didn't change.
+  #  #types: [labeled]
+  #  branches:
+  #    - 'develop'
```
geraldhumphries, line 14: This should be reenabled

.github/workflows/release.yml
```diff
@@ -1,83 +1,223 @@
 name: release
 
 on:
-  workflow_dispatch:
+  push:
+    branches:
+      - 'release'
```
geraldhumphries, line 6: This workflow will run when you push to a branch called `release`, is that correct or do you just use workflow_call?

geraldhumphries: In the future, I suggest something like tag-based releases, with docker/metadata-action to extract the release tag.

Like this:
```
  push:
    branches:
      - 'develop'
    tags:
      - 'v*'

...

steps:
     - name: Generate Docker metadata
        id: docker-metadata
        uses: docker/metadata-action@v4.3.0
        with:
          images: |
            ${{ env.IMAGE_REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE }}
           tags: |
             type=semver,pattern={{version}}
```

Tags like `v1.0.0` can be pushed and produce image tags with that version too.

.github/workflows/feature-datachoices.yml
```diff
@@ -27,7 +27,7 @@ env:
 
 jobs:
   build-and-test:
-    name: Build and test
+    name: Build and test - Data Choices
```
geraldhumphries, line 30: nit: `DataChoices`

.github/workflows/feature-alarm.yml
```diff
@@ -27,7 +27,7 @@ env:
 
 jobs:
   build-and-test:
-    name: Build and test
+    name: Build and test - Alarm
```
geraldhumphries, line 30: The only way I can think of that would get rid of the need to do this would be to combine the develop and release workflows, but I'm not so sure about that. The better option is to totally separate the image build workflows from the develop and release workflows. We can chat about that later.

.github/workflows/feature-parent-pom.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build-and-test:
-    name: Build and test
+    name: Build and test - Parent Pom
```
geraldhumphries, line 12: nit: Parent POM

geraldhumphries (changes requested):

.github/workflows/release.yml
jaymberry: Because we are trying to move to trunk based dev, we want to keep it automated and not through ui.

.github/workflows/release.yml
```diff
@@ -1,83 +1,223 @@
...
+
+          docker images
+
+      - name: Publish images
+        run: |
+
+          docker images
+   
+          #docker push opennms/${{ matrix.image }}:latest
+          docker push opennms/${{ matrix.image }}:test-${{ env.RELEASE_TAG }}
```
geraldhumphries, line 222: Should probably get rid of the `test-` prefix

.github/workflows/release.yml
```diff
@@ -1,83 +1,223 @@
...
+
+          ls -1R docker-assembly/
+
+          docker buildx create --use --driver docker-container --name maven
+
+          # Not sure how to push this without build for multi-arch.
+          docker buildx build --push \
+            --tag opennms/horizon-stream-minion:latest \
+            --tag opennms/horizon-stream-minion:test-${{ env.RELEASE_TAG }} \
+            docker-assembly/target/docker/opennms/horizon-stream-minion/latest/build/ \
```
geraldhumphries, line 129: test- prefix here

.github/workflows/release.yml
```diff
@@ -1,83 +1,223 @@
...
+      - name: Pull images
         run: |
+
+          docker pull ghcr.io/opennms-cloud/${{ matrix.image }}:${{ env.IMAGE_TAG }}
+
           # Tag and save the image as a file to be published. Both release version and latest.
-          docker tag ${{ matrix.image }}:${{ needs.tag.outputs.image-tag }} ${{ matrix.image }}:latest
-          docker push ${{ matrix.image }}:latest
+          docker tag ghcr.io/opennms-cloud/${{ matrix.image }}:${{ env.IMAGE_TAG }} opennms/${{ matrix.image }}:latest
+          docker tag ghcr.io/opennms-cloud/${{ matrix.image }}:${{ env.IMAGE_TAG }} opennms/${{ matrix.image }}:test-${{ env.RELEASE_TAG }}
```
geraldhumphries, line 212: Another test- prefix

geraldhumphries (changes requested):

geraldhumphries (approved):

PR #651, opened 2023-01-23
HS-704: Add generic configuration table and inventory

BenjaminJ:
HS-704

## Description
Add generic configuration table and inventory

## Jira link(s)
- https://issues.opennms.org/browse/HS-704

## Checklist


shared-lib/inventory/src/main/proto/configuration.proto
```diff
@@ -0,0 +1,59 @@
...
+  repeated ConfigurationDTO configurations = 1;
+}
+
+message ConfigurationKeyAndLocation {
+  string key = 1;
+  string location = 2;
+}
+
+service ConfigurationService {
+  rpc listConfigurationsByTenantId(google.protobuf.Empty) returns (ConfigurationList) {};
```
cgorantla, line 55: tenantId is given, no need to add that as part of service method name.
 It is passed in header so we don't that to be part of `ConfigurationDTO`

cgorantla: tenantId can be removed in all of the methods

cgorantla (commented):
looks good. few comments

inventory/src/main/java/org/opennms/horizon/inventory/repository/ConfigurationRepository.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.opennms.horizon.inventory.model.Configuration;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface ConfigurationRepository extends JpaRepository<Configuration, Long> {
+    List<Configuration> findByTenantId(String tenantId);
+
+    List<Configuration> findByKey(String tenantId, String key);
```
yangli2014, line 14: According to the database schema, the tenantId and key pair are unique, so this will return only one Configuration.

inventory/src/main/java/org/opennms/horizon/inventory/model/Configuration.java
```diff
@@ -0,0 +1,71 @@
...
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.validation.constraints.NotNull;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@RequiredArgsConstructor
```
yangli2014, line 48: This only works with private final fields.

yangli2014 (commented):
You might need to merge the develop branch into your branch since there are a lot of changes.

inventory/src/main/java/org/opennms/horizon/inventory/model/Configuration.java
BenjaminJ: Do you suggest to make the fields final, or to remove the annotations ?

yangli2014: I prefer to remove the annotation @RequiredArgsConstructor

inventory/src/main/java/org/opennms/horizon/inventory/repository/ConfigurationRepository.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Repository
+public interface ConfigurationRepository extends JpaRepository<Configuration, Long> {
+    List<Configuration> findByTenantId(String tenantId);
+
+    Optional<Configuration> findByTenantIdAndKey(String tenantId, String key);
+
+    List<Configuration> findByTenantIdAndLocation(String tenantId, String location);
+
+    Optional<Configuration> getByTenantIdAndKeyAndLocation(String tenantId, String key, String location);
```
yangli2014, line 18: I am not sure if we need this method, since the key and tenantId pair is a unique key in the DB, the location value is not necessary.

cgorantla (approved):
LGTM!

PR #653, opened 2023-01-23
HS-789: Add node tag manangement: add and remove a tag from node(s)

thomasbigger584:
## Description
HS-789: Add node tag manangement: add and remove a tag from node(s)

Tag has a many to many relationship with Node with a degree of idempotency on create/remove

Creating a Tag will associate it with a Node, if its not already associated.
Removing a Tag will remove the Tag association with the node, and if there is no other associations it will delete the Tag.

Deleting a node will delete the associations of the tag, if there is no other associations it will delete the tag also.

Endpoints add/remove a list of tags.

## Jira link(s)
- https://issues.opennms.org/browse/HS-789

## Checklist

### DIFF
--- a/inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
@@ -40,7 +40,7 @@ import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 
-import javax.validation.constraints.NotNull;
+import jakarta.validation.constraints.NotNull;
 import java.util.ArrayList;
 import java.util.List;
 
--- a/inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
@@ -1,0 +1,43 @@
+package org.opennms.horizon.inventory.model;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.JoinTable;
+import jakarta.persistence.ManyToMany;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+
+import javax.validation.constraints.NotNull;
+import java.util.ArrayList;
+import java.util.List;
+
+@Getter
+@Setter
+@RequiredArgsConstructor
+@Entity
+public class Tag {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+
+    @NotNull
+    @Column(name = "tenant_id")
+    private String tenantId;
+
+    @NotNull
+    @Column(name = "name")
+    private String name;
+
+    @ManyToMany
+    @JoinTable(
+        name = "node_tag",
+        joinColumns = @JoinColumn(name = "tag_id"),
+        inverseJoinColumns = @JoinColumn(name = "node_id"))
+    List<Node> nodes = new ArrayList<>();
+}
--- a/inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
@@ -1,3 +1,31 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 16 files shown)


inventory/src/main/java/org/opennms/horizon/inventory/repository/TagRepository.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Query("SELECT tag " +
+        "FROM Tag tag " +
+        "JOIN tag.nodes node " +
+        "WHERE tag.tenantId = :tenantId " +
+        "AND node.id = :nodeId " +
+        "AND tag.name = :name")
+    Optional<Tag> find(@Param("tenantId") String tenantId,
+                       @Param("nodeId") Long nodeId,
+                       @Param("name") String name);
```
yangli2014, line 57: We might need to find all tags by nodeId and tenantId

inventory/src/main/java/org/opennms/horizon/inventory/repository/TagRepository.java
```diff
@@ -0,0 +1,58 @@
...
+
+@Repository
+public interface TagRepository extends JpaRepository<Tag, Long> {
+    Optional<Tag> findByTenantIdAndId(String tenantId, Long id);
+
+    @Query("SELECT tag " +
+        "FROM Tag tag " +
+        "WHERE tag.tenantId = :tenantId " +
+        "AND tag.name = :name")
+    Optional<Tag> find(@Param("tenantId") String tenantId,
```
yangli2014, line 46: It is not a big issue, but I prefer using JPA repo standard syntax findByxxxAndxxx to make it simple.

inventory/src/main/java/org/opennms/horizon/inventory/repository/TagRepository.java
thomasbigger584: yeah this is not done as its not related to this story

inventory/src/main/java/org/opennms/horizon/inventory/repository/TagRepository.java
thomasbigger584: will update

yangli2014 (commented):
Also, I think there might be a use case the user wants to add more than one tag to a node. So the user has to add one by one, it is better if we can let user add more want with some syntax such as the semicolon-separated string.

cgorantla (commented):
Looks pretty good to me.

It aligns with our current thinking that tags are very generic.

As @yangli2014 said, we may want to add/delete multiple tags at the same time. 
May be it would be good to update services to accept lists.

thomasbigger584:
> Looks pretty good to me.
> 
> It aligns with our current thinking that tags are very generic.
> 
> As @yangli2014 said, we may want to add/delete multiple tags at the same time. May be it would be good to update services to accept lists.

sure will make it a list instead

inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
```diff
@@ -0,0 +1,71 @@
...
+    @NotNull
+    @Column(name = "name")
+    private String name;
+
+    @ManyToMany
+    @JoinTable(
+        name = "node_tag",
+        joinColumns = @JoinColumn(name = "tag_id"),
+        inverseJoinColumns = @JoinColumn(name = "node_id"))
+    List<Node> nodes = new ArrayList<>();
```
cgorantla, line 70: nit :
```suggestion
    private List<Node> nodes = new ArrayList<>();
```

inventory/src/main/java/org/opennms/horizon/inventory/service/TagService.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Transactional
+    public List<TagDTO> addTags(String tenantId, TagCreateListDTO request) {
+        if (request.getTagsList().isEmpty()) {
+            return Collections.emptyList();
+        }
+        Node node = getNode(tenantId, request.getNodeId());
+
+        return request.getTagsList().stream()
+            .map(tagCreateDTO -> createSingle(tenantId, node, tagCreateDTO))
```
cgorantla, line 65:
```suggestion
            .map(tagCreateDTO -> createTag(tenantId, node, tagCreateDTO))
```

inventory/src/main/java/org/opennms/horizon/inventory/service/TagService.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Transactional
+    public void removeTags(String tenantId, TagRemoveListDTO request) {
+        Node node = getNode(tenantId, request.getNodeId());
+
+        request.getTagIdsList().stream()
+            .map(Int64Value::getValue)
+            .map(tagId -> getTag(tenantId, tagId))
+            .toList()
+            .forEach(tag -> removeSingle(node, tag));
```
cgorantla, line 77:
```suggestion
            .forEach(tag -> removeTagFromNode(node, tag));
```

cgorantla (commented):
LGTM!  few minor comments

yangli2014:
We also need to list all tags for giving node id to avoid unnecessary network traffic.

inventory/src/main/java/org/opennms/horizon/inventory/model/Tag.java
thomasbigger584: good catch

thomasbigger584:
> We also need to list all tags for giving node id to avoid unnecessary network traffic.

sure, can do that in another story

cgorantla (approved):

PR #658, opened 2023-01-24
DC-456 Add Adapter and Enrichment Classes to metrics processor

FrancescoBuc86:
https://opennms.atlassian.net/browse/DC-456


patrick-schweizer (approved):
It's hard to review 364 files :-) But given that they all were copied from OpenNMS (and thus were approved once) I think it is ok.
I am not sure how nitpicky we want to be about the copyright header. I would suggest a pragmatic approach: If the classes stay as they are (just copying) we leave the header as is. If we change them after copying we fix the copyright year.

shared-lib/horizon-common-lang/pom.xml
```diff
@@ -20,6 +20,7 @@
     </description>
 
 
+
```
patrick-schweizer, line 23: why new line?

FrancescoBuc86:
@geraldhumphries @freddy61025 Almost all the classes are copied from classic, like the ones for the adapter / enrichment / classification purposes. Other classes have been moved from the flows-parser module to the shared-lib, so that we can reference them both from minion and metrics-processor. These classes are the ones needed for the UDP/TCP Listeners/Parsers.

cgorantla:
may be I'm missing something here.

My two cents: 
I think we wanted to move adapter to Minion on Horizon Stream 
If we are moving adapter to Cloud, we should create separate service to handle flows. It may not be needed to club with Metric processor.

geraldhumphries:
> may be I'm missing something here.
> 
> My two cents: I think we wanted to move adapter to Minion on Horizon Stream If we are moving adapter to Cloud, we should create separate service to handle flows. It may not be needed to club with Metric processor.

Chandra and I discussed this so this is just summing up that conversation. The adapter classes should probably be moved to the minion, but the flows processing shouldn't be split into a new service.

FrancescoBuc86:
Hi @geraldhumphries and @cgorantla, thanks for your comments. So the structure in this pr looks like this:
-- metrics-processor 
    -- flows
        - adapter 
        - processing
-- minion
   -- minion-flows  
      -- listeners
      -- parsers
-- shared-lib
   -- horizon-common-flows
       -- (containing all flows classes used both in minion and metrics processor)

In this way there's no dependency of the metrics-processor with the minion. 


Now what you're saying is that the structure should look like this:
-- metrics-processor 
    -- flows
       - (empty)
-- minion
   -- minion-flows  
      -- listeners
      -- parsers
      -- adapter 
      -- processing
-- shared-lib
   -- horizon-common-flows
       - (empty)
 
Is this correct?

geraldhumphries:
@FrancescoBuc86 to clarify, are the classes in `horizon-common-flows` only shared because of the adapter? E.g. if the adapter is moved to minion, there's no need to put them in shared-lib?

cgorantla:
IMO, 
Adapter functionality should still live on Minion.  And whatever we pulled into shared-lib should also exist on Minion as I see they are mostly protocol specific.

Only https://github.com/OpenNMS/opennms/blob/develop/features/telemetry/protocols/netflow/transport/src/main/proto/netflow.proto. should live shared-lib and generate  Java files out of it for FlowMessage.

Coming to processing (Enrichment) , I think this also needs to go into Minion ( if inventory can be made available on Minion)

freddy61025:
> IMO, Adapter functionality should still live on Minion. And whatever we pulled into shared-lib should also exist on Minion as I see they are mostly protocol specific.
> 
> Only https://github.com/OpenNMS/opennms/blob/develop/features/telemetry/protocols/netflow/transport/src/main/proto/netflow.proto. should live shared-lib and generate Java files out of it for FlowMessage.
> 
> Coming to processing (Enrichment) , I think this also needs to go into Minion ( if inventory can be made available on Minion)

We are currently in Phase 1 of migrating flow from classic to BTO. Our plan is listener & parser in minion, enrichment in metrics processor. And storage is still in elastic search. After POC, we will start migrating enrichment into minion.

cgorantla:
> > IMO, Adapter functionality should still live on Minion. And whatever we pulled into shared-lib should also exist on Minion as I see they are mostly protocol specific.
> > Only https://github.com/OpenNMS/opennms/blob/develop/features/telemetry/protocols/netflow/transport/src/main/proto/netflow.proto. should live shared-lib and generate Java files out of it for FlowMessage.
> > Coming to processing (Enrichment) , I think this also needs to go into Minion ( if inventory can be made available on Minion)
> 
> We are currently in Phase 1 of migrating flow from classic to BTO. Our plan is listener & parser in minion, enrichment in metrics processor. And storage is still in elastic search. After POC, we will start migrating enrichment into minion.

AFAIK, Adapter still doesn't do enrichment with NodeInfo.  It just moves protocol specific to common flow proto (netflow.proto). That's why I thought it could still go into Minion.  
Processing may be part of  metrics-processor and then can go into Minion at later point.

FrancescoBuc86:
> @FrancescoBuc86 to clarify, are the classes in `horizon-common-flows` only shared because of the adapter? E.g. if the adapter is moved to minion, there's no need to put them in shared-lib?

Hi @geraldhumphries, yes, because of the adapter and the processing. If everything is moved into minion, than we don't need anymore any shared-lib

PR #661, opened 2023-01-24
HS-876 - Process to pull base img put in github prvt registry

jaymberry:
## Description
Process to pull base img put in github private registry.

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-876

## Checklist


.github/base-image-tag-ref.json
```diff
@@ -0,0 +1,5 @@
+{
+  "image_tag_java": "eclipse-temurin:17-jre",
+  "test": "1",
```
geraldhumphries, line 3: Is "test" used for anything?

jaymberry: Yes, I use it for triggering ci-cd tests. I have these spread out through the repo, we can get rid of them at some point, but come in handy while still in early stage of development.
 
I will make a ticket to remvoe these by end of 2nd Quarter.

geraldhumphries (approved):
Approved. There is just a small amount of leftover testing data that should be removed but that should not block the merge.

.github/workflows/base-image-update.yml
```diff
@@ -0,0 +1,37 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Pull Image & Push
+        run: |
+
+          IMAGE=$(cat .github/base-image-tag-ref.json | jq .image_tag_java | sed 's/\"//g')
+
+          docker pull $IMAGE
```
artnaseef, line 32: I recommend quoting `"$IMAGE"` here - if spaces end up in the variable somehow, the quotes will make the errors more obvious and make sure the results are "correct".

artnaseef (approved):

.github/workflows/base-image-update.yml
jaymberry: ok, will implement. Thank you

PR #671, opened 2023-01-26
HS-889 Introduce ignite caches for Digital Twin and TaskSet processing

splatch:
## Description
This is proposal which simplifies some minion gateway internals by removal of several components which did not serve much functionality but contributed to overall complexity of flow.
Additionally I switched `AbstractTwinPublisher` to rely on Ignite cache which permits to distribute cache contents across network and store them in database as a longer term storage. My tests showed that its not the fastest way to store data, however it solved problem of TaskSet not being communicated to minions after gateway was restarted.

Updated code calls grpc twin publisher directly from internal grpc endpoint as twin will retain taskset and answer with taskset contents if new connection is made. Adjusted cache blob store configuration works only with PostgreSQL.

Some remarks - the CacheJdbcBlobStoreFactory is being serialized and de-serialized. This means that initially injected dataSource object will be gone. Internally the cache blob store factory attempt to lookup spring datasource bean by name. Once it is provided whole thing starts to roll.

Part of job which I took with this PR was separation of Ignite configuration from spring java config which we use to setup minion gateway. This is caused by basic fact that ignite configuration is memory dependent and it should go in line with JVM memory settings. Default configuration, both for standalone operation as for tilt/k8s is 512MB as maximum cache size. Its pretty small but I did not intend to stress developer boxes too much.

## Jira link(s)
- https://issues.opennms.org/browse/HS-889

## Checklist


artnaseef (approved):

minion-gateway/main/pom.xml
```diff
@@ -122,6 +126,14 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+        </dependency>
+        <dependency>
```
cgorantla, line 133: do we need h2database as well ?

cgorantla (approved):
LGTM! 

Awesome. Let's get this in!

PR #678, opened 2023-01-27
Add the rapid-build profile.  When enabled, Integration Tests (failsa…

artnaseef:
…fe) and Karaf feature validation are disabled.

## Jira link(s)
- https://issues.opennms.org/browse/HS-901

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,16 @@
 **Horizon Stream** is a new distribution of OpenNMS, inspired from the existing platform
 
 See [Getting Started on the Wiki](https://github.com/OpenNMS/horizon-stream/wiki/Getting-Started) for development instructions.
+
+# Rapid Development Profile
+
+In order to facilitate more rapid development,
+the Maven profile named rapid-build
+disables Intergation Tests (Failsafe) and
+the Karaf Maven Plugin's validation of features.
+
+* It can be enabled in the following ways:
+  * In standard Maven ways:
+    * Manually add `-P rapid-build` to individual `mvn` commands
+    * Add the `rapid-build` profile to the user's `settings.xml` file in the `<activeProfiles>` section
+  * Create a file at `${user.home}/.OPENNMS-RAPID-BUILD`
--- a/parent-pom/pom.xml
+++ b/parent-pom/pom.xml
@@ -22,6 +22,8 @@
         <sonar.organization>opennms</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
 
+        <skipITs>false</skipITs>
+
         <!-- Jib properties -->
         <application.docker.image>${application.docker.image.name}:${application.docker.image.tag}</application.docker.image>
         <!-- application.docker.image.name MUST be set by child modules that use Jib. -->
@@ -82,6 +84,7 @@
         <jsr250.version>1.0</jsr250.version>
         <jsr305.version>3.0.2</jsr305.version>
         <karaf.version>4.4.2</karaf.version>
+        <karaf.maven-plugin.version>${karaf.version}</karaf.maven-plugin.version> <!-- use separate property in case a need to separate from the base Karaf version arises -->
         <keycloak.version>18.0.0</keycloak.version>
         <liquibaseVersion>4.17.0</liquibaseVersion>
         <log4j2.version>2.17.1</log4j2.version>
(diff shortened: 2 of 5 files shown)


cgorantla:
mvn clean install -DskipTests=true already skips all tests (including ITs).
we just need to disable karaf verifications by default and let user specify (mvn verify) to validate features.
what do you think ?

parent-pom/pom.xml
```diff
@@ -927,6 +931,14 @@
                     </configuration>
                 </plugin>
 
+                <plugin>
+                    <groupId>org.apache.karaf.tooling</groupId>
+                    <artifactId>karaf-maven-plugin</artifactId>
+                    <version>${karaf.maven-plugin.version}</version>
+                    <configuration>
+                        <skip>${skipKarafVerify}</skip>
```
geraldhumphries, line 939: Does this skip just feature validation, or execution of the whole plugin?

parent-pom/pom.xml
```diff
@@ -22,6 +22,9 @@
         <sonar.organization>opennms</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
 
+        <skipITs>false</skipITs>
+        <skipKarafVerify>false</skipKarafVerify>
```
geraldhumphries, line 26: There is a built-in undocumented user property on the karaf-maven-plugin we *could* use. Setting `features.verify.skip` to true will skip feature verification. Since it is undocumented, I don't know if we should use it. I think using our own property like this will be safer.

minion/assembly/pom.xml
```diff
@@ -97,6 +97,7 @@
                 <version>${karaf.version}</version>
                 <extensions>true</extensions>
                 <configuration>
+                    <skip>${skipKarafVerify}</skip>
```
geraldhumphries, line 100: If this is already in parent-pom's pluginManagement, do we need this?

parent-pom/pom.xml
artnaseef: Oh good question - let me check.  It must be the entire plugin...

parent-pom/pom.xml
artnaseef: Undocumented isn't too much of a concern.  If it is standard and maintained, I think we can use it.  I'll check on that as well.

minion/assembly/pom.xml
artnaseef: Ahh correct - this is a left-over and should be removed.

README.md
```diff
@@ -3,3 +3,17 @@
...
+In order to facilitate more rapid development,
+the Maven profile named rapid-build
+disables Intergation Tests (Failsafe) and
+the Karaf Maven Plugin's validation of features.
+
+* It can be enabled in the following ways:
+  * In standard Maven ways:
+    * Manually add `-P rapid-build` to individual `mvn` commands
+    * Add the profile to the user's `settings.xml` file in the `<activeProfiles>` section
+    * Using environment variable setting `MAVEN_OPTS="-Prapid-build"`
```
artnaseef, line 18: NOTE this is not right - MAVEN_OPTS are JVM options.

parent-pom/pom.xml
artnaseef: In testing, it appears to only affect the feature validation.  Not sure why.

artnaseef: Found this in the code - looks like <skip> is only feature validation skip for better or worse:

    @Parameter(property` = "skip", defaultValue = "${features.verify.skip}")
        protected boolean skip;

artnaseef:
> mvn clean install -DskipTests=true already skips all tests (including ITs). we just need to disable karaf verifications by default and let user specify (mvn verify) to validate features. what do you think ?

I prefer that the tests and karaf verifications be enabled by default, but provide developers an easy way to disable them "by default" - the word default gets overloaded here.  To clarify, the default handling when running the `mvn` can be configured by explicit action to enable the build profile (e.g. by adding it to `settings.xml`).  At the same time, we want the "default" for that setting to be "enabled" so the tests and verifications run.

Note that the karaf maven plugin is tied to the `verify` phase, and always runs, or is skipped, during that phase.

geraldhumphries (approved):

PR #683, opened 2023-01-31
HS-895: add tenant id to notifications

james8hut:
## Description
Multi tenancy for notification component

## Jira link(s)
- https://issues.opennms.org/browse/HS-895

## Flagged for review
Uses Hibernate TenantId annotation which does all the work for us.

Question: In TenantIdentifierResolver should we always have a tenant id? I'm guessing we should, but don't want to do a load of rework in the tests if this is not necessary.

After some investigation, it appears we have to provide a default if the there isn't one available.

## Checklist

### DIFF
--- a/notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
+++ b/notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
@@ -1,0 +1,45 @@
+package org.opennms.horizon.notifications.grpc.config;
+
+import io.grpc.Context;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.context.spi.CurrentTenantIdentifierResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.orm.jpa.HibernatePropertiesCustomizer;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+import java.util.Optional;
+
+@Component
+class TenantIdentifierResolver implements CurrentTenantIdentifierResolver, HibernatePropertiesCustomizer {
+    private static final Logger LOG = LoggerFactory.getLogger(TenantIdentifierResolver.class);
+    private static final String DEFAULT_TENANT_ID = "opennms-prime";
+
+    @Autowired
+    TenantLookup tenantLookup;
+
+    @Override
+    public String resolveCurrentTenantIdentifier() {
+        Optional<String> tenantId = tenantLookup.lookupTenantId(Context.current());
+        if (tenantId.isPresent()) {
+            return tenantId.get();
+        } else {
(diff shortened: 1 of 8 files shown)


notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
```diff
@@ -0,0 +1,42 @@
...
+    TenantLookup tenantLookup;
+
+    @Override
+    public String resolveCurrentTenantIdentifier() {
+        Optional<String> tenantId = tenantLookup.lookupTenantId(Context.current());
+        if (tenantId.isPresent()) {
+            return tenantId.get();
+        } else {
+            LOG.warn("No tenant ID present");
+            return "Bootstrap";
```
thomasbigger584, line 29: should this be the default tenant id, or does this Bootstrap string mean something specific ?

james8hut: Apparently that should be the default tenant id. I thought perhaps we should be throwing an exception, but that results in spring initialization errors.

notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
```diff
@@ -0,0 +1,45 @@
+package org.opennms.horizon.notifications.grpc.config;
```
cgorantla, line 1: nit: add copyright headers, this should help you to make it easier
https://github.com/opennms/ide-setup

notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public String resolveCurrentTenantIdentifier() {
+        Optional<String> tenantId = tenantLookup.lookupTenantId(Context.current());
+        if (tenantId.isPresent()) {
+            return tenantId.get();
+        } else {
+            // Much as I think we should possibly throw an exception here, the internet says we should provide a default value.
+            // Attempting to throw an exception results in a failure during spring initialization.
+            LOG.warn("No tenant ID present");
+            return DEFAULT_TENANT_ID;
```
cgorantla, line 32: probably should use `DEFAULT_TENANT_ID` from GrpcConstants.

cgorantla (commented):
LGTM!  
couple of comments.

notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
james8hut: done

notifications/src/main/java/org/opennms/horizon/notifications/grpc/config/TenantIdentifierResolver.java
james8hut: done

PR #688, opened 2023-02-02
HS-963: Use consistent id on ignite client

cgorantla:
## Description
Use consistent id on ignite client

## Jira link(s)
- https://issues.opennms.org/browse/HS-963

## Checklist

### DIFF
--- a/minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/ignite/WorkerIgniteConfiguration.java
+++ b/minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/ignite/WorkerIgniteConfiguration.java
@@ -36,13 +36,11 @@ public class WorkerIgniteConfiguration {
 
         igniteConfiguration.setClientMode(false);
         igniteConfiguration.setMetricsLogFrequency(0);  // DISABLE IGNITE METRICS
-        String minionId = System.getenv("MINION_ID");
-        Optional.ofNullable(minionId).ifPresentOrElse(igniteConfiguration::setConsistentId, () -> {
-            String host = System.getenv("HOSTNAME");
-            if (host != null) {
-                igniteConfiguration.setConsistentId(host);
-            }
-        });
+
+        Optional.ofNullable(System.getenv("MINION_ID"))
+            .or(() -> Optional.ofNullable(System.getenv("HOSTNAME")))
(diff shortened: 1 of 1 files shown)


minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/ignite/WorkerIgniteConfiguration.java
```diff
@@ -34,7 +36,13 @@ public IgniteConfiguration prepareIgniteConfiguration() {
 
         igniteConfiguration.setClientMode(false);
         igniteConfiguration.setMetricsLogFrequency(0);  // DISABLE IGNITE METRICS
-
+        String minionId = System.getenv("MINION_ID");
+        Optional.ofNullable(minionId).ifPresentOrElse(igniteConfiguration::setConsistentId, () -> {
```
splatch, line 40:
```suggestion
        Optional.ofNullable(System.getenv("MINION_ID"))
            .or(() -> Optional.ofNullable(System.getenv("HOSTNAME")))
                .ifPresent(igniteConfiguration::setConsistentId);
```

splatch (approved):
:+1: It reflects current minion id assignment which is `${env:MINION_ID:-${env:HOSTNAME:-UNKNOWN}}` without last part which is OK because we then want to revert to ignite default.

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/ignite/WorkerIgniteConfiguration.java
```diff
@@ -35,6 +37,10 @@ public IgniteConfiguration prepareIgniteConfiguration() {
         igniteConfiguration.setClientMode(false);
         igniteConfiguration.setMetricsLogFrequency(0);  // DISABLE IGNITE METRICS
 
+        Optional.ofNullable(System.getenv("MINION_ID"))
+            .or(() -> Optional.ofNullable(System.getenv("HOSTNAME")))
+            .ifPresent(igniteConfiguration::setConsistentId);
+
```
artnaseef, line 43: Are we setting the same ID as the minion "system id"?  If so, can we inject the value from the same source as that value to ensure they are always the same in case the logic there changes?

If not, a comment here explaining the setting would be good since it is not obvious.

PR #698, opened 2023-02-06
HS-837: Common metric normalization

thomasbigger584:
## Description
HS-837: Common metric normalization

## Jira link(s)
- https://issues.opennms.org/browse/HS-837

## Checklist


inventory/src/main/resources/db/changelog/hs-0.1.0/tables/node-scan-type.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <changeSet id="0.1.0-node-scan-type" author="tbigg">
+        <addColumn tableName="node">
+            <column name="scan_type" type="text" />
```
james8hut, line 9: Is there a default value for rows that already exist in this table? Is null ok? What happens if null? (I can't see where this is used, so I'm guessing this PR is just for adding the extra column into the db?)

james8hut (commented):
Just some questions.

inventory/src/main/resources/db/changelog/hs-0.1.0/tables/node-scan-type.xml
thomasbigger584: good point, will have to add a default value

rest-server/src/main/java/org/opennms/horizon/server/service/metrics/normalization/Constants.java
```diff
@@ -0,0 +1,51 @@
...
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class Constants {
+    public static final String AZURE_MONITOR_TYPE = "AZURE";
+    public static final String SNMP_MONITOR_TYPE = "SNMP";
+
+    public static final String NODE_SCAN_TYPE = "NODE_SCAN";
+    public static final String AZURE_SCAN_TYPE = "NODE_SCAN";
```
thomasbigger584, line 40: happened during last minute refactoring - this should be AZURE_SCAN

shared-lib/inventory/src/main/proto/nodeService.proto
```diff
@@ -49,6 +49,7 @@ message NodeDTO {
   string system_location = 10;
   string system_contact = 11;
   repeated opennms.inventory.SnmpInterfaceDTO snmpInterfaces = 12;
+  string scan_type = 13;
```
yangli2014, line 52: The scan_type is not a node property, it sounds like an action type performed on the node. How about the node_type and the value can be "regular, azure, aws ... "?

thomasbigger584: node type was considered but it already means something on the UI. from what jerry and gerald was saying it can be either discoveryType or scanType and they preferred scanType

shared-lib/inventory/src/main/proto/nodeService.proto
```diff
@@ -49,6 +49,7 @@ message NodeDTO {
...
 }
 
 message NodeCreateDTO {
```
yangli2014, line 55: Also it is better include the nodeType in this NodeCreateDTO

thomasbigger584:
After hvaning a call with @geraldhumphries @cgorantla @splatch - the design of how we store nodes may likely change. This PR will remain open until a decision is made on what to do

thomasbigger584:
With a redesign of the node table (https://opennms.atlassian.net/browse/HS-1046) we will eventually be removing the scan type in favor of the inheritance approach.

thomasbigger584:
Azure node:
image

cgorantla (approved):
LGTM!

thomasbigger584:
rebase due to minor conflict, so merging

PR #699, opened 2023-02-06
HS-938: Add Default SNMP config and persist it

BenjaminJ:
## Description
Add default Snmp Config and persist this in config table
## Jira link(s)
- https://issues.opennms.org/browse/HS-938

## Checklist


cgorantla (commented):
There is lot of code that uses model object `org.opennms.horizon.shared.snmp.conf.xml.SnmpConfig` and related model objects in the same package in SnmpPeerFactory. It would be good to re-use these model objects.

cgorantla (commented):
I'm thinking we will have a default resource file  snmp-config.json and load this with  the same model objects from `org.opennms.horizon.shared.snmp.conf.xml.SnmpConfig`.  Update them as needed for marshal/unmarshalling from json

PR #705, opened 2023-02-07
HS-987 Fixed null pointer exception in node scan process

yangli2014:
## Jira link(s)
- https://issues.opennms.org/browse/HS-987

## Checklist

### DIFF
--- a/minion/plugins/nodescan-plugin/src/main/java/org/opennms/horizon/minion/nodescan/SNMPInterfaceTableTracker.java
+++ b/minion/plugins/nodescan-plugin/src/main/java/org/opennms/horizon/minion/nodescan/SNMPInterfaceTableTracker.java
@@ -208,7 +208,7 @@ public class SNMPInterfaceTableTracker extends TableTracker {
         }
 
         public SnmpInterfaceResult createInterfaceFromRow() {
-            return SnmpInterfaceResult.newBuilder()
+            SnmpInterfaceResult.Builder builder = SnmpInterfaceResult.newBuilder()
                 .setIfIndex(getIfIndex())
                 .setIfAdminStatus(getIfAdminStatus())
                 .setIfAlias(getIfAlias())
@@ -216,9 +216,11 @@ public class SNMPInterfaceTableTracker extends TableTracker {
                 .setIfName(getIfName())
                 .setIfOperatorStatus(getIfOperStatus())
                 .setIfSpeed(getIfSpeed())
-                .setIfType(getIfType())
-                .setPhysicalAddr(getPhysAddr())
-                .build();
+                .setIfType(getIfType());
+            if(getPhysAddr() != null) {
+                builder.setPhysicalAddr(getPhysAddr());
+            }
+            return builder.build();
         }
 
     }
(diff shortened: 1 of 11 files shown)


minion/plugins/nodescan-plugin/src/main/java/org/opennms/horizon/minion/nodescan/SNMPInterfaceTableTracker.java
```diff
@@ -208,17 +208,19 @@ private String getPhysAddr() {
...
                 .setIfAlias(getIfAlias())
                 .setIfDescr(getIfDescr())
                 .setIfName(getIfName())
                 .setIfOperatorStatus(getIfOperStatus())
                 .setIfSpeed(getIfSpeed())
-                .setIfType(getIfType())
-                .setPhysicalAddr(getPhysAddr())
-                .build();
+                .setIfType(getIfType());
+            if(getPhysAddr() != null) {
```
cgorantla, line 220: nit: formatting

minion/plugins/nodescan-plugin/src/main/java/org/opennms/horizon/minion/nodescan/SNMPInterfaceTableTracker.java
```diff
@@ -208,17 +208,19 @@ private String getPhysAddr() {
         }
 
         public SnmpInterfaceResult createInterfaceFromRow() {
-            return SnmpInterfaceResult.newBuilder()
+            SnmpInterfaceResult.Builder builder = SnmpInterfaceResult.newBuilder()
                 .setIfIndex(getIfIndex())
                 .setIfAdminStatus(getIfAdminStatus())
```
cgorantla, line 213: Intellij shows me all of these could be null
`getIfAdminStatus()`  `getIfOperStatus()`  `getIfSpeed()` , `getIfType()`.

We can handle these as optional

minion/plugins/nodescan-plugin/src/main/java/org/opennms/horizon/minion/nodescan/SNMPInterfaceTableTracker.java
```diff
@@ -208,17 +208,19 @@ private String getPhysAddr() {
...
                 .setIfAlias(getIfAlias())
                 .setIfDescr(getIfDescr())
```
cgorantla, line 215: let's also err on the side of caution. A device can give us null  for any of the fields.   Proto will throw NPE if you set null, we need to handle this for all of  String fields. 

We can consider using StringValue from https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/wrappers.proto   if we don't want  empty as default for Strings.

yangli2014: Sure, I will update all the trackers to handle potential null string values. I was thinking the StringValue, it seems will add more code everywhere to convert to String and back. I will try to use the mappers to prevent storing the empty string in DB.

cgorantla (approved):
LGTM!

PR #706, opened 2023-02-07
HS-954 Add pasted input box

chinh-cle:
## Jira link(s)
- https://issues.opennms.org/browse/HS-954-pasted-input-box

## Checklist


ui/package.json
```diff
@@ -94,6 +94,7 @@
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-vue": "^9.9.0",
     "happy-dom": "^8.1.5",
+    "ip-address-validator": "^1.0.9",
```
mikewrosey, line 97: We have the ip-regex package above, would it be possible to use that instead, to avoid having more than one validator? I think ip-regex is the definitive one as it has around 11mil weekly downloads. For examples, it's used in the useDiscoveryValidation file

chinh-cle: yes of course. I suspect that we already use one but forgot to ask you about it.

mikewrosey (approved):

PR #727, opened 2023-02-10
HS-952

artnaseef:
## Jira link(s)
- https://issues.opennms.org/browse/HS-952

## Checklist


artnaseef:
Note this is a draft now; there is significant effort required on the tests.  Note this version has been tested locally with the end-to-end external-it tests.

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/TaskSetServiceConfig.java
```diff
@@ -1,16 +1,29 @@
...
+
     @Bean
     public TaskSetPublisher taskSetService(GrpcTwinPublisher publisher) {
-        return new TaskSetPublisherImpl(publisher);
-    }
+        TaskSetPublisher result = new TaskSetPublisherImpl(publisher);
 
+        // Wire the publisher to listen for updates from the Task Set Storage
+        taskSetStorageListener = result::publishTaskSet;
+        taskSetStorage.addAllTwinPublisherSessionListener(taskSetStorageListener);
```
splatch, line 25: I was looking for actual use of `TaskSetStorageListener` and foud it only here in lambda form. Two questions arise - 1) is it worth replacing it with a full class, and 2) wouldn't it be better to mark `GrpcTwinPublisher` with all necessary interfaces directly?

artnaseef: There are merge conflicts that need to be resolved.  There's also significant changes to the testing, and some functionality that would be good to add to the cucumber tests.

However, at this point, the PR is in good shape to review and discuss the testing methodology as well as the functional changes.

artnaseef:
Squashed all of the commits in this PR to reduce the effort to resolve merge conflicts.  If anyone wants to see the individual commits, please let me know and I'll push up another branch with those commits.

.github/actions/jib-maven-sonar/action.yml
```diff
@@ -27,6 +27,34 @@ runs:
         path: ~/.sonar/cache
         key: ${{ runner.os }}-${{ runner.arch }}-sonar
 
+    - name: Clean up old artifacts from Maven Cache
+      run: |
```
j-white, line 31: Should we call `uses: ./.github/actions/maven-cache-cleanup-project-artifacts` here instead?

artnaseef: Is it possible to call an action from another action?

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/TaskSetServiceConfig.java
artnaseef: For `TaskSetStorageListener`, I think this pattern is alright - adding a full class is just boilerplate for the most part IMO.  Lambdas are a little less obvious code - do you think that's a problem in this case?  Since the `TaskSetStorageListener` interface is explicitly used to declare the `tasSetStorageListener` field, I think it's alright.

As for `GrpcTwinPublisher` - I think we're talking about the existing code structure with the abstract base class, right?  If so, I'm open to reworking it, but I'm going to leave that alone with this PR because this one has been delayed a long time now.

.github/actions/jib-maven-sonar/action.yml
artnaseef: I'm searching around for a solution to calling one action from another and finding it used to be impossible, but apparently it is now possible?   Seeing information suggesting it is possible, but the examples shown leave unanswered questions.

thomasbigger584 (approved):

.github/actions/jib-maven-sonar/action.yml
geraldhumphries: It is possible and actually pretty straightforward. You should be able to use this as the step:

```
      - name: Cleanup the Maven Cache
        uses: ./.github/actions/maven-cache-cleanup-project-artifacts
```

artnaseef: Excellent - trying that now

artnaseef: That worked - thank you!

artnaseef:
NOTE - I ran the external-it tests against the locally-running cluster, spun up with tilt.  The tests were successful.

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/internal/TaskSetIgniteStorageImpl.java
```diff
@@ -0,0 +1,144 @@
...
+                LOG.debug("Skipping task set update - returned task set is the original: tenant-id={}; location={}",
+                    tenantId, location);
+            }
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public void addAllTwinPublisherSessionListener(TaskSetStorageListener listener) {
```
splatch, line 99: While this method is clear when I look at it now, after going over code, I think naming this to `addSystemStorageListener` or something which reflects fundamental nature of that listener.

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/TaskSetPublisherImpl.java
```diff
@@ -29,18 +29,20 @@
...
 /**
  * Process task set updates, publishing them to downstream minions and storing the latest version to provide to minions
  *  on request.
+ *
+ *  This is the EGRESS part of task set management flow:
+ *      1. (INGRESS) updates received from other services, such as inventory
+ *      2. (STORE + AGGREGATE) task set updates made against the Task Set store
+ *      3. (EGRESS) On storage events, updates pushed downstream to Minions via Twin
  */
 public class TaskSetPublisherImpl implements TaskSetPublisher {
```
splatch, line 47: I'll keep my initial suggestion to make task set publisher implementation a storage listener. This way its clear how it is assembled. 

```suggestion
public class TaskSetPublisherImpl implements TaskSetPublisher, TaskSetStorageListener {
```

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/TaskSetPublisherImpl.java
```diff
@@ -67,10 +69,4 @@ public void publishTaskSet(String tenantId, String location, TaskSet taskSet) {
             throw new RuntimeException(e);
         }
     }
```
splatch, line 71:
```suggestion
    }

    // implementation of TaskSetStorageListener
    @Override
    public void publish(String tenantId, String location, TaskSet taskSet) throws IOException {
        publishTaskSet(tenantId, location, taskSet);
    }
```

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/TaskSetServiceConfig.java
splatch: I'd say making lambda in config class leads to situation makes unit testing more difficult to test. I do not see any other use of storage listener beyond pushing taskset update, and even if there will be more, triggering of taskset publishing is always going to go through `TaskSetPublisherImpl`. Making it explicitly turns it into self sufficient thing which can be unit tested with just one mock.

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/internal/TaskSetIgniteStorageImpl.java
```diff
@@ -0,0 +1,144 @@
...
+                } else {
+                    LOG.debug("Removing task set on update operation return null: tenant-id={}; location={}",
+                        tenantId, location);
+                    taskSetIgniteCache.remove(tenantKey);
+                }
+            } else {
+                LOG.debug("Skipping task set update - returned task set is the original: tenant-id={}; location={}",
+                    tenantId, location);
+            }
+        } finally {
```
splatch, line 93: Does it make sense to catch exceptions here?

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/internal/TaskSetIgniteStorageImpl.java
```diff
@@ -0,0 +1,144 @@
...
+        if (oldListener == null) {
+            taskSetIgniteCache.registerCacheEntryListener(listenerConfiguration);
+        } else {
+            // Why is the same TwinPublisher.Session being registered twice?
+            LOG.warn("Internal error - publisher session is already registered to receive cache events");
+        }
+    }
+
+    @Override
+    public void removeTwinPublisherListener(TaskSetStorageListener listener) {
```
splatch, line 128: I do not see this code being called anywhere. Given that storage listener is registered during bootstrap of context its nature is rather static. Not that I opt to remove this, but it seems to be semi dead branch of code.

j-white (approved):

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/internal/TaskSetTwinCacheListenerFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private final TaskSetStorageListener downstreamSession;
+
+    public TaskSetTwinCacheListenerFactory(TaskSetStorageListener downstreamSession) {
+        this.downstreamSession = downstreamSession;
+    }
+
+    @Override
+    public TaskSetTwinCacheListener create() {
+        LOG.debug("Creating listener for task set updates: tenant-id={}");
```
splatch, line 27: Log statement assumes presence of tenant-id which is missing in arguments.

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/internal/TaskSetToTwinCacheEntryEventFilterFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.cache.configuration.Factory;
+
+/**
+ *
+ */
+public class TaskSetToTwinCacheEntryEventFilterFactory implements Factory<TaskSetToTwinCacheEntryEventFilter> {
```
splatch, line 11: Is this code used?

minion-gateway/main/src/main/java/org/opennms/miniongateway/taskset/internal/TaskSetToTwinCacheEntryEventFilter.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.cache.event.CacheEntryEvent;
+import javax.cache.event.CacheEntryEventFilter;
+import javax.cache.event.CacheEntryListenerException;
+import java.util.Objects;
+
+/**
+ * Filter Ignite Cache Events for TaskSet updates to only those for the specific Tenant Location, identified by the
+ *  Tenant ID + Location.
+ */
+public class TaskSetToTwinCacheEntryEventFilter implements CacheEntryEventFilter<TenantKey, TaskSet> {
```
splatch, line 17: Is it being used?

splatch (approved):
I think this pull request is fine. There are few places where code is not obvious (at least going over github and intellij at the same time). Clearing this out would make it better to maintain.
Given amount of effort I'd rather get that PR in as soon as possible and make subsequent PR to address review/design comments if needed.

inventory/testtools/minion-gateway-wiremock/minion-gateway-wiremock-main/src/main/java/org/opennms/horizon/inventory/testtool/miniongateway/wiremock/rest/MinionGatewayWiremockRestController.java
```diff
@@ -0,0 +1,159 @@
...
+
+
+    @GetMapping(
+        path = "/taskset/updates",
+        produces = MimeTypeUtils.APPLICATION_JSON_VALUE
+    )
+    public Object getTaskSetUpdates() {
+
+        List<UpdateTasksRequest> records = mockGrpcServiceApi.getReceivedTaskSetUpdates();
+        List<Map<String, Object>> result = records.stream().map((record) -> convertProtobufToMap(record)).collect(Collectors.toList());
```
j-white, line 68: Much of the code in this class can be removed by a) defining a new type that is a list of `UpdateTasksRequest` and b) using `JsonFormat` to convert the proto directly to JSON

.github/actions/jib-maven-sonar/action.yml
```diff
@@ -38,8 +41,9 @@ runs:
         BASE_IMAGE=ghcr.io/opennms-cloud/$(cat ../.github/base-image-tag-ref.json | jq .image_tag_java | sed 's/\"//g')
 
         mvn -B -Dstyle.color=always \
-            -Pcoverage verify \
-            jib:build -Dapplication.docker.image=$IMAGE -Djib.from.image=$BASE_IMAGE \
+            -Pcoverage \
+            install \
```
geraldhumphries, line 45: Usage of `install` vs `verify` is something Art and I have already discussed enough, so I would like to get input from others. @artnaseef let me know if I miss or misrepresent anything. I also don't want to block this PR because of this, just open up the discussion.

The original goal was to remove the need to call the `jib:build` goal explicitly on the command line. The solution was to configure the jib plugin with an execution bound to a maven phase. Some microservices already have the `dockerBuild` goal bound to the `package` phase and minion-gateway has it bound to the `install` phase.

Here are the Maven related impacts of this PR:
- Removes the call to `jib:build` in favour of implicit plugin executions on each project.
- Changes the pipelines to use `install` instead of `verify`. This was done in the interest of "always use install" and to fix minion-gateway's build without `jib:build`.
- Caches the project's build artifacts alongside its dependencies due to the change from `verify` to `install`.
- Deletes old build artifacts from the runner's local file system after restoring the cache, to prevent issues that might happen because of stale artifacts.

To summarize what Art and I have been discussing, we want to pick between "use different phases depending on the context" and "always use install".

The pipelines currently run `verify`, which is where Maven looks at the results from integration tests and fails the build if a test failed. `install` is what developers are familiar with and use the most, kind of the defacto "default".  It's the phase that comes immediately after `verify` and copies build artifacts to Maven's `.m2` folder. That's the only real difference between running `mvn verify` and `mvn install` by default: with `verify`, only dependencies end up in the `.m2` folder. With install, build artifacts are copied in as well.

Our integration tests use the built Docker image directly. After removing `jib:build` from the pipeline's command, the build for minion-gateway failed during `integration-test` because it wasn't running `install` and never produced an image. 

Usually, binding Jib to `install` wouldn't work in the first place if the image is used for integration tests: the `integration-test` phase happens before `install`, so the image wouldn't be available to the tests. It works in minion-gateway because the image is built in a module that has no integration tests, which runs all the way to `install` before integration tests in other modules happen.

If we go with "always use install" and keep the execution of Jib bound to `install`, then `mvn verify`, `mvn integration-test` etc. won't work. We'd need to consider anything other than "install" as unsupported. We also need to deal with the presence of build artifacts in the .m2 folder/GHA cache and maintain a solution, as done in this PR with the `maven-cache-cleanup-project-artifacts` action.

I am looking for input from others who are familiar with Maven: do we always use `install` and consider other phases like `verify` unsupported, or do we fix those other phases and continue using them depending on the context?

inventory/testtools/minion-gateway-wiremock/minion-gateway-wiremock-main/src/main/java/org/opennms/horizon/inventory/testtool/miniongateway/wiremock/rest/MinionGatewayWiremockRestController.java
artnaseef: Agreed - I will follow-up after merging this update with a PR that uses that approach instead.

.github/actions/jib-maven-sonar/action.yml
splatch: I think that overall issue could be a bit deferred with "install at the end" https://maven.apache.org/plugins/maven-install-plugin/install-mojo.html#installAtEnd which makes "pollution" of local repository at the end of maven execution. Obviously it may have some implications for minion and karaf build which uses maven resolver, however it does address some of above concerns.
As of jib plugin binding - I'd say that given that jib lets us choice which build phase we want to use we can use anything from package and above, to have all resources needed to populate image. We keep manipulating plugin and their phases it a lot because in many places we have single module projects which forces us sorted plugins manually or by phases. As soon as docker based tests are moved away from main module, problem is, at least partially, gone.

PR #747, opened 2023-02-16
HS-951 Adding Adapter Step without Adapter module

FrancescoBuc86:
https://opennms.atlassian.net/browse/HS-951


minion/minion-flows/flows-parser/src/main/java/org/opennms/horizon/minion/flows/parser/TelemetryRegistryImpl.java
```diff
@@ -91,7 +102,7 @@ public FlowsListener getListener(ListenerConfig listenerConfig) {
         }
         for (var factory : listenerFactories) {
             if (factory.getClass().getName().contains(listenerConfig.getClassName())) {
```
freddy61025, line 104: if you use full class path in config. Then you may need factory.getListenerClass().getName().equals(listenerConfig.getClassName()). 

Also same for parser.

freddy61025 (changes requested):
there still have one test fail.

FrancescoBuc86:
> there still have one test fail.

Yes saw that, I'll take care of it

freddy61025 (approved):
LGTM

PR #749, opened 2023-02-16
HS-1013: Use Hibernate tenancy feature in inventory

james8hut:
## Description
Use Hibernate tenancy feature in inventory

## Jira link(s)
- https://issues.opennms.org/browse/HS-1013

## Flagged for review
Repo methods like findByTenantId are no longer needed, as hibernate takes care of all that stuff for us using the TenantIdentifierResolver. However they all still work even if they aren't removed.
If needed they can be removed in a future story.

## Checklist

### DIFF
--- a/inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
@@ -31,6 +31,7 @@ package org.opennms.horizon.inventory.component;
 import java.util.Map;
 import java.util.Optional;
 
+import io.grpc.Context;
 import org.opennms.horizon.grpc.heartbeat.contract.HeartbeatMessage;
 import org.opennms.horizon.inventory.exception.InventoryRuntimeException;
 import org.opennms.horizon.inventory.service.MonitoringSystemService;
@@ -58,8 +59,11 @@ public class MinionHeartbeatConsumer {
             String tenantId = Optional.ofNullable(headers.get(GrpcConstants.TENANT_ID_KEY)).map(o -> new String((byte[])o))
                 .orElseThrow(()->new InventoryRuntimeException("Missing tenant id"));
             log.info("Received heartbeat message for minion with tenant id: {}; id: {}; location: {}", tenantId, message.getIdentity().getSystemId(), message.getIdentity().getLocation());
-            Optional<Long> newCreatedId = service.addMonitoringSystemFromHeartbeat(message, tenantId);
-            newCreatedId.ifPresent(minionRpc::addSystem);
+            Context.current().withValue(GrpcConstants.TENANT_ID_CONTEXT_KEY, tenantId).run(()->
+            {
+                Optional<Long> newCreatedId = service.addMonitoringSystemFromHeartbeat(message, tenantId);
+                newCreatedId.ifPresent(minionRpc::addSystem);
+            });
         } catch (Exception e) {
             log.error("Error while processing heartbeat message: ", e);
         }
--- a/inventory/src/main/java/org/opennms/horizon/inventory/grpc/TenantIdentifierResolver.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/grpc/TenantIdentifierResolver.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.horizon.inventory.grpc;
+
+import io.grpc.Context;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.context.spi.CurrentTenantIdentifierResolver;
+import org.opennms.horizon.shared.constants.GrpcConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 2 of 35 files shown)


inventory/src/main/java/org/opennms/horizon/inventory/service/MonitoringLocationService.java
```diff
@@ -39,10 +42,15 @@ public List<MonitoringLocationDTO> findByLocationIds(List<Long> ids) {
     }
 
     public List<MonitoringLocationDTO> findAll() {
-        List<MonitoringLocation> all = modelRepo.findAll();
-        return all
-            .stream()
-            .map(mapper::modelToDTO)
-            .collect(Collectors.toList());
+        List<MonitoringLocationDTO> all = jdbcTemplate.query(
+            "Select id, tenant_id, location from monitoring_location",
```
geraldhumphries, line 46: Why change to a native query? (I assume the builder is used instead of the mapper because you're not using the entity)

inventory/src/test/java/org/opennms/horizon/inventory/cucumber/CucumberRunnerIT.java
```diff
@@ -139,6 +139,7 @@ private static void startApplicationContainer() {
             .withEnv("INVENTORY_AZURE_LOGIN_URL", wiremockUrl)
             .withEnv("INVENTORY_AZURE_MANAGEMENT_URL", wiremockUrl)
             .withEnv("INVENTORY_ENCRYPTION_KEY", RandomStringUtils.randomAlphanumeric(32))
+            .withEnv("SPRING_DATASOURCE_DRIVER-CLASS-NAME", "org.postgresql.Driver")
```
geraldhumphries, line 142: Should this be what's used everywhere? It could be part of the default properties in `application.yaml`.

splatch: I think another aspect is that `DRIVER-CLASS-NAME` might not work as it is invalid env variable.

inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
```diff
@@ -58,8 +59,11 @@ public void receiveMessage(@Payload byte[] data, @Headers Map<String, Object> he
             String tenantId = Optional.ofNullable(headers.get(GrpcConstants.TENANT_ID_KEY)).map(o -> new String((byte[])o))
                 .orElseThrow(()->new InventoryRuntimeException("Missing tenant id"));
             log.info("Received heartbeat message for minion with tenant id: {}; id: {}; location: {}", tenantId, message.getIdentity().getSystemId(), message.getIdentity().getLocation());
-            Optional<Long> newCreatedId = service.addMonitoringSystemFromHeartbeat(message, tenantId);
-            newCreatedId.ifPresent(minionRpc::addSystem);
+            Context.current().withValue(GrpcConstants.TENANT_ID_CONTEXT_KEY, tenantId).run(()->
```
splatch, line 62: Use of GRPC `Context` to propagate tenancy context across layers below GRPC itself is wrong as it ties a strictly speaking database and its abstraction layer code (`TenantIdentifierResolver`) to RPC mechanism. While we are in semi flat classpath it has other side effects such as use of GRPC specifics in Kafka consumers which have nothing to do with GRPC itself.
Main motivation beside `TenantLookup` was possibility to introduce different kinds of contexts while retaining same operation contract. I think it would be better if another context kind would be chosen if we go with implicit way. The GRPC `Context` is just glorified wrapper for `ThreadLocal`.

inventory/src/main/java/org/opennms/horizon/inventory/grpc/TenantIdentifierResolver.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.orm.jpa.HibernatePropertiesCustomizer;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+import java.util.Optional;
+
+@Component
+class TenantIdentifierResolver implements CurrentTenantIdentifierResolver, HibernatePropertiesCustomizer {
```
splatch, line 45: This type belongs rather to Model as it is closely tied with Hibernate related SPI.

inventory/src/main/java/org/opennms/horizon/inventory/model/TenantAwareEntity.java
```diff
@@ -0,0 +1,48 @@
...
+ * <p>Abstract Entity class.</p>
+ */
+@Getter
+@Setter
+@MappedSuperclass
+public abstract class TenantAwareEntity {
+
+    @TenantId
+    @Column
+    private String tenantId;
```
splatch, line 47: Is it possible to make use of composite keys with `@TenantId`? I ask cause currently we have a situation where primary key is composed only from sequence number. This means that all foreign keys managed via hibernate **do not validate `tenantId`** across entities. It is possible to provoke situation in which `MonixtoringSystem` of tenant X will be assigned to `MonitoringLocation` of tenant Y and database checks will be perfectly fine with it.

inventory/src/main/java/org/opennms/horizon/inventory/service/MonitoringLocationService.java
james8hut: This is only called from TrapConfigService.sendTrapConfigToMinionAfterStartup, which in turn is called on the ApplicationReadyEvent (on start up). Here we need all locations across all tenants. If we use a Repository method, we only get locations for the current tenant in context. If there's an alternate/better way, please advise.

inventory/src/main/java/org/opennms/horizon/inventory/model/TenantAwareEntity.java
james8hut: Will investigate in another story: https://opennms.atlassian.net/browse/HS-1002

inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
james8hut: Will create another story to look at this.

inventory/src/main/java/org/opennms/horizon/inventory/grpc/TenantIdentifierResolver.java
james8hut: Will move this in the other story to look at moving away from GrpcContext

inventory/src/test/java/org/opennms/horizon/inventory/cucumber/CucumberRunnerIT.java
james8hut: Have removed this line. Not needed.

inventory/src/main/java/org/opennms/horizon/inventory/service/MonitoringLocationService.java
james8hut: (Because I didn't use the JPA repo, I thought I shouldn't use the JPA Entity either, so I used the DTO builder to go directly to the DTO.)

inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
mszoek: Why another story? Isn't that introducing tech debt?

james8hut: The grpc Context has already been used in other micro services. Need another story to take care of all those services. So the tech debt is already there.

PR #754, opened 2023-02-16
HS-1022 Syslog form logic validation and save

chinh-cle:
## Description
- form validation
- save form
- display toaster (success/fail)
- display user guide modal (success)

## Jira link(s)
- https://issues.opennms.org/browse/HS-1022-syslog-form-logic

## Flagged for review

image

## Checklist


ui/src/components/Discovery/DiscoverySyslogSNMPTrapsForm.vue
```diff
@@ -54,47 +56,35 @@
       </div>
     </form>
   </div>
+  <DiscoverySuccessModal ref="successModalRef" />
```
mikewrosey, line 59: We have this rendered in the parent `Discovery.vue` and could avoid duplication if we pass the success function down into this component from the parent. Could do it with a prop like the azure form...

ui/src/components/Discovery/DiscoverySyslogSNMPTrapsForm.vue
```diff
@@ -118,22 +109,31 @@ const udpPort = {
...
-  console.log('args', args)
+let UDPPortEntered = ''
+const UDPPortEnteredListener = (str: string) => {
+  UDPPortEntered = str
 }
 
+const successModalRef = ref()
+const isSyslogSNMPTrapsFormSaveSuccess = computed(() => discoveryQueries.isSyslogSNMPTrapsFormSaveSuccess)
+watch(isSyslogSNMPTrapsFormSaveSuccess, (success: boolean) => {
+  if (success) successModalRef.value.openSuccessModal()
```
mikewrosey, line 125: Not sure if this is just for testing, but I think we can remove the whole `isSyslogSNMPTrapsFormSaveSuccess` from here and the store. Below we call `saveSyslogSNMPTrapsForm` that can be an async function that makes the call then returns true/falls for success.

ui/src/components/Discovery/DiscoverySyslogSNMPTrapsForm.vue
```diff
@@ -118,22 +109,31 @@ const udpPort = {
...
-  // if success
-  // stopSpinner()
-  // if error
-  // showSnackbar({
-  // msg: 'Save unsuccessfully!'
-  // })
+  contentEditableCommunityStringRef.value.validateAndFormat()
+  contentEditableUDPPortRef.value.validateAndFormat()
+
+  discoveryQueries.saveSyslogSNMPTrapsForm({
```
mikewrosey, line 131: To add details for above comment - this line would turn to `const success = await discoveryQueries.saveSyslogSNMPTrapsForm`

and then something like this follows
```
if (success) {
  openSuccessModal()
}
```

ui/src/store/Queries/discoveryQueries.ts
```diff
@@ -21,28 +45,61 @@ export const useDiscoveryQueries = defineStore('discoveryQueries', () => {
...
-        {
-          id: 1,
-          name: 'local',
-          tenantId: 'opennms-prime'
-        },
+      tagsUponTyping.value = success
+    }, 1000)
+  }
+
+  const saveSyslogSNMPTrapsForm = (values: unknown) => {
```
mikewrosey, line 66: and this would also turn into an `async` function with a return value of success true or false

mikewrosey:
looks like merging origin into here now has java files in the PR. I think merging 955 in here should fix it

ui/src/components/Discovery/DiscoveryContentEditable.vue
```diff
@@ -53,34 +53,42 @@ const props = defineProps({
...
+const htmlString = ref(props.defaultContent)
 
 const contentChange = () => {
   isContentNotEmpty.value = contentEditableRef.value.textContent.length as boolean
 }
 
 const validateAndFormat = () => {
+  // console.log('validateAndFormat')
   isContentInvalid.value = validateContent()
-  htmlString.value = formatContent()
```
anyarybalova, line 70: I guess isContentInvalid is used only in this function, no need to be a ref

mikewrosey (approved):

PR #770, opened 2023-02-21
HS-1025: Add Discovery ( PingSweep) support

cgorantla:
## Description
Add PingSweep Support as part of Discovery Scan

## Jira link(s)
- https://issues.opennms.org/browse/HS-1025
- https://opennms.atlassian.net/browse/HS-1024

## Checklist


minion/plugins/icmp-plugin/src/main/java/org/opennms/horizon/minion/icmp/PingScan.java
```diff
@@ -0,0 +1,217 @@
...
+        private final PingSweepResponse pingSweepResponse = new PingSweepResponse();
+
+        public void expectCallbackFor(InetAddress address) {
+            waitingFor.add(address);
+        }
+
+        @Override
+        public void handleResponse(InetAddress address, EchoPacket response) {
+            if (response != null) {
+                PingResult sweepResult = new PingResult(address, response.elapsedTime(TimeUnit.MILLISECONDS));
```
yangli2014, line 173: Can we use PingResponse directly here?

yangli2014 (commented):
Can you rebase develop branch into your branch? When creating a new node, we should add related tags (comments from the team).

minion/plugins/icmp-plugin/src/main/java/org/opennms/horizon/minion/icmp/PingScan.java
yangli2014: OK, but it is still being used in the minion plugin currently.

cgorantla: I know we still use this to parse request but it would be good to have separation. 
I don't want to force the implementation dependency on proto.

PR #790, opened 2023-02-28
HS-1103: Add Active Discovery Config table persistence

BenjaminJ:
- created table active-discovery-config
- created repository ActiveDiscoveryConfigRepository 
- created entity ActiveDiscoveryConfig

HS-1103

## Description
Currently, Active Discovery is being persisted in generic Config table.
But we need to move this to Active Discovery Config table that consists of 

1. tenant_id
2. location
3. profile name
4. IPAddresses (ranges/specific/cidr) 
5. SNMP community strings 
6. SNMP agent config ports. 

4-5-6 can be combined one DTO that can be saved as json.

This task involves creating **table**, **entity**, **repository**.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1103

## Checklist

### DIFF
--- a/inventory/src/main/java/org/opennms/horizon/inventory/model/ActiveDiscoveryConfig.java
+++ b/inventory/src/main/java/org/opennms/horizon/inventory/model/ActiveDiscoveryConfig.java
@@ -1,0 +1,65 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.horizon.inventory.model;
+
+import org.hibernate.annotations.JdbcTypeCode;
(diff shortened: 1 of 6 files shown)


inventory/src/main/java/org/opennms/horizon/inventory/model/ActiveDiscoveryConfig.java
```diff
@@ -0,0 +1,65 @@
...
+    private String location;
+
+    @NotNull
+    @Column(name = "profile_name")
+    private String profileName;
+
+    @NotNull
+    @Column(name = "discovery_config", columnDefinition = "jsonb")
+    @JdbcTypeCode( SqlTypes.JSON )
+    private DiscoveryConfig discoveryConfig;
```
thomasbigger584, line 64: was it not discussed that we should go inheritance approach for Active Discovery? we will also have azure active discovery which will be a subtype ?

inventory/src/main/java/org/opennms/horizon/inventory/model/DiscoveryConfig.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class DiscoveryConfig {
+
+    private List<String> ipAddresses;
+
+    private String communityString;
```
cgorantla, line 42: This also can be a list

inventory/src/main/java/org/opennms/horizon/inventory/model/ActiveDiscoveryConfig.java
cgorantla: Not sure about this.  What would be the benefit of inheritance approach ?

thomasbigger584: ok was under the impression we wanted strongly typed tables and was agreed by gerald. can also do this also

thomasbigger584: will require azure credential to also be the same format

thomasbigger584: frontend needs an endpoint to get all active discovery whether its of this icmp/snmp type or azure type

cgorantla: we are using strongly typed instead of common Config table for all the configs.

A list can be persisted either as array/json in postgres.  we can further divide DiscoverConfig into multiple columns but i want to start with this as this may require changes down the line.

cgorantla: > frontend needs an endpoint to get all active discovery whether its of this icmp/snmp type or azure type

Not sure why that constraint. Even then we shouldn't enforce this at entity level and handle this at API level.

cgorantla: yeah, we can handle this at API level. Currently we don't have create time for this, may be it's needed. let me know if you have an issue for this specific scenario.

cgorantla (approved):

PR #793, opened 2023-03-01
HS-1112: Consolidate event model to a single set of protobuf objects.

j-white:
JIRA: https://opennms.atlassian.net/browse/HS-1112

I also opted to make the tenant id explicit in the objects when going through this.

Tested by sending a trap to a Minion with: `snmptrap -v 2c -c public localhost:1162 '' 1.3.6.1.4.1.8072.2.3.0.1 1.3.6.1.4.1.8072.2.3.2.1 i 123456`, verifying that the node gets created, and verifying that events continue to be populated in the UI:
image


events/consumer/src/main/java/org/opennms/horizon/events/consumer/EventsConsumer.java
```diff
@@ -28,29 +28,24 @@
...
 import org.springframework.messaging.handler.annotation.Payload;
 import org.springframework.stereotype.Component;
 
-import java.net.UnknownHostException;
 import java.time.LocalDateTime;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
j-white, line 48: oops, need to adjust my IDE settings to avoid these

cgorantla (approved):
LGTM.
Need to resolve conflicts though!

j-white:
Thanks, will fix and merge.

PR #802, opened 2023-03-02
HS-1128: Revert TenantID annotation from inventory

james8hut:
This reverts commit c23cdee86dc9ba82b1159edb61ed2d280c614fda.

## Description
Reverting as per request from Chandra

## Jira link(s)
- https://issues.opennms.org/browse/HS-1013

## Flagged for review
Reverted via git revert. Please double check that nothing obvious has been got wrong.

## Checklist


inventory/main/src/main/java/org/opennms/horizon/inventory/model/MonitoringLocation.java
```diff
@@ -43,11 +43,14 @@
...
 @RequiredArgsConstructor
 @Entity
-public class MonitoringLocation extends TenantAwareEntity {
+public class MonitoringLocation {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private long id;
 
+    @NotNull
+    private String tenantId;
```
thomasbigger584, line 52: for consistency, this one is missing @Column ?

inventory/main/src/test/java/org/opennms/horizon/inventory/grpc/DiscoveryConfigGrpcItTest.java
```diff
@@ -77,6 +76,7 @@ public void prepare() throws VerificationException {
 
     @AfterEach
     public void cleanUp() throws InterruptedException {
+        configRepo.deleteAll();
         afterTest();
```
thomasbigger584, line 80: it looks like the `afterTest` now already does a clean of database? should this be updated instead ?

james8hut: removed

inventory/main/src/main/java/org/opennms/horizon/inventory/model/MonitoringLocation.java
james8hut: added in

inventory/main/src/main/java/org/opennms/horizon/inventory/grpc/MonitoringLocationGrpcService.java
```diff
@@ -109,7 +109,7 @@ public void listLocationsByIds(IdList request, StreamObserver<MonitoringLocation
     @Override
     public void searchLocations(StringValue request, StreamObserver<MonitoringLocationList> responseObserver) {
         List<MonitoringLocationDTO> locations = tenantLookup.lookupTenantId(Context.current())
-            .map(tenantId -> service.searchLocations(request.getValue())).orElseThrow();
+            .map(tenantId -> service.searchLocationsByTenantId(request.getValue(), tenantId)).orElseThrow();
```
splatch, line 112: While I know it may not be entirely accurate comment, but given basic fact that we will always have tenantId in service calls, can we make it first argument for services be tenantId or use composite structure (`TenantedId`) where we lookup things by id? For above case it would be `searchTenantLocations(tenant,location)`. Making it so, especially on repositories will make it easier to spot eventual gaps in search terms

inventory/main/src/main/java/org/opennms/horizon/inventory/model/ActiveDiscoveryConfig.java
```diff
@@ -45,11 +45,15 @@
 @Setter
 @RequiredArgsConstructor
 @Entity(name = "active_discovery_config")
-public class ActiveDiscoveryConfig extends TenantAwareEntity {
```
splatch, line 48: I think we could retain here `TenantAwareEntity` if we make it our `@MappedSuperclass`.

inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
```diff
@@ -0,0 +1,67 @@
+/*******************************************************************************
```
splatch, line 1: Classes in `inventory/src/...` were moved to `inventory/main/src/...`.

inventory/main/src/main/java/org/opennms/horizon/inventory/grpc/MonitoringLocationGrpcService.java
geraldhumphries: This PR is for reverting a different PR, so it might be best to do refactors as a separate PR.

james8hut: Agree with Gerald

inventory/main/src/main/java/org/opennms/horizon/inventory/model/ActiveDiscoveryConfig.java
james8hut: The TenantAwareEntity subclass was added for TenantId annotation. Given that we're rolling all this back, I think it should still go for now.

inventory/src/main/java/org/opennms/horizon/inventory/component/MinionHeartbeatConsumer.java
james8hut: Will remove

PR #830, opened 2023-03-09
HS-1170 Get updated location list in case of removal minion/node

chinh-cle:
## Jira link(s)
- https://issues.opennms.org/browse/HS-1170

## Checklist


ui/src/store/Common/locationsQueries.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { data, execute, isFetching, error } = useQuery({
+    query: ListLocationsDocument,
+    fetchOnMount: false,
+    cachePolicy: 'network-only'
+  })
+
+  watchEffect(() => {
+    if (!error.value) {
+      if (!isFetching.value) {
+        locations.value = data.value?.findAllLocations as Location[]
```
mikewrosey, line 17: to avoid type casting, maybe just `locations.value = data.value?.findAllLocations || []`
and then below `locations: computed(() => locations.value)`

ui/src/store/Common/locationsQueries.ts
```diff
@@ -0,0 +1,28 @@
...
+      }
+    } else {
+      // TODO: what kind of errors and how to handle them
+    }
```
mikewrosey, line 21: it should be possible to eventually handle all errors through the central gql service, so we probably don't need this

PR #848, opened 2023-03-13
HS-1199 - Add Filters to Flows page.

rcorry93:
## Description
This is a PR to add the needed filters to the Flows page. These are placeholders until the endpoints are ready.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1199

## Flagged for review
I have updated the BasicSelect to have chips for the selected value. Prop is set to false.
To view page, goto '/flows'.

## Checklist

### DIFF
--- a/ui/src/components/Common/BasicSelect.vue
+++ b/ui/src/components/Common/BasicSelect.vue
@@ -1,93 +1,33 @@
 <template>
-  <div>
-    <FeatherSelect
-      :style="{ width: size + 'px' }"
-      label=""
-      :options="list"
-      text-prop="name"
-      v-model="selectedItem"
-      hideLabel
-      :disabled="isDisabled"
-      @update:modelValue="setSelectedItem"
-    >
-    </FeatherSelect>
-    <FeatherChipList
-      v-if="selectedItem && selectedItem.id && props.showChip"
-      label="Item chip list with icon"
-      data-test="fds-select-chip"
-      class="basic-select-chip-list">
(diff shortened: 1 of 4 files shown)


ui/src/components/Common/BasicSelect.vue
```diff
@@ -1,36 +1,93 @@
...
+      <FeatherChip>
+          <span>{{ selectedItem.name }}</span>
+          <template v-slot:icon>
+            <FeatherIcon
+              @click="unselectItem()"
+              :icon="CancelIcon"
+        /></template>
+    </FeatherChip>
+  </FeatherChipList>
+</div>
```
mikewrosey, line 28: This basic select component was originally designed for monitoring policies, and the designs do not have the chip. I'm not 100% sure if monitoring policies will continue using this or switch to the standard feather select but perhaps for now it would be better to create BasicChipSelect.vue or something, which we can use going forward for all places that need it. We'll leave this comp as is until we know for sure how it will be used.

rcorry93: Your changes have been made. Should the BasicSelect component be in the common components if it was designed for monitoring policies? Thats the reason i added a chip prop as i thought the component could be used everywhere.

mikewrosey: It could be used in other places if we keep it. Probably it will be removed altogether though. Work on MP is this sprint.

mikewrosey (approved):

PR #858, opened 2023-03-14
HS-1204: Add aggregator POM at root of the repo

geraldhumphries:
## Description
- Adds an aggregator POM that defines top-level projects as modules
- IntelliJ's "Cannot reconnect" Maven issue no longer appears to happen
- Initial project import is much more smooth
- Changing branches works well, no more issues caused by new/removed modules
- Cleanup of unnecessary version elements in pom files

## Jira link(s)
- https://issues.opennms.org/browse/HS-1204

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <name>OpenNMS Horizon Stream</name>
+    <description>
+        Aggregator POM for Horizon Stream.
+
+        This POM file is only to enable better IDE support and should not be used to instrument builds. Top-level projects should be built individually.
+
+        `maven-enforcer-plugin` is used to discourage its use.
+    </description>
+
+    <groupId>org.opennms.horizon</groupId>
+    <artifactId>horizon-stream</artifactId>
+    <version>0.0.0</version>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>alert</module>
+        <module>datachoices</module>
+        <module>events</module>
+        <module>external-it</module>
+        <module>inventory</module>
+        <module>metrics-processor</module>
+        <module>minion</module>
+        <module>minion-gateway</module>
+        <module>minion-gateway-grpc-proxy</module>
+        <module>notifications</module>
+        <module>parent-pom</module>
+        <module>rest-server</module>
+        <module>shared-lib</module>
+        <module>tooling/ignite-tool</module>
+    </modules>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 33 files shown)


thomasbigger584:
Im a frequent user of `./build-tools/basic/compile.sh` for compiling all the services without running. Does this need updated if we no longer have the parent-pom module ?

geraldhumphries:
> Im a frequent user of `./build-tools/basic/compile.sh` for compiling all the services without running. Does this need updated if we no longer have the parent-pom module ?

Good catch, I've fixed that script.

thomasbigger584 (approved):

pom.xml
artnaseef, line 1: This is solving a bottom-up problem (tool not working, so change the project structure).  Have we thought about the top-down implications?  In this case, they are huge.

I'm very concerned that this change moves us towards a monolith, and lock-step updates across services.

geraldhumphries: Can you elaborate? This doesn't change anything about how we version and release things, and from my knowledge none of the builds for top-level projects are affected. It also won't stop us from doing something like versioning and releasing the parent pom itself, which would allow different top-level projects to depend on different versions of it.

There are two changes here that affect project structure:

- Moved the location of the parent pom from the `parent-pom` folder to the root of the repo.
- Add a modules section to the parent pom.

The first change is fairly inconsequential. The builds of top-level projects work as they did before, moving the file and configuring the child poms with the new path doesn't really change how any of it works. Since the project now follows the standard Maven layout, configuring the top-level poms properly just meant removing `relativePath` from the `parent` elements as our project structure now matches the default of `..`.

The second change has a bigger impact, turning the parent pom into an aggregator.

It does not have any effect on the builds of the top-level projects like `mvn ... -f alert` as the parent isn't part of the reactor build, so its modules section isn't used. Here is what it does affect:

- We can run Maven commands like `mvn clean install` from the root of the repo. All of the listed modules/top-level projects will be part of the reactor build, and it will successfully build everything in the right order. This isn't used anywhere, but It could be a more convenient way to work with the project outside of Tilt than what we have now.
- When we want to build the parent pom alone we need to tell that to Maven. The `-N` flag is used in the parent-pom workflow to make the build non-recursive. This could also be done with `mvn ... -pl horizon-parent`.
- IntelliJ hooks into this to manage its own module system. Its behaviour was really poor before this change, but now it properly sets up your workspace for you and even appears to index faster. Switching between branches that add or remove modules is significantly less clunky. It seems to fix the problem in the Jira issue, switching branches doesn't kill IntelliJ's interaction with Maven.

I suspect the root cause of the problem in the Jira issue is an IntelliJ bug that needs to be fixed by them. Given the above, I didn't find any downsides to using the standard structure for our project.

Side note: If we want to version and release shared-lib, we also need to version and release the parent pom. When shared-lib is depended on and downloaded/installed, the parent pom also needs to be installed or maven won't be able to find the variables it provides. This has always been true and this PR doesn't change it. I tested out `flatten-maven-plugin` to resolve the parent pom's variables into shared-lib's pom file and it worked fine, it got rid of the need to explicitly install it in the pipelines before shared-lib was installed.

If we decide that it (or any of the other projects) needs a specific version of the parent pom then this doesn't really matter, we would need to version and release the parent pom anyways.

geraldhumphries: @artnaseef can you please elaborate? @fooker any thoughts?

geraldhumphries:
Closing this, the IntelliJ issue seems to have been fixed in 2023.1.

PR #862, opened 2023-03-15
[HS-1220] Use tilt-helm-values in run-cluster script

geraldhumphries:
## Description

- Updates the run-cluster script to use the main tilt-helm-values file
- Should help with maintainability, these files were identical and sometimes there were problems when they weren't kept in sync.
- Cleans up the old values file, and an unused custom resource file

## Jira link(s)
- https://issues.opennms.org/browse/HS-1220

## Checklist


splatch (approved):
I think that both value files are now pretty close to each other, thus I do not see issues with removal of one under build-tools.

geraldhumphries:
@splatch the alternative development workflow presents a maintenance challenge whenever we do anything beyond the basics with Tilt. When I first opened this PR the files were identical, but the addition of secret creation to Tilt for the certificates caused some big differences. This probably breaks the workflow of the `build-tools/basic` scripts.

splatch:
@geraldhumphries I think we can gather all 4 calls of these scripts into one file and just refer it once in tiltfile, install-local and run-cluster.sh stuff. After all, each of these calls will have to stay in sync across all places. From logical flow it is 1) create cluster & namespace 2) inject secrets 3) call helm to bring deployments. WDYT?

geraldhumphries:
Idea is sound but the Tiltfile can't do step 1 and 3, it needs to only do 2 - I think that's easy enough.

build-tools/basic/run-cluster.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-helm upgrade -i opennms ./charts/opennms -f build-tools/basic/helm-values.yaml \
+helm upgrade -i opennms ./charts/opennms -f tilt-helm-values.yaml \
```
splatch, line 3:
```suggestion
./install-local/load-or-generate-secret.sh "default" "root-ca-certificate"  "opennms-ca" "target/tmp/server-ca.key" "target/tmp/server-ca.crt"
./install-local/generate-and-sign-certificate.sh "default" "opennms-minion-gateway-certificate" "minion.onmshs.local" "target/tmp/server-ca.key" "target/tmp/server-ca.crt"
./install-local/generate-and-sign-certificate.sh "default" "opennms-ui-certificate" "onmshs.local" "target/tmp/server-ca.key" "target/tmp/server-ca.crt"
./install-local/load-or-generate-secret.sh "default" "client-root-ca-certificate" "client-ca" "target/tmp/client-ca.key" "target/tmp/client-ca.crt"
helm upgrade -i opennms ./charts/opennms -f tilt-helm-values.yaml \

```

PR #864, opened 2023-03-15
HS-1196: Create table chart component

rcorry93:
## Description
Created a basic chart component. Used that to create the TableChart component needed for a future story.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1196

## Flagged for review
Potential improvements for BasicChart.
Exported flows type to index as file could not be resolved.

## Checklist

image


ui/src/types/index.ts
```diff
@@ -1,6 +1,7 @@
 import { SORT } from '@featherds/table'
 import { PointerAlignment, PopoverPlacement } from '@featherds/popover'
 
+export * from './flows.d'
```
chinh-cle, line 4: it's good to know this works!

chinh-cle (approved):
LGTM

ui/src/types/index.ts
rcorry93: I believe the issue might be the '.d.ts'. The path can't be resolved when it has the 'd' as my error would say it can't find "@src/types/flows" instead of "@src/types/flows.d". Just a guess.

ui/src/components/Flows/TableChart.vue
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+})
+
+//Potentially handled by BE
+const addValues = (a: number, b: number) => {
+  const total = (a + b).toString()
+  return (parseFloat(total).toPrecision(3))
+}
+const postFixGB = (value: any) => value.toString() + ' GB'
```
mikewrosey, line 139: is value always a number here? 
might be able to remove `.toString()` as a `number + ' GB'` will convert to a string anyway

mikewrosey (approved):

ui/src/components/Flows/TableChart.vue
rcorry93: Updated

PR #875, opened 2023-03-17
Feature/hs 1212 openshift routing

jkapponms:
## Description
Openshift Routes and Documentation

## Jira link(s)
- https://issues.opennms.org/browse/HS-1212

### DIFF
--- a/charts/opennms/templates/ingress/openshift/onms-api-route.yaml
+++ b/charts/opennms/templates/ingress/openshift/onms-api-route.yaml
@@ -5,7 +5,7 @@ metadata:
   name: opennms-api-route
   namespace: {{ .Release.Namespace }}
 spec:
-  subdomain: {{ .Values.Host }}
+  host: {{ .Values.Host }}.<Ingress_Domain>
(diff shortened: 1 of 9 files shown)


charts/opennms/templates/ingress/openshift/onms-api-route.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: opennms-api-route
   namespace: {{ .Release.Namespace }}
 spec:
-  subdomain: {{ .Values.Host }}
+  host: {{ .Values.Host }}.apps-crc.testing
```
geraldhumphries, line 8: As someone who isn't familiar with OpenShift, what is `.apps-crc.testing`? It looks like a development option?

jkapponms: It's the default DNS that comes installed with a local OpenShift cluster. This being hardcoded is a workaround to get a local install of ONMS working. Kicking the can down the road to make this more dynamic for SaaS in the future.

jaymberry (approved):

PR #880, opened 2023-03-18
HS-1258: Use TenantedTaskSetResults  in Minion Gateway.  

artnaseef:
Mapper between tenanted and bare added with test ensuring the two structure are in-sync.

## Description
WORK IN PROGRESS.  Posting the PR early to share the incoming pattern.
NOTE the first iteration adds TenantedTaskSetResults and a mapper, and changes the Minion Gateway to push it onto the Kafka topic consumed by Metrics Processor.  Metrics Processor will BREAK with the current state of the code because Tenant ID is no longer set on the Kafka header, and it has not YET been updated to use TenantedTaskSetResults.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1258

## Flagged for review
- tenantedTaskSetProto in shared-lib/protobuf/src/main/proto
- TenantedTaskSetResultMapperImpl in shared-lib/protobuf sources.
- TenantedTaskSetResultMapperImplTest in shared-lib/protobuf sources.  Notice the verification of field parity between TenantedTaskSetResults and TaskSetResults.

## Checklist

### DIFF
--- a/inventory/docker-it/src/test/java/org/opennms/horizon/inventory/cucumber/steps/AzureDiscoveryStepDefinitions.java
+++ b/inventory/docker-it/src/test/java/org/opennms/horizon/inventory/cucumber/steps/AzureDiscoveryStepDefinitions.java
@@ -50,15 +50,14 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 
 public class AzureDiscoveryStepDefinitions {
-    private static InventoryBackgroundHelper backgroundHelper;
+    private final InventoryBackgroundHelper backgroundHelper;
     private AzureActiveDiscoveryCreateDTO createDiscoveryDto;
     private AzureActiveDiscoveryDTO discoveryDto;
     private TagCreateDTO tagCreateDto1;
     private TagListDTO tagList;
 
-    @BeforeAll
-    public static void beforeAll() {
-        backgroundHelper = new InventoryBackgroundHelper();
+    public AzureDiscoveryStepDefinitions(InventoryBackgroundHelper backgroundHelper) {
+        this.backgroundHelper = backgroundHelper;
     }
 
     /*
--- a/shared-lib/protobuf/src/main/proto/tenatedTaskSet.proto
+++ b/shared-lib/protobuf/src/main/proto/tenatedTaskSet.proto
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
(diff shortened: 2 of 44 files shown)


artnaseef:
Looks like an update is imminent - having a test failure on the latest at the moment.

shared-lib/protobuf/src/main/proto/tenatedTaskSet.proto
```diff
@@ -0,0 +1,45 @@
...
+
+import "google/protobuf/any.proto";
+import "taskSet.proto";
+
+package opennms.taskset;
+option java_multiple_files = true;
+option java_package = "org.opennms.taskset.contract";
+
+message TenantedTaskSetResults {
+  string tenantId = 100;
```
j-white, line 42: Curious why we chose to create a wrapper, vs just adding the `tenantId` field to the existing `TaskSetResults` message.

j-white (commented):
I like the idea 👍

shared-lib/protobuf/src/main/proto/tenatedTaskSet.proto
artnaseef: When we say wrapper, I think of composition - so `TenantedTaskSetResults` would consist of fields `tenantId` and a `TaskSetResult`.  We chose to use fully-hydrated structures while having the ones sent from Minion to Cloud never contain Tenant ID to avoid:

- the related confusion (when someone looking at Minion code sees this field and asks "where do I get this value?") 
- and possibility of the cloud incorrectly trusting the value sent by the Minion

If we used a composition-based wrapper, then the mapper goes away and the need to ensure the two data structures are in-sync goes away; and the need for a mapper goes away.  Of course, if protobuf supported inheritence, we could also eliminate the impact of keeping the definitions in-sync, but there is still a mapping to do between them.

artnaseef: Am I the only one that reads this and thinks, "didn't the industry solve basic data-structure handling yet"? ;-)

splatch (approved):
Small suggestion I have is removal of header arguments from kafka consumers (i.e. `MinionHeartbeatConsumer#receiveMessage`).
Other point - the GRPC contracts are expected to be just contracts with proto files, so any java code/interfaces placed in shared-lib might soon become a problem.

inventory/docker-it/src/test/java/org/opennms/horizon/inventory/cucumber/steps/AzureDiscoveryStepDefinitions.java
```diff
@@ -50,15 +50,14 @@
...
+    private final InventoryBackgroundHelper backgroundHelper;
     private AzureActiveDiscoveryCreateDTO createDiscoveryDto;
     private AzureActiveDiscoveryDTO discoveryDto;
     private TagCreateDTO tagCreateDto1;
     private TagListDTO tagList;
 
-    @BeforeAll
-    public static void beforeAll() {
-        backgroundHelper = new InventoryBackgroundHelper();
+    public AzureDiscoveryStepDefinitions(InventoryBackgroundHelper backgroundHelper) {
```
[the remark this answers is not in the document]
artnaseef: Cucumber does it - it manages all of the instances of Test Set Definition classes during the test.  The `cucumber-picocontainer` dependency adds the support for this type of injection.

PR #886, opened 2023-03-20
HS-1201: Add expanding panel with charts to flows page

rcorry93:
## Description
Added component to expand and hide charts with dialog filter options

## Jira link(s)
- https://issues.opennms.org/browse/HS-1201

## Flagged for review
New component and store methods.

## Checklist


ui/src/components/Flows/ExpandingChartWrapper.vue
```diff
@@ -0,0 +1,81 @@
...
+    :model-value="modelValue"
+  >
+    <template #title>
+      <div class="expanding-chart-title-wrapper">
+        <div class="title">{{ props.title }}</div>
+        <div class="divider">|</div>
+        <div class="filter-button">
+          <FeatherButton
+            icon="Filter"
+            @click="(e) => onFilterClick(e)"
```
mikewrosey, line 13: should be able to do `@click.stop.prevent="onFilterClick"` and you won't have to pass the event up to the store just to stop propagation / prevent default.

ui/src/components/Flows/TableChart.vue
```diff
@@ -21,25 +20,25 @@
           </tr>
         </thead>
         <tbody>
-          <tr v-for="(data, index) in tableData" :key="index">       
+          <tr
+            v-for="(data, index) in tableData"
+            :key="index"
```
mikewrosey, line 25: small thing but does the key in this particular case matter? seems like the default "in-place patch" strategy works here

ui/src/containers/Flows.vue
```diff
@@ -31,23 +32,99 @@
...
 <script setup lang="ts">
 import { useFlowsStore } from '@/store/Views/flowsStore'
 
 const flowsStore = useFlowsStore()
+const appDialogLabels = ref({
+  title: 'Top Ten Applications (24 Hrs) - Total'
+})
+const expDialogLabels = ref({
+  title: 'Top Ten Exporters (24 Hrs) - Total'
+})
```
mikewrosey, line 127: do these titles need to be reactive? looks like it would work without wrapping them in refs

rcorry93: Updated now.

ui/src/components/Flows/TableChart.vue
rcorry93: Only reason i added it is because i get an ESLint error, may be related to my linting setup.
image

ui/src/components/Flows/ExpandingChartWrapper.vue
rcorry93: Good spot, forgot about that feature!

PR #890, opened 2023-03-21
Feature/hs 1183 tags

anyarybalova:
## Description
Discovery page => get tags by discovery Id and show saved tags when discovery is selected

## Jira link(s)
- https://issues.opennms.org/browse/HS-1183

## Checklist


ui/src/components/Common/BasicAutocomplete.vue
```diff
@@ -178,4 +189,8 @@ defineExpose({
     order: 1;
   }
 }
+
+:deep(.feather-input-sub-text) {
```
rcorry93, line 193: I was considering on doing this too, but i was worried that this is the div used to display errors. Unsure if this component is designed for errors anyway. Just something to think about in the future.

rcorry93 (approved):
LGTM, just left a small comment on error handling but i don't think it matters to the component.

ui/src/components/Common/BasicAutocomplete.vue
anyarybalova: this is on optional field, for this form, but yeah this component is more generic, so I would have to remove it...

ui/src/graphql/Discovery/tagsByDiscoveryIdQuery.gql
```diff
@@ -0,0 +1,15 @@
...
+  }
+}
+
+query TagsByPassiveDiscoveryId($discoveryId: Long!) {
+  tagsByPassiveDiscoveryId(passiveDiscoveryId: $discoveryId) {
+    id
+    name
+    tenantId
+  }
+}
```
mikewrosey, line 15: As these queries are strictly for the discovery page, I probably wouldn't create a new file but just add them to the existing discoveryQueries.graphql file

ui/src/components/Discovery/DiscoveryAzureForm.vue
```diff
@@ -101,6 +101,14 @@ const props = defineProps<{
   discovery: AzureActiveDiscovery | null
 }>()
 
+const tags = computed(() => (props.discovery?.id ? discoveryQueries.tagsByActiveDiscoveryId : []))
+
+onMounted(() => {
+  if (props.discovery?.id) {
+    discoveryQueries.getTagsByActiveDiscoveryId(props.discovery?.id)
```
mikewrosey, line 108: small thing but can rm the question mark on this line, as we are making sure it exists with the conditional. Same for the two watchers below

ui/src/components/Discovery/DiscoverySnmpForm.vue
```diff
@@ -123,7 +133,7 @@ const setSnmpConfig = (property: string, val: (string | number)[] | null) => {
 
 const tagsAutocompleteRef = ref()
 const tagsSelectedListener = (tags: Record<string, string>[]) => {
-  discoveryInfo.value.tags = tags.map(({ name }) => ({ name }))
+  ;(discoveryInfo.value as IcmpActiveDiscoveryCreateInput).tags = tags.map(({ name }) => ({ name }))
```
mikewrosey, line 136: was there a previous console.log for testing above this line? otherwise, i don't think this first semi-colon is needed..  not giving me any errors without it

ui/src/store/Queries/discoveryQueries.ts
```diff
@@ -7,13 +7,18 @@ import {
   ListDiscoveriesDocument,
   ActiveDiscovery,
   IcmpActiveDiscovery,
-  AzureActiveDiscovery
+  AzureActiveDiscovery,
+  TagsByActiveDiscoveryIdDocument,
+  TagsByPassiveDiscoveryIdDocument
```
mikewrosey, line 12: Probably related to earlier changes and not this PR but looks like `IcmpActiveDiscovery`, `AzureActiveDiscovery ` and `DiscoveryType` imports can all be removed

ui/src/components/Common/BasicAutocomplete.vue
```diff
@@ -97,19 +102,27 @@ const selectedItems = ref<IAutocomplete[]>([])
...
     _text: item.name
   }))
 })
 
+onMounted(() => {
+  props.getItems()
+})
+
+watch(props, () => {
+  if (props.preselectedItems?.length > 0) {
```
mikewrosey, line 116: getting an issue where if I click a discovery that has tags and then click a discovery that has no tags, the tags from the first one will appear on the second. Might be this conditional here, we still need to set the tags even if it's an empty array

ui/src/store/Queries/discoveryQueries.ts
```diff
@@ -41,19 +46,47 @@ export const useDiscoveryQueries = defineStore('discoveryQueries', () => {
...
+  // active discoveries
+  const { data: tagsByDiscoveryIdData, execute: tagsByDiscoveryIdExecute } = useQuery({
+    query: TagsByActiveDiscoveryIdDocument,
+    variables: discoveryId
+  })
+
+  const getTagsByActiveDiscoveryId = async (id: number) => {
+    discoveryId.value.discoveryId = id
+    await tagsByDiscoveryIdExecute()
+    tagsByDiscovery.value = tagsByDiscoveryIdData.value?.tagsByActiveDiscoveryId || []
```
mikewrosey, line 64: One thing that seems to work is removing these two lines so that we just have:
`const getTagsByActiveDiscoveryId = async (id: number) => discoveryId.value.discoveryId = id`
which should automatically trigger the call and update the data. 
Below we would have:
`tagsByActiveDiscoveryId: computed(() => tagsByDiscoveryIdData.value?.tagsByActiveDiscoveryId || [])`

I tested it and it seems to work, but I could be missing some edge case. If it does work, the same should be applied to the passive query.

mikewrosey (approved):

PR #900, opened 2023-03-22
Discovery test  HS-1280

chuklov:
## Description
Simple discovery test that check that we have discovery working and creating a node that has been discovered. 

## Jira link(s)
- https://issues.opennms.org/browse/HS-1280

### DIFF
--- a/external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/helper/TestsExecutionHelper.java
+++ b/external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/helper/TestsExecutionHelper.java
@@ -1,0 +1,142 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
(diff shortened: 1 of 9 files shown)


external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/helper/TestsExecutionHelper.java
```diff
@@ -0,0 +1,142 @@
...
+        LOG.info("checkTheStatusOfTheNode");
+
+        try {
+            URL url = formatIngressUrl("/api/graphql");
+            String accessToken = getUserAccessTokenSupplier().get();
+
+            return executePost(url, accessToken, gqlQuery);
+
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
artnaseef, line 137: Recommend using `LOG.error("...", e)` here instead of `e.printStackTrace()`

external-it/external-horizon-stream-it/src/main/resources/org/opennms/horizon/it/horizon-stream.feature
```diff
@@ -40,3 +40,8 @@ Feature: Minion Monitoring via Echo Messages Logged in Prometheus
     Then Add a device with label "NodeDown" IP address "192.168.0.4" and location "Default"
     Then Check the status of the Node with expected status "DOWN"
     Then Delete the first node from inventory
+
+  Scenario: Create discovery and check the status of the discovered node
+    Then Add a new active discovery for the name "Automation Discovery Tests" at location "Default" with ip address "127.1.0.5" and port 161, readCommunities "public"
```
artnaseef, line 45: Is something listening on port 161 at `127.1.0.5`?  Also, as we discussed earlier today, it's best to avoid using fixed port numbers if feasible.

For the end-to-end tests, that's not as critical as it is for other tests, but still valuable - as long as the effort involved is not too large.

external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/helper/TestsExecutionHelper.java
chuklov: Corrected

artnaseef (approved):

PR #903, opened 2023-03-23
HS-1205 Add detailed comment about FlowDocument Fields

FrancescoBuc86:
https://opennms.atlassian.net/browse/HS-1205


shared-lib/flows/src/main/proto/flow_document.proto
```diff
@@ -73,33 +79,49 @@ message FlowDocument {
...
+    // Calculated during enrichment phase from 'location', 'dstAddress', 'tenantId' fields
     NodeInfo src_node = 35;
+    // Calculated during enrichment phase from 'location', 'exporterAddress', 'tenantId' fields
     NodeInfo exporter_node = 36;
+    // Calculated during enrichment phase from 'location', 'srcAddress', 'tenantId' fields
     NodeInfo dest_node = 37;
+    // Calculated during enrichment phase
     string application = 38;
     string host = 39;
     string location = 40;
```
freddy61025, line 102: I think this is minion location.

shared-lib/flows/src/main/proto/flow_document.proto
```diff
@@ -73,33 +79,49 @@ message FlowDocument {
...
     google.protobuf.UInt32Value vlan = 34; // VLAN ID.
+    // Calculated during enrichment phase from 'location', 'dstAddress', 'tenantId' fields
     NodeInfo src_node = 35;
+    // Calculated during enrichment phase from 'location', 'exporterAddress', 'tenantId' fields
     NodeInfo exporter_node = 36;
+    // Calculated during enrichment phase from 'location', 'srcAddress', 'tenantId' fields
     NodeInfo dest_node = 37;
+    // Calculated during enrichment phase
     string application = 38;
     string host = 39;
```
freddy61025, line 101: i have did a quick search. It seems it never been use.

FrancescoBuc86: but it's used in the ClassificationRequest to set the ExporterAddress

freddy61025 (approved):

PR #905, opened 2023-03-23
HS-1284: Add JdbcInitComponent for Ignite

artnaseef:
Add JdbcInitComponent that waits - up to a timeout - for the database to be accessible, and make the ignite bean depend on it.

This works-around the problem that Ignite only tries one time to get connected to the database for cache JDBC storage, thereby making the service not resilient to slowness, or temporary DB outage, during startup.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1284
## Checklist

### DIFF
--- a/minion-gateway/main/src/main/java/org/opennms/miniongateway/jdbc/JdbcInitComponent.java
+++ b/minion-gateway/main/src/main/java/org/opennms/miniongateway/jdbc/JdbcInitComponent.java
@@ -1,0 +1,159 @@
+/*
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *
+ */
+
+package org.opennms.miniongateway.jdbc;
+
+import lombok.Setter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.PostConstruct;
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.SQLException;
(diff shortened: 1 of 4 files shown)


splatch:
Rather a generic question, could we shift that functionality to connection pool validation query?

Looking at unified options I come across https://github.com/spring-projects/spring-boot/issues/2183 which means that we'd need to decide on use of specific connection pool (tomcat/dbcp/hikari) which supports that.

cgorantla:
> Rather a generic question, could we shift that functionality to connection pool validation query?
> 
> Looking at unified options I come across [spring-projects/spring-boot#2183](https://github.com/spring-projects/spring-boot/issues/2183) which means that we'd need to decide on use of specific connection pool (tomcat/dbcp/hikari) which supports that.

I think we should use standard connection pool for this. Classic uses hikari connection pool.

artnaseef:
> > Rather a generic question, could we shift that functionality to connection pool validation query?
> > Looking at unified options I come across [spring-projects/spring-boot#2183](https://github.com/spring-projects/spring-boot/issues/2183) which means that we'd need to decide on use of specific connection pool (tomcat/dbcp/hikari) which supports that.
> 
> I think we should use standard connection pool for this. Classic uses hikari connection pool.



> > Rather a generic question, could we shift that functionality to connection pool validation query?
> > Looking at unified options I come across [spring-projects/spring-boot#2183](https://github.com/spring-projects/spring-boot/issues/2183) which means that we'd need to decide on use of specific connection pool (tomcat/dbcp/hikari) which supports that.
> 
> I think we should use standard connection pool for this. Classic uses hikari connection pool.

Will the connection pool work the same way at startup - that is, blocking until a connection can be made and aborting service startup after a timeout?

I believe we still need this explicit bean dependency for the ignite bean to get the entire service to fail on startup.  If we just add a connection pool and use that with Ignite, it's startup-failure logic will still apply - meaning that repeated failures to connect to the database will leave the service in an unusable state.

artnaseef:
Let me assert - this is working now.  Let's get it in and stop the immediate problem, and then improvements can come in a follow-on ticket.

Also, off the top of my head, I believe the Spring Boot datasource setup does use a connection pool by default.

splatch:
> I believe we still need this explicit bean dependency for the ignite bean to get the entire service to fail on startup. If we just add a connection pool and use that with Ignite, it's startup-failure logic will still apply - meaning that repeated failures to connect to the database will leave the service in an unusable state.

To my knowledge validation query should be executed before pool passes connection to application for further use. In this regard validation should block use of connection. The intermediate result would be failure to initialize application in general.

artnaseef:
> > I believe we still need this explicit bean dependency for the ignite bean to get the entire service to fail on startup. If we just add a connection pool and use that with Ignite, it's startup-failure logic will still apply - meaning that repeated failures to connect to the database will leave the service in an unusable state.
> 
> To my knowledge validation query should be executed before pool passes connection to application for further use. In this regard validation should block use of connection. The intermediate result would be failure to initialize application in general.

So this will take a little more effort to investigate and prove out.  So again, I request that we merge this PR and then if the extra effort is warranted, let's create a ticket for that.

This is causing a lot of trouble, and this fix is reasonable, even if an alternative exists and is better in many ways.

minion-gateway/main/src/main/java/org/opennms/miniongateway/jdbc/JdbcInitComponent.java
```diff
@@ -0,0 +1,159 @@
...
+            LOG.error("Timed out attempting to connect to the database; aborting startup: connection-attempt-count={}; timeout={}", count, timeout);
+            throw new RuntimeException("Timed out attempting to connect to the database; aborting startup", lastException);
+        }
+    }
+
+//========================================
+// Internals
+//----------------------------------------
+
+    private boolean attemptConnect() {
```
cgorantla, line 124: Instead of designing our own retry logic, can we depend on `RetryTemplate` from Spring, an example here in our code: [RetryTemplate](https://github.com/OpenNMS-Cloud/horizon-stream/pull/910/files)

cgorantla (approved):

PR #906, opened 2023-03-23
HS-1082 Support for propagation of metric collection time.

splatch:
## Description
Propagation of metric collection timestamp from minion to opennms backend.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1082

## Flagged for review
I am not clear on timezone handling. Minions will span across different zones hence I think we should go with UTC times, but then there is a question of how cortex/prometheus is interpreting timestamp attached to metric.

Finally there is a question of out-of-order samples received with a significant delay.

## Checklist


BenjaminJ (approved):

splatch:
> Should we extend this to all metrics including Collector ?

I think so, as issue with time offsets and delays in processing are same in both cases.

artnaseef (approved):
As discussed separately - I believe it is not possible to preserve timestamps when Prometheus is scraping metrics from the application.  However, for Cortex - where we push the samples - I believe the timestamp is used.

There is one nit comment, but otherwise LGTM.

PR #922, opened 2023-03-28
HS-1348 Fixed the creating policy error due to the protobuf default enum values

yangli2014:
## Jira link(s)
- https://issues.opennms.org/browse/HS-1348

## Checklist

### DIFF
--- a/shared-lib/alert/src/main/proto/monitoringPolicy.proto
+++ b/shared-lib/alert/src/main/proto/monitoringPolicy.proto
@@ -34,6 +34,17 @@ package opennms.inventory;
 option java_multiple_files = true;
 option java_package = "org.opennms.horizon.shared.alert.policy";
 
+message CreateMonitorPolicyRequest {
+  string name = 1;
+  string memo = 2;
+  repeated CreatePolicyRuleRequest rules = 3;
+  repeated string tags = 4;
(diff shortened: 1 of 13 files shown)


shared-lib/alert/src/main/proto/monitoringPolicy.proto
```diff
@@ -34,25 +34,42 @@ package opennms.inventory;
 option java_multiple_files = true;
 option java_package = "org.opennms.horizon.shared.alert.policy";
 
+message CreateMonitorPolicyRequest {
```
j-white, line 37: Curious why introduce new objects instead of just using the `MonitorPolicyProto`?

Seems like we're just omitting `id` and `tenantId`, but the fields are optional so they can be ignored.

BenD10 (commented):
No comments from me other than what Jesse said.

shared-lib/alert/src/main/proto/monitoringPolicy.proto
yangli2014: That is true. Just saw this article https://medium.com/@akhaku/protobuf-definition-best-practices-87f281576f31 to separate the request and response. It seems not very useful for our case.

BenD10 (approved):
LGTM! 🎸

PR #923, opened 2023-03-29
HS-1329: GRPC logging of connection attempts and failures.  Uses a wo…

artnaseef:
…rk-around to get logging with SLF4J at INFO level instead of JUL logging at TRACE (FINER / FINEST) level.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1329

## Checklist

### DIFF
--- a/alert/pom.xml
+++ b/alert/pom.xml
@@ -50,8 +50,9 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>io.grpc</groupId>
-            <artifactId>grpc-core</artifactId>
+            <groupId>org.opennms.horizon.shared</groupId>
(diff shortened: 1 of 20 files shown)


artnaseef:
The key part of this PR for review - the shaded grpc-core artifact, horizon-shaded-grpc-core, which replaces an internal logging class with our own.

One of the big reasons for this approach - to get the logging out of GRPC core at reasonable levels, rather than the default TRACE level we are getting now.  While turning the one logger up to TRACE level would be acceptable since that's not generating a lot of messages, the biggest issue that opens is that most console appenders limit to INFO level and above, and getting these TRACE level messages onto the console may not be feasible.

splatch (approved):

j-white (approved):
Nice workaround 👍

PR #933, opened 2023-03-29
Adding more logging to tests

chuklov:
## Description
Adding more logging to testing

## Jira link(s)
- https://issues.opennms.org/browse/HS-1280

## Flagged for review
Do not approve this, it is only for testing

### DIFF
--- a/external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/InventoryTestSteps.java
+++ b/external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/InventoryTestSteps.java
@@ -157,7 +157,7 @@ public class InventoryTestSteps {
             Awaitility
                 .await()
                 .ignoreExceptions()
-                .atMost(30, TimeUnit.SECONDS)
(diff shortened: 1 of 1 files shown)


external-it/external-horizon-stream-it/src/main/java/org/opennms/horizon/it/InventoryTestSteps.java
```diff
@@ -157,7 +157,7 @@ public void checkTheStatusOfTheFirstNodeTest(String status) {
             Awaitility
                 .await()
                 .ignoreExceptions()
-                .atMost(30, TimeUnit.SECONDS)
+                .atMost(120, TimeUnit.SECONDS)
```
artnaseef, line 160: I think I mentioned this in the past - it would be nice to have the Cucumber pass in the timeout, so that the timeouts are explicit when reading the cucumber steps themselves.

Going to approve this PR in spite of it though - it's not too important at this time.

artnaseef (approved):

PR #943, opened 2023-04-03
HS-1140 Allow matching multiple monitoring policies for alerts

BenD10:
## Description
Alerts may have multiple monitoring policies associated with them, so the notification service needs to honor that.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1140

## Checklist

### DIFF
--- a/notifications/src/main/java/org/opennms/horizon/notifications/service/NotificationService.java
+++ b/notifications/src/main/java/org/opennms/horizon/notifications/service/NotificationService.java
@@ -39,6 +39,7 @@ import org.opennms.horizon.notifications.tenant.WithTenant;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.util.List;
 import java.util.Optional;
 
 @Service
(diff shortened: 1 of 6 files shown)


vserrago (commented):
Looks good to me, but I'll defer to one of the horizon stream maintainers.

yangli2014 (approved):
LGTM

notifications/src/main/java/org/opennms/horizon/notifications/service/NotificationService.java
```diff
@@ -53,12 +54,12 @@ public class NotificationService {
...
     public void postNotification(Alert alert) {
-        Optional<MonitoringPolicy> dbPolicy = monitoringPolicyRepository.findByTenantIdAndId(
+        List<MonitoringPolicy> dbPolicies = monitoringPolicyRepository.findByTenantIdAndIdIn(
             alert.getTenantId(),
-            alert.getMonitoringPolicyId()
+            alert.getMonitoringPolicyIdList()
         );
 
-        dbPolicy.ifPresentOrElse(policy -> {
+        dbPolicies.forEach(policy -> {
```
james8hut, line 62: Question on this: If we have multiple monitoring policies which each have the notifyByPagerDuty flag set, then we'll get multiple calls to PagerDuty for the same alert. Is this really what we want here?

BenD10: That's a good point, maybe it has to be more a deduplication, if PD is enabled in at least one policy, send it, if email is enabled in atleast one of the policies, send it, etc.

james8hut: As I'm just back from vacation and at a loose end, I'll create a story for this and do it. :-)

PR #951, opened 2023-04-04
HS-859: operator will ping the UI as a health check on the instance

jkapponms:
## Description
Operator will send a request to the UI service, and will mark the instance as ready and healthy if the UI responds with a 200 OK. 

## Jira link(s)
- https://issues.opennms.org/browse/HS-859


operator/internal/reconciler/health_check_test.go
```diff
@@ -0,0 +1,102 @@
...
+			},
+			Release: values.HelmRelease{
+				Namespace: namespace,
+			},
+		},
+	}
+	return instance
+}
+
+func Test_InstanceReady_200OK(t *testing.T) {
```
kbakanov, line 69: Looks good, but please add precondition and expected result, e.g. Test_InstanceReady_LoginAvailable_ReturnsTrue

kbakanov: The same goes for other ones too

kbakanov: Btw, probably best to use lower case "i", i.e. instanceReady instead of InstanceReady.

jkapponms:
I'll work on splitting the test methods up in other parts of the code base as I return to them.

kbakanov (approved):

PR #956, opened 2023-04-04
HS-1311: opennms-minion-ssl-gateway running nginx to terminate SSL fr…

artnaseef:
…om Minion for dev.  Tools to create certs.  Insecure dev certs.  Minion can connect with SSL.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1311

## Checklist


artnaseef:
Still working on wrapping this up.

There is a good amount of content in this PR already, and there may be questions or opinions around how it works.  Looking for thoughts / feedback.

Here are some of the questions coming up:

- The ZIP file handling currently is based on PEM files only; should we add explicit support for a PKCS12 file to it?
- The private key format can be either PEM or PKCS12; the config flag `grpc.client.private.key.is-pkcs12` is used to select; do we want to support more formats?  Only support PKCS12? ...
- The bootstrap process that extracts contents from the ZIP ignores entries with unknown names; do we want to fail the process in that case?  Do we want to fail if expected contents are missing?
- Do we want to more closely link the ZIP file to the Minion's SSL certificate configuration?  Currently the configuration for the bootstrap process is separate from the SSL certificate configuration, but with some overlap (see `org.opennms.minion.bootstrap.cfg` and `org.opennms.core.ipc.grpc.client.cfg`)
- In some environments the Minion may need to trust a self-signed CA (because the cloud SSL server certs are signed by that CA), but in others the server cert will be signed by a public CA; therefore, do we keep the support for including the CA certificate in the ZIP file dropped into the Minion, or do we find an alternative way to get it in there?
- End-to-End tests - do we want to add SSL here to these tests as well?

jaymberry (approved):

kbakanov:
> Still working on wrapping this up.
> 
> There is a good amount of content in this PR already, and there may be questions or opinions around how it works. Looking for thoughts / feedback.
> 
> Here are some of the questions coming up:
> 
> * The ZIP file handling currently is based on PEM files only; should we add explicit support for a PKCS12 file to it?
> * The private key format can be either PEM or PKCS12; the config flag `grpc.client.private.key.is-pkcs12` is used to select; do we want to support more formats?  Only support PKCS12? ...
> * The bootstrap process that extracts contents from the ZIP ignores entries with unknown names; do we want to fail the process in that case?  Do we want to fail if expected contents are missing?
> * Do we want to more closely link the ZIP file to the Minion's SSL certificate configuration?  Currently the configuration for the bootstrap process is separate from the SSL certificate configuration, but with some overlap (see `org.opennms.minion.bootstrap.cfg` and `org.opennms.core.ipc.grpc.client.cfg`)
> * In some environments the Minion may need to trust a self-signed CA (because the cloud SSL server certs are signed by that CA), but in others the server cert will be signed by a public CA; therefore, do we keep the support for including the CA certificate in the ZIP file dropped into the Minion, or do we find an alternative way to get it in there?
> * End-to-End tests - do we want to add SSL here to these tests as well?

@artnaseef we had a discussion today with @jbeuree @splatch and @BenjaminJ about PKCS12 vs ZIP. We decided to try going ahead with the PKCS12 format instead of ZIP. From my end, it seems to be doable on the PKI side but I will build a prototype before starting major work to make sure that it is in fact feasible.

artnaseef:
> @artnaseef we had a discussion today with @jbeuree @splatch and @BenjaminJ about PKCS12 vs ZIP. We decided to try going ahead with the PKCS12 format instead of ZIP. From my end, it seems to be doable on the PKI side but I will build a prototype before starting major work to make sure that it is in fact feasible.

@kbakanov thank you for the update.  @splatch caught me up on this topic.  We can make the necessary changes on the Minion end based on the result of that decision.  My current plan is to continue to finish up this PR in spite of the possible need to change the ZIP file handling - the code is currently 100% backward compatible, and whatever changes we choose, I believe they will be easy enough to implement.

splatch:
I agree that we're good to merge these changes, and we shall not queue PRs which are self-contained for later adoption. If we pkcs12 comes in we can amend code to new format.

splatch (approved):

PR #971, opened 2023-04-08
Add support for JSON logging for spring boot services

deejgregor:
**I highly suggest reviewing commit by commit.** Please also see the questions section further down. This is my first PR for horizon-stream, so I'll be extremeley happy with any guidance/etc. that can be provided.

Supporting infrastructure:
- [Add spring-boot-app-config ConfigMap that is mounted in /app/config](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/1f7fc420074419342ad36f03b2273853d3e149a8) 
- [Add spring-boot-env ConfigMap that adds env to all Spring Boot services](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/bd9572ad9caa41f51d6352615e912666669fd1b7)

The parts that actually matter:
- [HS-1419](https://opennms.atlassian.net/browse/HS-1419)[: Add spring-boot-logback-json to support JSON logging](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/a8b72b15a0aba3c22f845c702d418b10c4ec2fc0) 
- [HS-1419](https://opennms.atlassian.net/browse/HS-1419)[: Add spring-boot-logback-json to our spring-boot services](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/c02c22bfb031ecaa97931bbf49d2ec1c3072239d)

Cleanups I did along the way: 
- [This looks like it accidentally had application.docker.image.name set](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/b17482a9ae93874e255624e091a3c44fbec8f166) 
- [Ooops, looks like we were building events and tagging it as datachoices](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/89cb11ae8406f966213f87eadf1956f105fe35f3)
- [Clean up a System.out.println](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/0d4d59c9307edeac0578d2c3059665cbe5439ba0)
- [Add .helmignore files so vim's .swp files don't mess up Tilt/Helm](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/8a883fdffc28e5ea476e601315cd2267a67a2a7b)

A note on why `logback-spring.xml` is used instead of `logback.xml`: https://github.com/spring-projects/spring-boot/issues/15649

## Description

Support an option to enable JSON logging for Spring Boot-based services and enable it by default in our Helm chart. This will make it easy for log aggregation tools to consume log output in our dev/stg/prod clusters, especially for multi-line log content, log messages with stack traces, etc..

Also cleaned up a bug or two along the way.

## Key questions for reviewers
- Are the two new global config maps for Spring Boot containers that I introduced a good structure? Names okay?
- Is `spring-boot-logback-json` a good name for the library? Actually, it's not even a library, just a reference to two JARs. Should it be a `pom` artifact instead?
- Are we okay with JSON logging being disabled by default? This seemed a bit more user-friendly for local development, in particular.
- Are we okay with the use of Spring profiles? This is activated with `SPRING_PROFILES_ACTIVE=json-logging` environment variable.
- Are we okay with global Helm values? Is `OpenNMS.global.enableJsonLogging` a good name? Note: I'm trying to fit into the existing naming structure, but also trying to adhere to the [Helm value naming conventions](https://helm.sh/docs/chart_best_practices/values/#naming-conventions) for anything new. We should come back and lowercase the initial character in our values before we get too far down the road, BTW.

## (near) future improvements
- Get JSON log support working in `rest-server`. It's running an older 2.x version of Spring Boot, so the solution as we have here doesn't work for it.
- Add JSON log support to the Minion
- Add things like the tenant ID, location, etc. to MDC

## Jira link(s)
- https://issues.opennms.org/browse/HS-1419

## Checklist

## Other things that helped out
* https://dev.to/luafanti/spring-boot-logging-with-loki-promtail-and-grafana-loki-stack-aep
* https://vishnu-g.medium.com/creating-custom-spring-boot-starter-to-solve-cross-cutting-concerns-d76f555beb64
* https://medium.com/@asegu/build-your-own-spring-boot-starter-90ad1f426227
* https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin

## Other approaches
* https://github.com/blueswen/spring-boot-observability#loki-docker-driver
* https://spring.io/blog/2022/10/12/observability-with-spring-boot-3

## Interesting thing I stumbled across that we use
* https://github.com/Hakky54/log-captor -- to help capture logs for tests!

## For future minion work
* https://ops4j1.jira.com/wiki/spaces/paxlogging/pages/499286109/Configuration
* https://github.com/ops4j/org.ops4j.pax.logging/blob/main/pax-logging-logback/pom.xml
* https://stackoverflow.com/a/24685621
* https://ops4j1.jira.com/wiki/spaces/paxlogging/pages/499580944/Using+Pax+Logging+API+without+backend

[HS-1419]: https://opennms.atlassian.net/browse/HS-1419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HS-1419]: https://opennms.atlassian.net/browse/HS-1419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/alert/pom.xml
+++ b/alert/pom.xml
@@ -267,6 +267,9 @@
                             <jvmFlag>--add-opens=java.base/java.text=ALL-UNNAMED</jvmFlag>
                             <jvmFlag>--add-opens=java.desktop/java.awt.font=ALL-UNNAMED</jvmFlag>
                         </jvmFlags>
+                        <extraClasspath>
+                            <entry>${application.jib.app.config}</entry>
+                        </extraClasspath>
                     </container>
                 </configuration>
                 <executions>
--- a/alert/pom.xml
+++ b/alert/pom.xml
@@ -41,6 +41,11 @@
     <dependencies>
         <dependency>
             <groupId>org.opennms.horizon.shared</groupId>
+            <artifactId>spring-boot-logback-json</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.opennms.horizon.shared</groupId>
             <artifactId>grpc-common-constants</artifactId>
             <version>${project.version}</version>
         </dependency>
--- a/parent-pom/pom.xml
+++ b/parent-pom/pom.xml
@@ -101,6 +101,7 @@
         <keycloak.version>20.0.3</keycloak.version>
         <liquibaseVersion>4.17.0</liquibaseVersion>
         <log4j2.version>2.19.0</log4j2.version>
+        <logback.contrib.version>0.1.5</logback.contrib.version>
         <logcaptor.version>2.7.10</logcaptor.version>
         <mapstruct.version>1.5.3.Final</mapstruct.version>
         <maven-resources.version>3.3.0</maven-resources.version>
@@ -964,7 +965,16 @@
                 <artifactId>kafka</artifactId>
                 <version>${testcontainers.version}</version>
             </dependency>
-
+            <dependency>
+                <groupId>ch.qos.logback.contrib</groupId>
+                <artifactId>logback-json-classic</artifactId>
+                <version>${logback.contrib.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>ch.qos.logback.contrib</groupId>
+                <artifactId>logback-jackson</artifactId>
+                <version>${logback.contrib.version}</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
--- a/parent-pom/pom.xml
+++ b/parent-pom/pom.xml
@@ -27,6 +27,7 @@
 
         <!-- Jib properties -->
         <application.jib.defaultGoal>dockerBuild</application.jib.defaultGoal> <!-- dockerBuild for local only; build to push also -->
+        <application.jib.app.config>/app/config</application.jib.app.config>
         <application.docker.skip>false</application.docker.skip>
         <application.docker.image>${application.docker.image.name}:${application.docker.image.tag}</application.docker.image>
         <!-- application.docker.image.name MUST be set by child modules that use jib-maven-plugin. -->
--- a/build-tools/basic/mk-docker-images.sh
+++ b/build-tools/basic/mk-docker-images.sh
@@ -72,7 +72,7 @@ time {
 	echo "==="
 	echo "=== DATACHOICES IMAGE"
 	echo "==="
-	mvn -f events install -Djib.container.creationTime=USE_CURRENT_TIMESTAMP -Dapplication.docker.image=opennms/horizon-stream-datachoices:local-basic
+	mvn -f datachoices install -Djib.container.creationTime=USE_CURRENT_TIMESTAMP -Dapplication.docker.image=opennms/horizon-stream-datachoices:local-basic
 
 	echo ""
 	echo "==="
--- a/charts/opennms-operator/.helmignore
+++ b/charts/opennms-operator/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/opennms/.helmignore
+++ b/charts/opennms/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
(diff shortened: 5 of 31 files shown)


parent-pom/pom.xml
```diff
@@ -954,7 +956,21 @@
...
+            <dependency>
+                <groupId>ch.qos.logback.contrib</groupId>
+                <artifactId>logback-jackson</artifactId>
+                <version>${logback.contrib.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.codehaus.janino</groupId>
+                <artifactId>janino</artifactId>
+                <version>${janino.version}</version>
+            </dependency>
```
artnaseef, line 973: What is janino?

shared-lib/spring-boot-starter-logback-json/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+
+<configuration>
+    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
```
artnaseef, line 5: Are these resource paths internals of spring, or are they part of a contract?

artnaseef (commented):
This is an interesting approach.

Is there a reason why using logback configuration files in our deployments to achieve the same results wouldn't work?  There may be other tweaks we want to make to logging, and externalizing the logback configs seems like a reasonable solution to that end.

parent-pom/pom.xml
deejgregor: Good eye. I was testing you. :-P

I initially used a different approach that used `janino` instead of spring profiles. This part needs to go, so I'll fix that.

deejgregor:
> Is there a reason why using logback configuration files in our deployments to achieve the same results wouldn't work?

@artnaseef There is no major reason why I didn't do it aside from "not there yet"--this is just what I could get working relatively easily with little experience with project so far.

A few reasons why I leaned this way *for now*:
1. There is one reason why it can't **just** be done with logback config files--it also [needs two libraries](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/75707eb8ccb39a73c88d3227f227f73bbfb6622d#diff-119c0c4de9656f931269f214a01e8a3e4c9f54bff1c8e21632b6959fa4a1204eR29).
2. I didn't see anything externalizing files yet--at least in the few examples I looked at. :-) If I missed any examples, please let me know.
3. Also, I wanted something that I could update once and have it apply to all of our spring boot containers, and this seemed like a good way to go to start with.
4. Lastly, I figured that starting a starter (heh) would be useful to start standardizing on some platform-ish stuff that we'll want in all of our services.

> There may be other tweaks we want to make to logging, and externalizing the logback configs seems like a reasonable solution to that end.

Definitely, and I think that's a fine thing to externalize down the road (or now, if we prefer--I do tend to be more of an incrementalist, and I'm definitely itching to make multi-line logs more readable in our cloud solution ASAP).

This could probably be split into two pieces now without too much more work, I think:
1. A starter that does the libraries (they need to end up on the jib classpath for this to work properly, AFAICT)--or is there a better way to accomplish that?
2. Move the logback config into the Helm chart.

parent-pom/pom.xml
deejgregor: Done

artnaseef:
> 1. There is one reason why it can't **just** be done with logback config files--it also [needs two libraries](https://github.com/OpenNMS-Cloud/horizon-stream/pull/971/commits/75707eb8ccb39a73c88d3227f227f73bbfb6622d#diff-119c0c4de9656f931269f214a01e8a3e4c9f54bff1c8e21632b6959fa4a1204eR29).

Ahh, that definitely requires a source change.

> 2. I didn't see anything externalizing files yet--at least in the few examples I looked at. :-) If I missed any examples, please let me know.

True - we haven't gotten to a place were logging needed to be tweaked.  At least, not that I am aware.

> 3. Also, I wanted something that I could update once and have it apply to all of our spring boot containers, and this seemed like a good way to go to start with.

There is value to making this opinionated and an easy drop-in.  OTOH, my gut is saying that leaving it to downstreams to drop in the logback config is the right way to go - just because there are a LOT of things that might get adjusted over time, and having our own template can get in the way.

> This could probably be split into two pieces now without too much more work, I think:
> 
> 1. A starter that does the libraries (they need to end up on the jib classpath for this to work properly, AFAICT)--or is there a better way to accomplish that?
> 2. Move the logback config into the Helm chart.

Just adding those 2 as dependencies to each project will be sufficient - Spring Boot's all-in-one packaging picks up all project dependencies.  It does not do any special resolution/filtering on them.

jaymberry (approved):

alert/pom.xml
```diff
@@ -257,6 +262,9 @@
                             <jvmFlag>--add-opens=java.base/java.text=ALL-UNNAMED</jvmFlag>
                             <jvmFlag>--add-opens=java.desktop/java.awt.font=ALL-UNNAMED</jvmFlag>
                         </jvmFlags>
+                        <extraClasspath>
+                            <entry>${application.jib.app.config}</entry>
+                        </extraClasspath>
```
artnaseef, line 267: The `extraClasspath` makes sense in the `jib-maven-plugin`; when I saw it earlier, I thought it was in the `spring-boot-maven-plugin`.

artnaseef (approved):

PR #972, opened 2023-04-10
Jer/secure minion adr

jbeuree:
## Description
Adding the secure minion ADR after updates from review.


JohnWang-NMS (approved):
Do we need to support certificate rotation? If not, how do we ensure the expiry date on the Minio (expiry being indefinite can still be some remote date such as the end of the century)  matches the server side (in cloud)?

jbeuree:
This decision is just around deciding to always use the secure approach. The actual design and evolution is documented separately.

PR #973, opened 2023-04-10
AR-30: ADR for cross-team APIs

geraldhumphries:
## Description

This ADR is about how to manage cross-team APIs and proto files. It has already been merged as "Proposed", but I've made some changes to it and opened this PR so it can be reviewed and discussed.

Please leave feedback as comments on this PR, suggest changes, or approve it if you think everything looks good. Before it gets merged I will change the status to "Approved" or "Rejected" depending on what we decide.

## Jira link(s)
- https://issues.opennms.org/browse/AR-30

### DIFF
--- a/doc/architecture/decisions/0002-store-cross-project-api-contracts-in-a-separate-repository.md
+++ b/doc/architecture/decisions/0002-store-cross-project-api-contracts-in-a-separate-repository.md
@@ -22,21 +22,17 @@ We need to publish libraries for Java to a publicly accessible Maven repository.
 
 We should automate this process, and a CI system like GitHub Actions would be a good fit. It can validate the proto files, generate the code, package it into an artifact, and publish it to the repository without any human intervention.
 
-
-
 ## Decision
 
 We will store cross-project API contracts in a new git repository. Changes to these API contracts will be done with pull requests. All proto files that are shared between projects will be moved to this new repository.
 
 In order for a PR to be merged, a team member from each of its dependent projects will be required to approve it. A GitHub CODEOWNERS file can be used to enforce this by specifying each of the teams as an owner of the proto file.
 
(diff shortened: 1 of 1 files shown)


JohnWang-NMS (approved):
when new version of the artifacts being released, will the new approach generate changelogs to catch the changes?

does it support backwards compatibility?

maximbrener:
Who will be the owner of that repo and the contract files?

geraldhumphries:
> Who will be the owner of that repo and the contract files?

There is no single owner of the repo or proto files, it should function as neutral ground where no team can unilaterally change the APIs.

The pipelines that produce the libraries can have an owner but that isn't defined here. It may be DevOps, or the teams who are interested in the Java/Go libs.

> when new version of the artifacts being released, will the new approach generate changelogs to catch the changes?
> 
> does it support backwards compatibility?

The changelogs would be part of the release process, but that isn't been defined yet. I think release notes are a great idea but I think they would need to be automated.

Backwards compatibility is supported two ways:

- There are [best practices for updating proto files](https://protobuf.dev/programming-guides/proto3/#updating), and a huge part of them are about maintaining backwards compatibility. We will need to ensure we follow them ourselves during code review.
- The libraries that projects will depend on will be versioned. If there is a breaking change introduced in a new version of each library, projects can continue depending on the old versions until they can be brought up to date.

geraldhumphries:
> Out of curiosity, why is a separate repo chosen over keeping it integrated in the mono repo?

This isn't for any of the APIs that are only used internally in Horizon Stream, it's for cross-project APIs that shouldn't have an owner. The reasoning for that is in the Context section. If it were in the monorepo then it would imply that the HS project/team is the owner, defeating the purpose.

>Also, would assume that we don't really need a strategy for deprecating major releases of these libraries, given this is all internal communication (e.g. compared to external third party libraries).

Right now, no, but we may want one in the future! That isn't defined here.

maximbrener (approved):

maximbrener:
LGTM. I think from POs perspective its good to go

PR #991, opened 2023-04-13
HS-1397 Add support for SMTP email notifications

BenD10:
## Description
Allow the notification service to reach out to the a configured SMTP server for sending emails. I can see the emails going through to MailHog.
image

## Jira link(s)
- https://issues.opennms.org/browse/HS-1397

## Checklist


notifications/src/test/java/org/opennms/horizon/notifications/api/email/SmtpEmailAPITest.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.Mockito.times;
+
+@RunWith(MockitoJUnitRunner.class)
```
vserrago, line 35: Why JUnit4 instead of Junit5? From what I can tell, it looks like this project's pom lists JUnit Jupiter dependencies.

notifications/src/main/java/org/opennms/horizon/notifications/api/email/SmtpEmailAPI.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private final JavaMailSender sender;
+
+    @Override
+    public void postNotification(List<String> emailAddresses, Alert alert) throws NotificationException {
+        try {
+            sendEmail(emailAddresses, alert.getLogMessage(), createEmailBody(alert));
+        } catch (MessagingException e) {
+            // TODO This may be a retryable exception
+            throw new NotificationAPIException(e);
```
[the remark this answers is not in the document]
BenD10: Was going to defer for a later PR since this is largish right now. Would also like to get the full alert flow working first and then come back to iterate on this.

notifications/src/test/java/org/opennms/horizon/notifications/api/email/SmtpEmailAPITest.java
BenD10: Yeah I'm not sure, I copied the style of the other unit tests which are all using 4, but some components use 5. It's not super consistent across the project. I can update this to JUnit 5 though.

notifications/src/main/java/org/opennms/horizon/notifications/api/keycloak/KeyCloakAPI.java
```diff
@@ -0,0 +1,35 @@
...
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.horizon.notifications.api.keycloak;
+
+import java.util.List;
+
+public interface KeyCloakAPI {
```
james8hut, line 33: Probably nit-picking, but is there a need to have an interface here? Feels like there'll only ever be one implementation of it. Probably just my personal preference not to have them, mainly due to how cumbersome they make debugging and navigating the code.

BenD10: I think there will have to be a second implementation that supports multitenancy, but KeyCloak in this repo doesn't support it.

notifications/src/test/java/org/opennms/horizon/notifications/api/email/SmtpEmailAPITest.java
vserrago: I see. Since the PR is large already I'll leave it up to you to address or not.

My general opinion with JUnit4/5 mixes is that unless there's a reason to stick with 4, ie a specific JUnit 4 Runner that doesn't have an equivalent JUnit5 extension, we should use Junit 5 for new test classes.

PR #998, opened 2023-04-14
[HS-1463] Support to azure key vault

bahka:
Jira Issue: https://opennms.atlassian.net/browse/HS-1463 

We need to store some sensitive data for tests, so KeyVault is a good option to do that.


chuklov (approved):
LGTM

PR #1009, opened 2023-04-17
HS-1376 Include monitoring policy ID in generated alerts

BenD10:
## Description
The monitoring policy ID is used by the notification service to determine which forms of notification to use for a given alert.

Note that the monitoring policy ID isn't saved alongside the alert in the DB since it can be easily recalculated using the trigger event as discussed with Yang.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1376

## Checklist


alert/src/main/java/org/opennms/horizon/alertservice/service/AlertEventProcessor.java
```diff
@@ -105,19 +103,24 @@ public Optional<Alert> process(Event e) {
...
             clearKey = String.format(alertDefinition.getClearKey(), event.getTenantId(), event.getNodeId());
         }
         TriggerEvent triggerEvent = triggerEventRepository.getReferenceById(alertDefinition.getTriggerEventId());
-        return new AlertData(reductionKey, clearKey, alertDefinition.getType(), triggerEvent.getSeverity(), triggerEvent.getCount(), triggerEvent.getOvertime(), triggerEvent.getOvertimeUnit());
+        // TODO: An alert could match multiple monitoring policies, each having their own notifications.
+        Optional<MonitorPolicy> policy = monitorPolicyRepository.findMonitoringPolicyByTriggerEvent(triggerEvent.getId());
+        if (policy.isPresent()) {
+            return new AlertData(reductionKey, clearKey, alertDefinition.getType(), triggerEvent.getSeverity(), triggerEvent.getCount(), triggerEvent.getOvertime(), triggerEvent.getOvertimeUnit(), List.of(policy.get().getId()));
+        }
+        return new AlertData(reductionKey, clearKey, alertDefinition.getType(), triggerEvent.getSeverity(), triggerEvent.getCount(), triggerEvent.getOvertime(), triggerEvent.getOvertimeUnit(), Collections.emptyList());
```
vserrago, line 118: Nit: I find these really long lines affect the readability of the code, and prefer they be wrapped. 
```suggestion
            return new AlertData(
                reductionKey,
                clearKey,
                alertDefinition.getType(),
                triggerEvent.getSeverity(),
                triggerEvent.getCount(),
                triggerEvent.getOvertime(),
                triggerEvent.getOvertimeUnit(),
                List.of(policy.get().getId()));
        }
        // Alternative wrapping
        return new AlertData(
            reductionKey, clearKey, alertDefinition.getType(), triggerEvent.getSeverity(),
            triggerEvent.getCount(), triggerEvent.getOvertime(), triggerEvent.getOvertimeUnit(),
            Collections.emptyList()
        );
```

alert/src/main/java/org/opennms/horizon/alertservice/service/AlertEventProcessor.java
```diff
@@ -105,19 +103,24 @@ public Optional<Alert> process(Event e) {
...
-    private AlertData getAlertData(Event event, AlertDefinition alertDefinition) {
+    private @Nullable AlertData getAlertData(Event event, AlertDefinition alertDefinition) {
         var reductionKey = String.format(alertDefinition.getReductionKey(), event.getTenantId(), event.getUei(), event.getNodeId());
         String clearKey = null;
         if (!Strings.isNullOrEmpty(alertDefinition.getClearKey())) {
             clearKey = String.format(alertDefinition.getClearKey(), event.getTenantId(), event.getNodeId());
         }
         TriggerEvent triggerEvent = triggerEventRepository.getReferenceById(alertDefinition.getTriggerEventId());
-        return new AlertData(reductionKey, clearKey, alertDefinition.getType(), triggerEvent.getSeverity(), triggerEvent.getCount(), triggerEvent.getOvertime(), triggerEvent.getOvertimeUnit());
+        // TODO: An alert could match multiple monitoring policies, each having their own notifications.
```
vserrago, line 113: Just to confirm, this TODO will be addressed relatively shortly in another PR, right?

vserrago (commented):
Overall looks good to me

alert/src/main/java/org/opennms/horizon/alertservice/service/AlertEventProcessor.java
BenD10: Hard to say, I'll have to discuss with @yangli2014 since it gets more complicated

vserrago: I see. My next question would be then, should it be captured in JIRA?

BenD10: Yes you're right, created https://opennms.atlassian.net/browse/HS-1485

vserrago (approved):

alert/src/main/java/org/opennms/horizon/alertservice/service/AlertEventProcessor.java
james8hut: Given the large number of parameters here, I'd prefer to only have one return statement here, as if there are modifications to the parameters it would be easy to make a mistake.

So code like:
if (policy.isPresent()) {
policies = List.of(policy.get().getId());
} else {
policies = Collections.emptyList()
}

return new AlertData(..., policies);

alert/src/main/java/org/opennms/horizon/alertservice/service/AlertEventProcessor.java
```diff
@@ -126,6 +129,9 @@ protected org.opennms.horizon.alertservice.db.entity.Alert addOrReduceEventAsAle
             return null;
         }
         AlertData alertData = getAlertData(event, alertDefOpt.get());
+        if (alertData == null) {
```
james8hut, line 132: Can this ever be null?

BenD10: Oops you're right, that's a holdout, we both refactored how the trigger events were retrieved and this is a holdout after I rebased on top of your changes.

alert/src/test/java/org/opennms/horizon/alertservice/stepdefs/MonitorPolicySteps.java
```diff
@@ -207,4 +211,25 @@ public void verifyMonitoringPolicyTopic(String tenant) {
             assertEquals(1, messages.size());
         });
     }
+
+    @Then("valid monitoring policy ID is set in alert for tenant {string}")
```
james8hut, line 215: NIT: I would add the word Verify (1) for consistency purposes and (2) for clarity as to what is actually happening when someone reads the feature file.
@Then("Verify valid monitoring policy ID is set in alert for tenant {string}")

james8hut (commented):
Two suggested changes and one question.

Sorry, I should have put them all in a review, rather than adding them as single comments.

PR #1012, opened 2023-04-17
[HS-1430] Prepare an instance and minion

bahka:
- Add testcontainers and MinionContainer
- Move all @Before/@After method to CucumberHooks 
- add  Before(@cloud) for cloud tests (@cloud) that prepare a new cloud instance, connect a new minion to it, and login to the cloud instance.
- separate features for portal and cloud


chuklov (approved):
LGMT

PR #1017, opened 2023-04-17
[HS-1480] Add Udpgen container that sends flows traffic to test minion

bahka:
- `Update` testcontainers version to the latest one (to expose udp ports)
- `Add` UdpGen container 
- `Update` MinionContainer (expose UDP ports + method to obtain mapped UDP port)
- `Add` steps for UdpGen to send flows data to test minion

Overall: now we can check cases with Flows graphs.

### DIFF
--- a/external-it/hs-system-tests/src/test/java/testcontainers/MinionContainer.java
+++ b/external-it/hs-system-tests/src/test/java/testcontainers/MinionContainer.java
@@ -1,9 +1,12 @@
 package testcontainers;
 
+import com.github.dockerjava.api.model.ExposedPort;
+import com.github.dockerjava.api.model.InternetProtocol;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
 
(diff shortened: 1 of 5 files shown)


external-it/hs-system-tests/src/test/java/testcontainers/MinionContainer.java
```diff
@@ -12,10 +15,20 @@ public class MinionContainer extends GenericContainer<MinionContainer> {
     public String minionId;
     public String minionLocation;
```
chuklov, line 17: Let's make them variables instead of comments.

bahka: I would like to keep only variables that we want to re-use further. 

`withExposedPorts(8101, 8181)` works in combination with 
`.waitingFor(Wait.forListeningPort().withStartupTimeout(Duration.ofMinutes(5)));`

chuklov (approved):
LGTM

PR #1023, opened 2023-04-18
[HS-1461] clean up after tests

bahka:
Tests create some instances, so now they delete all of them when they finished.


chuklov (approved):
LGMT

PR #1029, opened 2023-04-19
HS-1496 Make sure trust manager is configured only if custom CA file exists

splatch:
## Description
This PR bring back previous behavior and allows to connect minion to cloud instances.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1496
- 
## Checklist

### DIFF
--- a/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/ssl/MinionGrpcSslContextBuilderFactoryImpl.java
+++ b/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/ssl/MinionGrpcSslContextBuilderFactoryImpl.java
@@ -46,7 +46,9 @@ import java.security.NoSuchAlgorithmException;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.util.function.Supplier;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class MinionGrpcSslContextBuilderFactoryImpl implements MinionGrpcSslContextBuilderFactory {
 
     @Setter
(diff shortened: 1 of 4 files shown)


minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/ssl/MinionGrpcSslContextBuilderFactoryImpl.java
```diff
@@ -73,19 +75,31 @@ public class MinionGrpcSslContextBuilderFactoryImpl implements MinionGrpcSslCont
     public SslContextBuilder create() {
         SslContextBuilder builder = grpcSslClientContextFactory.get();
 
-        if (trustCertCollectionFilePath != null) {
-            builder.trustManager(new File(trustCertCollectionFilePath));
+        if (trustCertCollectionFilePath != null && !trustCertCollectionFilePath.isBlank()) {
+            File trustCertCollectionFile = new File(trustCertCollectionFilePath);
```
cgorantla, line 79: May be better to move  these checks into one method such as `isTlsEnabled()` ?  and use that for enabling TLS

splatch: This piece of code is called only if `isTlsEnabled` is true, but at `MinionGrpcClient`. This code was extracted from client code itself, thus it is a bit harder to spot that requirement.

minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/ssl/MinionGrpcSslContextBuilderFactoryImpl.java
```diff
@@ -73,19 +75,31 @@ public class MinionGrpcSslContextBuilderFactoryImpl implements MinionGrpcSslCont
...
+            if (trustCertCollectionFile.exists()) {
+                builder.trustManager(trustCertCollectionFile);
+            } else {
+                throw new RuntimeException("Configured trust store" + trustCertCollectionFile.getAbsolutePath() + " does not exist");
```
cgorantla, line 83: But when you throw exception, MinionGrpcClient component won't start at all, right ?
How will it solve the original issue ?

splatch: Yes, connection will not be started. I initially had warnings but then changed to exception cause user supplied configuration which is required to make connection is wrong.
The log statement can be easily not seen.

artnaseef: Misconfigured GRPC is worthy of fail-fast, so the application failing.  The original issue was that the code required a trust-store, now the `isSet` check makes it optional.

artnaseef (approved):

PR #1034, opened 2023-04-20
HS-1384: Improve stability/recovery of gRPC connections

j-white:
JIRA: https://opennms.atlassian.net/browse/HS-1384

This doesn't solve the problem completely, but does address some known stability issues which I believe are worth getting in while we continue to dive deeper.

Changes include:
* Upgrade gRPC from 1.43.2 to 1.54.1 (includes a Netty upgrade and addresses some race conditions)
* Address synchronization problems in the SimpleReconnectStrategy class

### DIFF
--- a/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/SimpleReconnectStrategy.java
+++ b/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/SimpleReconnectStrategy.java
@@ -2,13 +2,16 @@ package org.opennms.horizon.minion.grpc;
 
 import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 public class SimpleReconnectStrategy implements Runnable, ReconnectStrategy {
-
+    private static final Logger LOG = LoggerFactory.getLogger(SimpleReconnectStrategy.class);
     private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
     private final ManagedChannel channel;
     private final Runnable onConnect;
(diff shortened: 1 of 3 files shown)


minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/SimpleReconnectStrategy.java
```diff
@@ -22,18 +25,24 @@ public SimpleReconnectStrategy(ManagedChannel channel, Runnable onConnect, Runna
     }
 
     @Override
-    public void activate() {
+    public synchronized void activate() {
+        if (reconnectTask != null) {
+            // Silently ignore additional activate requests if one is already in progress
```
deejgregor, line 30: Is there any value in knowing when we reached this block? a debug or a trace message?

(I'm not familiar at all with this to know, but I always worry a little bit when we silently ignore things, as those can be messy to debug when something odd happens)

j-white: It's normal/expected for this to be called multiple times, so nothing worth logging really.
Could add something on DEBUG/TRACE if we really want to know though.

minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/SimpleReconnectStrategy.java
```diff
@@ -22,18 +25,24 @@ public SimpleReconnectStrategy(ManagedChannel channel, Runnable onConnect, Runna
     }
 
     @Override
-    public void activate() {
+    public synchronized void activate() {
```
artnaseef, line 28: These 2 methods using `synchronized` keyword with callbacks are creating coarse critical sections (i.e. not light and fast).  The `onConnect` callback makes a GRPC method call to create a stream, and sends a message over that stream.

j-white: Good point. Will need a different approach then.

artnaseef (approved):

PR #1038, opened 2023-04-20
HS-1398 Add support for Azure Communication Service Email

BenD10:
## Description
Now Horizon Stream supports two methods of sending emails for
notifications, regular SMTP or via Azure Communication Services.
The implementation will be selected based provided properties.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1398

## Checklist


parent-pom/pom.xml
```diff
@@ -154,6 +154,7 @@
         <jmh.version>1.33</jmh.version>
         <zip4j.version>2.11.5</zip4j.version>
         <velocity.engine.version>2.3</velocity.engine.version>
+        <azure-communication-email.version>1.0.0</azure-communication-email.version>
```
vserrago, line 157: Huh, I didn't realize, but ECS [was released](https://techcommunity.microsoft.com/t5/azure-communication-services/simpler-faster-azure-communication-services-email-now-generally/ba-p/3788541)

notifications/src/test/java/org/opennms/horizon/notifications/config/EmailConfigTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Test
+    void testACSUsed() {
+        EmailClient mockClient = Mockito.mock(EmailClient.class);
+
+        try(MockedConstruction<EmailClientBuilder> builder = Mockito.mockConstruction(EmailClientBuilder.class, (mock, ctx) -> {
+            Mockito.when(mock.retryPolicy(any())).thenReturn(mock);
+            Mockito.when(mock.connectionString(any())).thenReturn(mock);
+            Mockito.when(mock.buildClient()).thenReturn(mockClient);
+        })) {
```
vserrago, line 30: Nit: I tend to not like constructor mocking, but I guess the alternative would be adding an `EmailClientBuilderFactory` 🥗

notifications/src/main/java/org/opennms/horizon/notifications/api/email/EmailAPI.java
```diff
@@ -28,11 +28,8 @@
...
 package org.opennms.horizon.notifications.api.email;
 
-import org.opennms.horizon.alerts.proto.Alert;
 import org.opennms.horizon.notifications.exceptions.NotificationException;
 
-import java.util.List;
-
 public interface EmailAPI {
     public void sendEmail(String emailAddress, String subject, String body) throws NotificationException;
 }
```
vserrago, line 35: It looks like both ACS and the SMTP implementations expect HTML for the body, but the interface signature is ambiguous; I think we should make it clear here that we're expecting HTML text via javadoc, parameter name (ie `bodyHtml`), or both.

vserrago (changes requested):

notifications/src/test/java/org/opennms/horizon/notifications/config/EmailConfigTest.java
BenD10: Yeah that seemed a little overkill 😅

vserrago (approved):

notifications/src/main/java/org/opennms/horizon/notifications/api/email/SmtpEmailAPI.java
```diff
@@ -31,18 +31,13 @@
...
 import org.springframework.mail.MailException;
 import org.springframework.mail.javamail.JavaMailSender;
 import org.springframework.mail.javamail.MimeMessageHelper;
-import org.springframework.stereotype.Component;
 
-import java.util.List;
-
-@Component
 @RequiredArgsConstructor
 public class SmtpEmailAPI implements EmailAPI {
```
yangli2014, line 42: Same comment as the above one.

yangli2014 (approved):
Approved with minor comments

notifications/src/main/java/org/opennms/horizon/notifications/api/email/ACSEmailAPI.java
```diff
@@ -0,0 +1,59 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+
+/**
+ * Implements email using Azure Communication Services instead of SMTP
+ */
+@RequiredArgsConstructor
+@ConditionalOnBean(EmailClient.class)
+public class ACSEmailAPI implements EmailAPI {
```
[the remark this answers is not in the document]
BenD10: This one I disagree with, anything that isn't an interface is already an implementation. The standard library doesn't have ArrayListImpl or LinkedListImpl since they implement the List interface. I think this raises a point that this repo maybe needs a style guide?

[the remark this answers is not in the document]
yangli2014: In that case, don't use "API" at the end of the name. It is just my opinion.

PR #1039, opened 2023-04-20
HS-1405 Add locations minions list

chinh-cle:
## Jira link(s)
- https://issues.opennms.org/browse/HS-1402

## Checklist


rcorry93 (commented):
What is this new component for? is it the footer section of a dialog or form? It may need a better name to describe where it should be placed.

chinh-cle:
@rcorry93 ya I'm thinking to have it as a generic, thus we could use it wherever we need, without worrying about style, also to keep consistency, as with the Headlines...

rcorry93 (approved):
LGTM - If we have another use in the future for some sort of footer, we can rename it then, as of now, it may be the only footer component we will need.

anyarybalova (approved):

PR #1043, opened 2023-04-21
[HS-1504] Delete instance feature

bahka:
- Add PageObject and Steps for DeleteInstance feature
- Add scenarios for DeleteInstance feature


chuklov (approved):
LGMT

PR #1044, opened 2023-04-21
Jira/hs 1464 pageobjects

chuklov:
## Description
Page object for Appliance page

## Jira link(s)
- https://issues.opennms.org/browse/HS-1464


external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private static final SelenideElement addDeviceButton = $("[data-test='add-node-btn']");
+    private static final SelenideElement deviceNameInput = $("[data-test='name-input'] input");
+    private static final SelenideElement deviceIpInput = $("[data-test='ip-input'] input");
+
+    private static final SelenideElement saveButton = $("[data-test='save-btn']");
+    private static final SelenideElement cancelButton = $("[data-test='cancel-btn']");
+
+    public static void setDeviceNameInput(String name) {
+        deviceNameInput.sendKeys(name);
```
bahka, line 25: usually I add .shouldBe(enabled) for all actions (before sendKeys, click, etc) - to be sure that it's not disabled (it checks it visible and enabled and wait this state for 4 seconds by default).

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static void setDeviceIpInput(String ipAddress) {
```
bahka, line 28: `setDeviceIp` I suppose Input is extra here

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static void clickSaveButton() {
+        saveButton.click();
+    }
+
+    public static void clickCancelButton() {
+        cancelButton.click();
+    }
+
+    public static void clickAddDevice() {
```
bahka, line 40: you have some method with `button` and some without. so it's better to use the same naming patter

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static void clickRemoveMinion() {
+        minionRemoveButton.click();
+    }
+
+    public static boolean checkIsRemoveButtonShown() {
+        return minionRemoveButton.isDisplayed();
+    }
+
+    public static boolean checkIsAddDeviceButtonVisible() {
```
bahka, line 69: I prefer keep selenide check instead of assert, since selenide one does screenshots.

also, isDisplayed - is a current status, when you use `shouldBe(x)` it waits up to 4 seconds, it's nice since UI works not so fast .

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
```diff
@@ -0,0 +1,73 @@
+package org.opennms.horizon.systemtests.pages.cloud;
```
bahka, line 1: copyrights (c)

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/ConfirmDeleteMinionPopup.java
```diff
@@ -0,0 +1,15 @@
+package org.opennms.horizon.systemtests.pages.cloud;
```
bahka, line 1: copyrights (c)

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Then("check {string} minion in the list")
+    public static void waitWhenMinionAppear(String minionName) {
+        if (minionName.equals("DEFAULT")) {
+            minionName = MINIONS.get(0).minionId;
+        }
+
+        for (int i = 0; i < 2; i++) {
+            // since the minion doesn't show up we need to wait and refresh the page
+            if (!$(byText(minionName.toUpperCase())).isDisplayed()) {
```
bahka, line 51: it's already should be upper case in MINIONS.

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
```diff
@@ -0,0 +1,99 @@
...
+import static com.codeborne.selenide.Selenide.$;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.opennms.horizon.systemtests.CucumberHooks.MINIONS;
+
+public class ApplianceSteps {
+
+    @Then("check {string} minion in the list")
+    public static void waitWhenMinionAppear(String minionName) {
+        if (minionName.equals("DEFAULT")) {
```
bahka, line 45: I do we have another cases? If we operate only by default minion we can simplify it with hidden minion name.

also we have `TestDataStorage` class where we can add a method : mapMinionName(minionName) to avoid this condition check in different steps. (the same as emails)

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
```diff
@@ -0,0 +1,99 @@
...
+            // since the minion doesn't show up we need to wait and refresh the page
+            if (!$(byText(minionName.toUpperCase())).isDisplayed()) {
+                waitHeartbeat(1);
+            }
+        }
+    }
+
+    @Then("wait for {long} heartbeats")
+    public static void waitHeartbeat(long count) {
+        Selenide.sleep(count * 30_000);
```
bahka, line 59: I just have thoughts that we should do that by API call, later

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
```diff
@@ -0,0 +1,99 @@
...
+        Selenide.refresh();
+    }
+
+    @Then("check that the remove Minion button is displayed")
```
bahka, line 63: I suppose it's `delete` button. 
Also I don't think that it make sense to check it's visibility  (the only visibility `isDisplayed`) and action in 2 different steps.  I prefer just functional style - we verify that some element is visible by acting with it (click, set, get text...) 

when you do `click` on element if you use `shouldBe(enabled)` it will be automatically checked.

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
```diff
@@ -0,0 +1,99 @@
...
+    public void checkIsRemoveMinionButtonShown() {
+        assertTrue(AppliancePage.checkIsRemoveButtonShown());
+    }
+
+    @Then("check the status of the minion is {string}")
+    public void checkTheStatusOfTheMinionIs(String status) {
+        AppliancePage.waitMinionStatus(status);
+    }
+
+    @Then("click on the delete button for minion")
```
bahka, line 73: the button name I set in ' ' even if it's just icon button.

external-it/hs-system-tests/src/test/resources/cloud-features/appliance-tests.feature
```diff
@@ -0,0 +1,28 @@
+@cloud
+Feature: User can see a minion connected to the instance
+
+  Scenario: Verify adding a device manually
+    Given Navigate to the "appliances" through the left panel
+    Then check 'Add Device' button is accessible and visible
```
bahka, line 6: just click on this button (use .shouldBe(enabled) to verify visibility and accessibility) for all that cases when we check visibility and then act with the field in 2 steps

external-it/hs-system-tests/src/test/resources/cloud-features/appliance-tests.feature
```diff
@@ -0,0 +1,28 @@
...
+    Then click on 'Add Device' button to open a pop up window
```
bahka, line 7: do we need this part `to open a pop up window`? 

I add another step for such case:

> Then `name` popup appear 

(to wait when this popup will appear) also I have the opposite case to check validation case (when popup isn't closed)

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
```diff
@@ -0,0 +1,99 @@
...
+    public void checkAddDeviceButtonIsAccessibleAndVisible() {
+        assertTrue(AppliancePage.checkIsAddDeviceButtonVisible());
+    }
+
+    @Then("click on 'Add Device' button to open a pop up window")
+    public void clickOnAddDeviceButtonToOpenAPopUpWindow() {
+        AppliancePage.clickAddDevice();
+    }
+
+    @Then("fill details of a newly added device with name {string} and ip address {string}")
```
bahka, line 88: do you consider to have more atomic steps 

@Then("set device name as {string}")
@Then("set device ip as {string}")
@Then("click 'SAVE' button")

external-it/hs-system-tests/src/test/resources/cloud-features/appliance-tests.feature
```diff
@@ -0,0 +1,28 @@
...
+    Then fill details of a newly added device with name "Localhost" and ip address "127.0.0.1"
+    Then check the status of the device with name "Localhost" as status "DOWN"
+
+  Scenario: Verify that user can delete a minion with status DOWN
+    Given Navigate to the "appliances" through the left panel
+    Then check "DEFAULT" minion in the list
+    Then check the status of the minion is "UP"
```
bahka, line 14: I assume it worth to add a step to verify that we don't have this 'delete' button when it's UP

bahka (changes requested):

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/steps/cloud/ApplianceSteps.java
chuklov: This is different functionality. The button should be visible only in case the minion is inactive. At this step, I'm not interacting with it. I'm checking the functionality of the button visibility.

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
chuklov: isDisplayer returns false or true while shouldBe(x) crashes in case the element is not displayed.

external-it/hs-system-tests/src/test/java/org/opennms/horizon/systemtests/pages/cloud/AppliancePage.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private static final SelenideElement addDeviceButton = $("[data-test='add-node-btn']");
+    private static final SelenideElement deviceNameInput = $("[data-test='name-input'] input");
+    private static final SelenideElement deviceIpInput = $("[data-test='ip-input'] input");
+
+    private static final SelenideElement saveButton = $("[data-test='save-btn']");
+    private static final SelenideElement cancelButton = $("[data-test='cancel-btn']");
+
+    public static void setDeviceNameInput(String name) {
+        deviceNameInput.shouldBe(visible, enabled).sendKeys(name);
```
bahka, line 55: `enable` is enough, it's a mix of `visible` and `clickable`

bahka: FYI: all inputs are working in that way:

sendKeys - adds the text to existed text. (I suppose UI decided that)

In case if you want to change text in that input you need to use :
.setValue("").sendKeys(txt) 

if not - it's ok as is. (it make sense when you check validation cases)

bahka (approved):
LGTM

PR #1048, opened 2023-04-21
HS-1467: Add users.properties

cgorantla:
## Description
This will set default users.properties in Karaf.
Will facilitate easy acess to karaf shell.
Users.properties content copied from Classic.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1467

## Checklist


splatch (approved):
:+1:  If we make this, then I think its worth to remove `karafUsers` from minion config map we deploy into kubernetes.

j-white (approved):
👍

artnaseef (commented):
Are we not concerned about the security implications of backing in admin/admin into the base assembly that is used by all our deployments (including production)?

I would feel a lot better if it were somehow kept out of that base image and used in a dev-only image.

cgorantla:
> Are we not concerned about the security implications of backing in admin/admin into the base assembly that is used by all our deployments (including production)?
> 
> I would feel a lot better if it were somehow kept out of that base image and used in a dev-only image.

Since this is a configurable item, we can always override the defaults.

PR #1053, opened 2023-04-24
HS-1523 Add location edit form

chinh-cle:
## Jira link(s)
- https://issues.opennms.org/browse/HS-1523

## Checklist


ui/src/components/Common/ButtonText.vue
```diff
@@ -1,26 +1,35 @@
...
-    default: 'button'
+type ButtonText = {
+  label: string | undefined
+  type?: string
+  cb: fncArgVoid
+  cbArgs?: {
+    id: number
   }
-})
+  textTransform?: string
```
mikewrosey, line 25: This works but do we need this as a prop? If there is a class/style added to `<ButtonText>` it will inherit it and we could avoid a line or two of js. Also, it seems like we could just pass in the label the way we want it, like `GENERATE KEYS` etc.

ui/src/components/Locations/LocationsEditForm.vue
```diff
@@ -0,0 +1,282 @@
...
+}
+
+const downloadCredentials = {
+  label: 'ready to download',
+  cb: () => ({}),
+  textTransform: 'uppercase'
+}
+
+const instructions = {
+  textButton: 'Needd instructions for minion deployment at this location?',
```
mikewrosey, line 160: small typo `Needd`

ui/src/components/Common/FooterSection.vue
```diff
@@ -24,12 +24,13 @@
 </template>
 
 <script setup lang="ts">
-import { fncVoid } from '@/types'
+import { fncVoid, fncArgVoid } from '@/types'
+import { DisplayType } from '@/types/locations.d'
```
mikewrosey, line 28: If this enum is for a common component like this, we might want to bring it out of locations and put it in the index.ts file

ui/src/components/Common/InstructionsStepper.vue
```diff
@@ -119,112 +119,108 @@ const icon = markRaw({
...
 
-.btn-open-toggle {
-  color: var(variables.$primary);
-  border: none;
-  background-color: transparent;
-  padding: 0;
-  margin-bottom: var(variables.$spacing-l);
-  &:hover {
-    cursor: pointer;
+.instructions-stepper {
```
mikewrosey, line 122: maybe this is still being worked on but the stepper for me goes from step 1 to 4 when clicking on the chevron. Can still reach the other steps if I click on the numbers though.

mikewrosey (approved):

anyarybalova (approved):

ui/src/components/Common/InstructionsStepper.vue
chinh-cle: @mikewrosey any suggestion for the name instead of stepper?

mikewrosey: For this component? Name seems good to me.

PR #1055, opened 2023-04-24
Add Checkstyle to (eventually) enforce inclusion of copyright headers

j-white:
Copyright headers are missing from many source files in the code-base, and this is often, but not always highlight in PRs.

In order to remove the ambiguity we should enforce this at compile time, requiring the headers to be present, following a known format, and leaving facilities for exceptions.

In this PR, we add Checkstyle to the parent pom, leaving it disabled by default for the time being, and update the 'minion' project to compile cleanly with it enabled.

Next steps:
* Update headers in all of the other projects
* Enable the profile by default
* Stop having to point out missing copyright headers during PRs


j-white:
No feedback from anyone? :(

PR #1063, opened 2023-04-26
HS-1332: Passive discovery validations

rabzobi:
## Description
Allow update to keep the same location
Validate community string according to rules in ticket

## Jira link(s)
- https://issues.opennms.org/browse/HS-1332

## Flagged for review
Not sure of validation requirement for community strings and comma's. I have implemented as per my understanding from the ticket.

## Checklist
* [X] Follows Horizon Stream's [development guidelines.](https://github.com/OpenNMS/horizon-stream/wiki/Development-Guidelines)
* [X] Appropriate reviewer(s) have been selected.
* [X] Jira issue(s) have been updated to "In Review".

### DIFF
--- a/inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
+++ b/inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
@@ -171,9 +171,8 @@ public class PassiveDiscoveryService {
         if (snmpCommunities.length() > 128) {
             throw new InventoryRuntimeException("Snmp communities string is too long");
         }
-        for (byte b:snmpCommunities.getBytes()){
+        for (byte b: snmpCommunities.getBytes()){
             char c = (char) b;
-            log.info("b="+b+" c="+c);
             if (c > 127){
                 throw new InventoryRuntimeException("All characters must be 7bit ascii");
             }
--- a/inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
+++ b/inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
@@ -166,7 +166,7 @@ public class PassiveDiscoveryService {
     private void validateCommunityStrings(PassiveDiscoveryUpsertDTO passiveDiscovery) {
         String snmpCommunities = "";
         for (String snmpCommunity: passiveDiscovery.getCommunitiesList()){
-            snmpCommunities += snmpCommunity.replace(",","") + " "; // As be requirement on HS-1332
+            snmpCommunities += snmpCommunity.replace(",","") + " ";
         }
         if (snmpCommunities.length() > 128) {
             throw new InventoryRuntimeException("Snmp communities string is too long");
--- a/inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
+++ b/inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
@@ -163,7 +163,7 @@ public class PassiveDiscoveryService {
         }
     }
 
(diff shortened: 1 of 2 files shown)


james8hut (changes requested):
Any JUnit tests?

inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
```diff
@@ -160,6 +163,22 @@ private void validateSnmpPorts(PassiveDiscoveryUpsertDTO dto) {
         }
     }
 
+    private void validateCommunityStrings(PassiveDiscoveryUpsertDTO passiveDiscovery) {
+        String snmpCommunities = "";
+        for (String snmpCommunity: passiveDiscovery.getCommunitiesList()){
+            snmpCommunities += snmpCommunity.replace(",","") + " "; // As be requirement on HS-1332
```
james8hut, line 169: I'm not sure what the comment means exactly.
I'd also be inclined to remove it anyway.

rabzobi: It was supposed to say "as per"... but I have removed it

rabzobi:
> Any JUnit tests?

I have added some tests

inventory/main/src/test/java/org/opennms/horizon/inventory/service/PassiveDiscoveryServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        Exception exception = assertThrows(InventoryRuntimeException.class, () -> {
+            List<String> communities = new ArrayList<>();
+            communities.add("123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789");
+            PassiveDiscoveryUpsertDTO tooLong = PassiveDiscoveryUpsertDTO
+                .newBuilder().addAllCommunities(communities)
+                .build();
+            passiveDiscoveryService.validateCommunityStrings(tooLong);
+        });
+        assertTrue(exception.getMessage().equals("Snmp communities string is too long"));
```
james8hut, line 80: I would create a separate test method for each of these things. If the too long test fails then the 7 bit ascii test doesn't get to run, so I always find it annoying as a developer, if I was to fix the too long test, and only then find out the 7 bit ascii test was going to fail as well.

inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
```diff
@@ -97,8 +97,11 @@ public PassiveDiscoveryDTO updateDiscovery(String tenantId, PassiveDiscoveryUpse
             throw new InventoryRuntimeException("Passive discovery not found for id: " + id);
         }
 
-        validateDiscovery(tenantId, request);
+        if (!request.getLocation().equals(discoveryOpt.get().getLocation())) {
+            validateDiscovery(tenantId, request);
```
james8hut, line 101: Can you add a test to ensure validateDiscovery method isn't called.
Plan to do this:
Comment out the if statement
Test should validateDiscovery and throw an exception. (And fail)
Uncomment the if statement
Test now should pass, as it doesn't call validateDiscovery.

james8hut (changes requested):

inventory/main/src/main/java/org/opennms/horizon/inventory/service/discovery/PassiveDiscoveryService.java
rabzobi: I've added two heavily mocked test cases which cover the exception scenario (and no exception scenario)

inventory/main/src/test/java/org/opennms/horizon/inventory/service/PassiveDiscoveryServiceTest.java
rabzobi: I have moved them all into separate test methods

james8hut (approved):

PR #1083, opened 2023-05-02
HS-1310 Make sure runInPlace.sh works with tilt

splatch:
I've introduced two new switches:

- `-g AUTHORITY` - assigns custom authority to server, can be used with -h if host is an IP address which shows certificate for specific domain.
- `-T` - extract secrets via kubectl from development tilt setup. -s

## Jira link(s)
- https://issues.opennms.org/browse/HS-1310

## Checklist

### DIFF
--- a/minion/assembly/runInPlace.sh
+++ b/minion/assembly/runInPlace.sh
@@ -5,12 +5,12 @@ set -e
 MINION_ID='minion-standalone'
 MINION_LOCATION='minion-standalone-loc'
 IGNITE_SERVER_ADDRESSES=localhost
-MINION_GATEWAY_HOST=localhost
+MINION_GATEWAY_HOST=minion.onmshs.local
 # TODO - port 9443 for TLS?
-MINION_GATEWAY_PORT=8990
-MINION_GATEWAY_TLS="false"
+MINION_GATEWAY_PORT=1443
+MINION_GATEWAY_TLS="true"
 
-CERT_ROOTDIR="$(cd ../../tools/SSL; pwd)"
+CERT_ROOTDIR="$(pwd)/target"
 CA_CERT_FILE="${CERT_ROOTDIR}/CA.cert"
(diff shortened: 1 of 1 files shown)


minion/assembly/runInPlace.sh
```diff
@@ -64,11 +67,25 @@ do
...
+if [ "${EXTRACT_TILT_CERTS}" == "true" ]; then
+  mkdir -p $CERT_ROOTDIR || (echo "Could not create $CERT_ROOTDIR" && exit 1)
 
+  # extract client mtls ca certificate
+  kubectl get secret client-root-ca-certificate -ogo-template='{{index .data "tls.crt" }}' | base64 --decode > "$CERT_ROOTDIR/client-ca.crt"
+  kubectl get secret client-root-ca-certificate -ogo-template='{{index .data "tls.key" }}' | base64 --decode > "$CERT_ROOTDIR/client-ca.key"
+
+  openssl genrsa -out "$CERT_ROOTDIR/client.key.pkcs1" 2048
+  openssl pkcs8 -topk8 -in "$CERT_ROOTDIR/client.key.pkcs1" -out "$CLIENT_KEY_FILE" -nocrypt
+  openssl req -new -key "$CLIENT_KEY_FILE" -out "$CERT_ROOTDIR/client.unsigned.cert" -subj "/C=CA/ST=TBD/L=TBD/O=OpenNMS/CN=local-minion/OU=L:TestLocation/OU=T:opennms-prime"
```
artnaseef, line 84: The `OU=L:` needs to use the `MINION_LOCATION` setting.  Do we want to make Tenant ID configurable too?

splatch: Good point, I was actually asking myself same question about tenant. In local cluster we will always have a single tenant and there is no easy way to switch tenant used for UI. Bringing tenant as a variable into this script will let people test visibility of connected minion in UI.
Automated testing will not use this script so it will bring no value for e2e tests.

artnaseef (approved):

PR #1084, opened 2023-05-02
HS-1512 Use ad-hoc certificates for embedded development-time minion

splatch:
## Description
Use mtls by default for inner cluster minion which so far connected through minion-gateway-grpc-proxy.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1512

## Flagged for review
Any references to minion-gateway-grpc-proxy.

## Checklist


splatch:
@artnaseef Please have a look - with tls we are able to route inner cluster minion to the ingress controller. Scripts which are used with `runInPlace.sh` are also here to initialize client certificate and inject trusted CA certificate.

tilt-helm-values.yaml
```diff
@@ -40,6 +40,54 @@ OpenNMS:
...
+            openssl req -new -key /cert/client.key -out /cert/client.unsigned.cert -subj "/C=CA/ST=TBD/L=TBD/O=OpenNMS/CN=local-minion/OU=L:TestLocation/OU=T:opennms-prime"
+            openssl x509 -req -in /cert/client.unsigned.cert -days 14 -CA /run/secrets/mtls/tls.crt -CAkey /run/secrets/mtls/tls.key -out /cert/client.signed.cert
+        securityContext:
+          privileged: false
+          runAsUser: 10001
+          runAsGroup: 10001
+          fsGroup: 10001
+        volumeMounts:
+          - name: certificate
+            mountPath: "/cert"
```
artnaseef, line 80: In this extra container, the certificate folder is mounted at `/cert` while it is `/opt/karaf/certs` in the minion container, right?  If so, would be nice to have a comment to clarify.  It's not too hard to figure out, but it's not entirely obvious - especially when reading the `openssl` commands above.

artnaseef (approved):
One nit pick.  Otherwise, LGTM.  Really nice to see that grpc proxy get removed.

tilt-helm-values.yaml
splatch: I've added a bit longer note in `ExtraInitContainer`, which I renamed from ExtraContainer to make it consistent with place where it is applied.

PR #1088, opened 2023-05-03
HS-1384: use OK-HTTP instead of NETTY for GRPC connections to elimina…

artnaseef:
…te connection leaking.

- Update ingress-nginx version to v1.7.0
  - Updated cluster role permissions to match
- Use OK HTTP GRPC client channel code instead of NETTY GRPC client code
  - Almost completely eliminates lingering connections after a GOAWAY is received from the server (nginx)
  - Cleans up the lingering connections so we don't leak over time
  - Includes OK HTTP method of configuring SSL which is notably different than the NETTY method
- Basic build tools:
  - update helm chart values to add new certificates and SSL configuration
  - Use image "horizon-stream-minion-certificate-verifier:local-basic"

## Jira link(s)

## Checklist

### DIFF
--- a/build-tools/basic/compile.sh
+++ b/build-tools/basic/compile.sh
@@ -38,3 +38,4 @@ mvn clean install -P rapid-build -DskipTests=true -f notifications
 mvn clean install -P rapid-build -DskipTests=true -f metrics-processor
 mvn clean install -P rapid-build -DskipTests=true -f events
 mvn clean install -P rapid-build -DskipTests=true -f datachoices
+mvn clean install -P rapid-build -DskipTests=true -f minion-certificate-manager
--- a/build-tools/basic/compile.sh
+++ b/build-tools/basic/compile.sh
@@ -38,4 +38,3 @@ mvn clean install -P rapid-build -DskipTests=true -f notifications
 mvn clean install -P rapid-build -DskipTests=true -f metrics-processor
 mvn clean install -P rapid-build -DskipTests=true -f events
 mvn clean install -P rapid-build -DskipTests=true -f datachoices
-mvn clean install -P rapid-build -DskipTests=true -f minion-certificate-manager
--- a/build-tools/basic/mk-docker-images.sh
+++ b/build-tools/basic/mk-docker-images.sh
@@ -88,12 +88,6 @@ time {
 
 	echo ""
 	echo "==="
-	echo "=== CERTIFICATE MANAGER IMAGE"
-	echo "==="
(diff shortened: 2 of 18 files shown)


build-tools/basic/compile.sh
```diff
@@ -38,3 +38,4 @@ mvn clean install -P rapid-build -DskipTests=true -f notifications
 mvn clean install -P rapid-build -DskipTests=true -f metrics-processor
 mvn clean install -P rapid-build -DskipTests=true -f events
 mvn clean install -P rapid-build -DskipTests=true -f datachoices
+mvn clean install -P rapid-build -DskipTests=true -f minion-certificate-manager
```
splatch, line 41: As far I can see this module is listed in line 32 of this script

build-tools/basic/mk-docker-images.sh
```diff
@@ -86,6 +86,12 @@ time {
 	echo "==="
 	mvn -f datachoices install -Djib.container.creationTime=USE_CURRENT_TIMESTAMP -Dapplication.docker.image=opennms/horizon-stream-datachoices:local-basic
 
+	echo ""
+	echo "==="
+	echo "=== CERTIFICATE MANAGER IMAGE"
+	echo "==="
+	mvn -f minion-certificate-manager install -Djib.container.creationTime=USE_CURRENT_TIMESTAMP -Dapplication.docker.image=opennms/horizon-stream-minion-certificate-manager:local-basic
```
splatch, line 93: See line 39 of script.

build-tools/basic/push-images-to-kind.sh
```diff
@@ -51,6 +51,7 @@ kind load docker-image "opennms/horizon-stream-datachoices:local-basic"
 # 		"opennms/horizon-stream-inventory:local-basic" \
 # 		"opennms/horizon-stream-keycloak:local-basic" \
 # 		"opennms/horizon-stream-metrics-processor:local-basic" \
+# 		"opennms/horizon-stream-minion-certificate-manager:local-basic" \
```
splatch, line 54: See line 36.

splatch (changes requested):
Please update shell scripts as they do contain duplicate build calls which will impact build time.

build-tools/basic/compile.sh
artnaseef: Good catch

build-tools/basic/push-images-to-kind.sh
artnaseef: This one isn't actually a duplicate - notice it is commented-out.

build-tools/basic/mk-docker-images.sh
artnaseef: Another good catch.

splatch (approved):

PR #1093, opened 2023-05-04
Draft: HS-1582: Prevent Tilt from removing certificate secrets during 'tilt down'

geraldhumphries:
## Description

Uses the obscure `"tilt.dev/down-policy": keep` option to tell Tilt not to remove these resources. The only reference to it in the documentation is on [this page](https://docs.tilt.dev/cli/tilt_down.html).

Also adds a new DevMode flag to the chart that we should leverage for other things, like toggling debug ports on containers.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1582

## Flagged for review

- I haven't tested if we actually get the desired behaviour, just that these objects stick around.

## Checklist


splatch:
That's very nice finding, let me test it!

charts/opennms/values.yaml
```diff
@@ -1,6 +1,7 @@
 Host: onmshs
-Port: 443 #set depending on TLS.Enabled and the Ingress ports, do not change
-Protocol: https #set depending on TLS.Enabled, do not change
+Port: 443 # set depending on TLS.Enabled and the Ingress ports, do not change
+Protocol: https # set depending on TLS.Enabled, do not change
+DevMode: false # enable development-friendly options, not appropriate for production
```
deejgregor, line 4: nice-to-have: f you could, please change to `devMode`. I'm trying to get us to start following the [Helm naming conventions](https://v2-14-0.helm.sh/docs/chart_best_practices/#naming-conventions) on anything new (and eventually we'll come back and clean up what we have). I started doing that a few lines down with `global` and the keys below it.

charts/opennms/templates/cert/client-ca-certificate.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     "helm.sh/resource-policy": keep
+    {{- if .Values.DevMode }}
```
deejgregor, line 9: Do we ever run `Tilt` in a non-development mode? It might be awkward to have this annotation there all of the time, but I'm not sure we technically need to optionally include it. It will be a no-op unless something is looking for it, right? Maybe just always put it in like we do for the Helm resource policy.

charts/opennms/templates/cert/client-ca-certificate.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
...
+    "tilt.dev/down-policy": keep
```
deejgregor, line 10: Should we add anything to our docs making it clear that `tilt down` doesn't remove this objects (unless Tilt already says something like Helm does) and/or give users the command to remove these objects?

deejgregor (approved):
Approving. I had a few suggestions, but they are all optional. Happy to quickly re-review if any tweaks are made (just poke me on mattermost).

geraldhumphries:
@deejgregor with these changes it still doesn't keep the same certs around - Tilt applies the secrets with regenerated values and replaces what already exists. Saving them to the local filesystem and making Tilt apply them if they exist might be a much better solution. 

@splatch I moved cert-manager to a subchart but I am rethinking that now. Helm's dependency management is a bit of a pain, it requires adding Helm repos and updating deps ahead of time. I assume cert-manger will be dev-only, so making it a hard dependency probably isn't the way to go. The way you had it before might be a better overall solution.

splatch:
@deejgregor @geraldhumphries After doing HS-1512 I can tell that cert-manager use might be out of question. We can make static certs with scripts. If we will need to get a script for generating root CA's then we can easily get few more lines to generate these two ingress certs we need.

Overall I made HS-1512 working without cert-manager because I didn't want to rely on helm dependency management. It worked fine for integration tests. If we could achieve similar thing with Tilt firing scripts and feeding secrets into k8s, then we could unify development side and install-local which is used now primary by cicd pipeline.

geraldhumphries:
Closed, won't be using this solution.

PR #1095, opened 2023-05-04
HS-1363: cleanup all the streams on disconnect.

artnaseef:
## Jira link(s)
- https://issues.opennms.org/browse/HS-1363

## Checklist

### DIFF
--- a/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
+++ b/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
@@ -173,6 +173,7 @@ public class MinionGrpcClient extends AbstractMessageDispatcherFactory<String> i
         //  not match the hostname used to connect the server.
         if (! Strings.isNullOrEmpty(overrideAuthority)) {
             channelBuilder.overrideAuthority(overrideAuthority);
+            LOG.info("Configuring GRPC override authority {}", overrideAuthority);
         }
 
         if (tlsEnabled) {
@@ -202,7 +203,9 @@ public class MinionGrpcClient extends AbstractMessageDispatcherFactory<String> i
         if (rpcStream != null) {
             rpcStream.onCompleted();
         }
-        channel.shutdown();
+        if (channel != null) {
+            channel.shutdown();
+        }
         LOG.info("Minion at location {} with systemId {} stopped", ipcIdentity.getLocation(), ipcIdentity.getId());
     }
 
@@ -340,6 +343,7 @@ public class MinionGrpcClient extends AbstractMessageDispatcherFactory<String> i
                 asyncStub.cloudToMinionMessages(identity, new CloudMessageObserver(cloudMessageHandler));
                 LOG.info("Initialized cloud receiver stream");
             });
+
     }
 
     private void sendBlockingSinkMessage(SinkMessage sinkMessage) {
--- a/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
+++ b/minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
@@ -45,6 +45,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 
 import com.google.common.base.Strings;
+import io.grpc.Context;
 import io.grpc.okhttp.NegotiationType;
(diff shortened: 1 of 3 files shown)


minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
```diff
@@ -300,8 +336,14 @@ private void initializeCloudReceiver() {
             .setLocation(ipcIdentity.getLocation())
             .setSystemId(ipcIdentity.getId())
             .build();
-        asyncStub.cloudToMinionMessages(identity, new CloudMessageObserver(cloudMessageHandler));
-        LOG.info("Initialized cloud receiver stream");
+
+        cloudToMinionStreamCancellableContext = Context.current().withCancellation();
+        cloudToMinionStreamCancellableContext.run(
+            () -> {
+                asyncStub.cloudToMinionMessages(identity, new CloudMessageObserver(cloudMessageHandler));
```
splatch, line 343: I think code needs to be wrapped into condition cause context can be cancelled before runnable is executed.

```suggestion
                if (!Context.current().isCancelled()) {
                    asyncStub.cloudToMinionMessages(identity, new CloudMessageObserver(cloudMessageHandler));
                }
```

minion/minion-grpc/grpc-client/src/test/java/org/opennms/horizon/minion/grpc/MinionGrpcClientTest.java
```diff
@@ -108,6 +128,74 @@ void testStartTls() throws Exception {
             target.setTlsEnabled(true);
             target.start();
 
+            //
+            // Verify the Results
+            //
+            Predicate<LogEvent> matcher1 =
+                (logEvent) ->
+                    (
+                        Objects.equals("TLS enabled for gRPC", logEvent.getMessage() )
```
splatch, line 137: What this test is attempting to prove? It is better to verify interactions with mocked components which prove behavior of code under test. Log output which can be disabled at any time is not a reliable behavior verification.

splatch (approved):

minion/minion-grpc/grpc-client/src/test/java/org/opennms/horizon/minion/grpc/MinionGrpcClientTest.java
artnaseef: The verification is ensuring the TLS path is hit.  Unfortunately `channelBuilder` cannot be mocked, so it's not possible to verify that way.

minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
artnaseef: Do you have more information?  I found this snippet recommended online and followed that pattern.

artnaseef: Looking at that carefullly, what happens if we don't check `Context.current().isCancelled()` first?  Also, is there something we should be doing when it is cancelled?

artnaseef: Oh it seems there's another concern that creates; from the javadocs:

`Create a new context which is independently cancellable and also cascades cancellation from its parent. Callers must ensure that either [Context.CancellableContext.cancel(Throwable)](https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.CancellableContext.html#cancel-java.lang.Throwable-) or [Context.CancellableContext.detachAndCancel(Context, Throwable)](https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.CancellableContext.html#detachAndCancel-io.grpc.Context-java.lang.Throwable-) are called at a later point, in order to allow this context to be garbage collected.`

splatch: It is indeed note from javadoc and example of how to use cancellable context.  I suppose that calling `cancel(null)` when we shutdown channel or entire client is fine.

artnaseef: Good news - we do call cancel on it normally, so we should not have a leak.  The `handleDisconnect` method ensures that.

PR #1107, opened 2023-05-08
HS-1591: Refactor Spring Boot services to use Spring-provided KafkaTemplate

geraldhumphries:
## Description

- Removes our own KafkaTemplate beans in favour of the one provided by Spring
- Allows us to use all `spring.kafka` properties from Spring Boot's [common application properties](https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html)
- Fixes an error while saving a monitoring policy with a cold start condition
- Refactors alert service to follow Spring Boot conventions more closely
- Uses ConfigurationProperties for structured config of Kafka topics in alert service
- Fixes an incorrect port forward for Kafka in Tilt
- Use record key to get ordered events instead of relying on a single partition for monitoring-policies topic

## Jira link(s)
- https://issues.opennms.org/browse/HS-1591
- https://issues.opennms.org/browse/HS-1595

## Flagged for review

Changes to consumer group ids will cause the broker to see them as new consumers and add an offset. `auto-offset-reset: earliest` would cause these consumers to reprocess the whole queue, but was unused in most services as it wasn't being passed to the KafkaTemplate.

Not sure about structure of KafkaTopicProperties - I wanted to separate the values used for producers/consumers from the values used for setting up the NewTopic beans. It's redundant as a result, looking for opinions.

Monitoring policy topic uses a record ID instead of relying on a single partition now. If we're using this pattern instead of domain events it would be good to make the topic compacted and produce tombstone records when policies are deleted. This is out of scope of this PR.

## Checklist


BenD10 (approved):
LGTM!

alert/src/main/resources/application.yaml
```diff
@@ -49,7 +44,12 @@ grpc:
...
-    monitoring-policy: "monitoring-policy"
+    event: "events"
     tag-operation: "tag-operation"
+    alert: "alerts"
+    monitoring-policy: "monitoring-policy"
+    create-topics:
+      alert:
+        name: "alerts"
+      monitoring-policy:
+        name: "monitoring-policy"
```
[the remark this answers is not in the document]
geraldhumphries: I set them to 10 and used their IDs as record keys, check the "flagged for review" part of the PR for an explanation. Wasn't aware of the refactor!

events/main/src/main/resources/application.yml
```diff
@@ -9,12 +9,11 @@ spring:
     change-log: classpath:db/changelog/changelog.xml
 
   kafka:
+    bootstrap-servers: localhost:59092
     consumer:
-      group-id: "traps-consumer"
-      auto-offset-reset: earliest
+      group-id: "event-service"
```
cgorantla, line 14: This is really a traps consumer, events are not generated yet.

cgorantla (commented):
overall, LGTM!
Might require validation of some of the features to make sure nothing is broken.

events/main/src/main/resources/application.yml
cgorantla: yeah, may need some updates on the configuration file here.

There are three topics we consume from.
```
kafka:
  traps-topic: "traps"
  events-topic: "events"
  internal-topic: "internal-event"
```

Consumer group is missing on each of these topics and these looks to be separated from spring config.

geraldhumphries: Yeah, Spring's config only lets you set up a "default topic" for KafkaListeners which is why these values are separate from the spring config. The group ID works the same, right now it is set in spring's config as `spring.kafka.consumer.group-id` and it gets used for each KafkaListener/topic. 

I didn't think it was necessary, but if we want to have different groupIDs per topic we'd need to do something like this:

```
@KafkaListener(topics = "${kafka.topics.task-set-results.name}", groupId = "${kafka.topics.task-set-results.group-id}")
```

BenD10 (approved):

PR #1112, opened 2023-05-09
HS-1582 Remove cert manager and generate certificates.

splatch:
Certificate generation happens via Helm (TLS.EnableGeneration) with appropriate retain policy or via script which is shared between dev and test phases. Tilt is just an orchestration tool to call scripts. 

## Description
Getting rid of cert-manager which turns to not be required for dev/testing workloads. Attempted to capture scripted part into small / re-usable part.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1582

## Checklist


install-local/generate-and-sign-certificate.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+directory=$(dirname $caKeyFile)
+id=$(uuidgen)
+
+openssl req -newkey rsa:2048 -nodes -keyout "$directory/$id.key" -subj "/CN=$domain/O=Test/C=US" -out "$directory/$id.csr"
+openssl x509 -req -extfile <(printf "subjectAltName=DNS:$domain") -days 14 -in  "$directory/$id.csr" \
+  -CA $caCrtFile -CAkey $caKeyFile -CAcreateserial -out "$directory/$id.crt" || (echo "Certificate $id signing failed" && exit 1)
+kubectl -n $namespace delete "secrets/$secret" || true
+kubectl -n $namespace create secret tls $secret --cert "$directory/$id.crt" --key "$directory/$id.key"
```
geraldhumphries, line 24: This might be more flexible for Tilt if we separate the deployment of the K8s objects or give it an option to toggle  `--dry-run=client`.

install-local/load-or-generate-secret.sh
```diff
@@ -0,0 +1,26 @@
...
+  openssl genrsa -out $keyFile
+  openssl req -new -x509 -days 14 -key $keyFile -subj "/CN=$domain/O=Test/C=US" -out $crtFile
+fi
+
+kubectl -n $namespace delete "secrets/$secret" 2>&1 >/dev/null || true
+# create secret which is generic instead of tls, by default tls secrets have no "ca.crt" field which is mandatory
+# in case if we wish to use given secret for mtls purposes. Since it makes to harm to other cases, make it default
+kubectl -n $namespace create secret generic $secret --from-file=tls.crt=$keyFile \
+  --from-file=tls.key=$crtFile \
+  --from-file=ca.crt=$crtFile || (echo "Could not create $secret in namespace $namespace" && exit 1)
```
geraldhumphries, line 26: Same as above, and option to toggle `--dry-run=client` would be helpful.

install-local/generate-and-sign-certificate.sh
splatch: If we optimize for tilt then there will be less logic we will share with install-local.sh. Not that I strive to have a full coverage. Problem is - processing output of shell commands in tilt feels (to me) cumbersome. How would we need to return data from command to make it useful for tilt?

geraldhumphries: It is a bit awkward, but that's how Tilt's k8s_custom_deploy works: https://docs.tilt.dev/api.html#api.k8s_custom_deploy

On second thought I don't think it's worth doing now.

geraldhumphries (approved):

PR #1115, opened 2023-05-11
HS-1247 NetflowV5 doesn't have  tag

FrancescoBuc86:
https://opennms.atlassian.net/browse/HS-1247

### DIFF
--- a/metrics-processor/flows/src/main/java/org/opennms/horizon/flows/processing/DocumentEnricherImpl.java
+++ b/metrics-processor/flows/src/main/java/org/opennms/horizon/flows/processing/DocumentEnricherImpl.java
@@ -28,24 +28,6 @@
 
 package org.opennms.horizon.flows.processing;
 
-import com.codahale.metrics.MetricRegistry;
-import com.google.protobuf.TextFormat;
-import com.google.protobuf.UInt64Value;
-import io.grpc.Status;
-import io.grpc.StatusRuntimeException;
-import org.opennms.dataplatform.flows.document.FlowDocument;
-import org.opennms.dataplatform.flows.document.Locality;
-import org.opennms.dataplatform.flows.document.NodeInfo;
-import org.opennms.horizon.flows.classification.ClassificationEngine;
-import org.opennms.horizon.flows.classification.ClassificationRequest;
-import org.opennms.horizon.flows.classification.persistence.api.Protocols;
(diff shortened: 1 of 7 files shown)


metrics-processor/flows/src/main/java/org/opennms/horizon/flows/processing/DocumentEnricherImpl.java
```diff
@@ -82,6 +84,13 @@ public List<FlowDocument> enrich(final Collection<org.opennms.dataplatform.flows
 
         return flows.stream().flatMap(flow -> {
             final var document = FlowDocument.newBuilder(flow);
+            try {
+                TextFormat.merge(flow.toString(), document);
```
freddy61025, line 88: I believe it is will not solve the problem.
May be the problem is because we should not use 0 for V5. Where 0 is always default value for protobuf. So when persist to ingester flowDoc.toString() it will skip that value due to protobuf think value is not set yet.

FYI:
FlowDocument.class line: 2478 (FlowDocument.newBuilder(flow))
                if (other.netflowVersion_ != 0) {
                    this.setNetflowVersionValue(other.getNetflowVersionValue());
                }

In protobuf.

enum NetflowVersion {
    V5 = 0;
    V9 = 1;
    IPFIX = 2;
    SFLOW = 3;
}

FrancescoBuc86: good point

FrancescoBuc86: But if this is the problem, why netflowVersion is missing for V9 and IPFix?

shared-lib/flows/src/main/proto/flow_document.proto
```diff
@@ -9,8 +9,9 @@ import "google/protobuf/wrappers.proto";
 
 
 enum Direction {
-    INGRESS = 0;
-    EGRESS = 1;
+    NONE = 0;
```
freddy61025, line 12: I think it is better to use the same word either all NONE or UNDEFINED.

freddy61025 (changes requested):

freddy61025:
Make sure to notify DevOps and testers that existing data will be wrong once it is merged.

freddy61025 (approved):

PR #1131, opened 2023-05-15
HS-1460: minion sends messages without location; minion gateway maps …

artnaseef:
…them to messages with location (and Tenant).

## Jira link(s)
- https://issues.opennms.org/browse/HS-1460

## Checklist


metrics-processor/flows/src/main/java/org/opennms/horizon/flows/integration/FlowRepositoryImpl.java
```diff
@@ -19,18 +23,104 @@ public class FlowRepositoryImpl implements FlowRepository {
...
+        );
+    }
+
+//========================================
+// Internals
+//----------------------------------------
+
+    // NOTE: this will hopefully be simplified in the future
+    private FlowDocument mapFlowDocument(TenantLocationSpecificFlowDocument tenantLocationSpecificFlowDocument) {
+        FlowDocument result =
```
splatch, line 53: Haven't tested below, but I think it might get you implicit mapping of fields having same name:
```
        Map<FieldDescriptor, Object> sourceFields = tenantLocationSpecificFlowDocument.getAllFields();
        Builder builder = FlowDocument.newBuilder();
        Map<FieldDescriptor, Object> targetFields = builder.getAllFields();

        for (Entry<FieldDescriptor, Object> entry : sourceFields.entrySet()) {
            FieldDescriptor sourceField = entry.getKey();

            FieldDescriptor targetDescriptor = targetFields.keySet().stream().filter(targetFiled -> targetFiled.getName().equals(sourceField.getName()))
                .findFirst().orElse(null);

            if (targetDescriptor != null) {
                Object value = entry.getValue();
                builder.setField(targetDescriptor, value);
            }
        }
        return builder.build();
```

metrics-processor/flows/src/main/java/org/opennms/horizon/flows/processing/impl/FlowDocumentClassificationRequestMapperImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import org.opennms.horizon.flows.document.TenantLocationSpecificFlowDocument;
+import org.opennms.horizon.flows.processing.FlowDocumentClassificationRequestMapper;
+
+import java.util.function.Function;
+
+public class FlowDocumentClassificationRequestMapperImpl implements FlowDocumentClassificationRequestMapper {
+
+    // Testability (TODO: stop using static methods in Protocols and replace this Op with an injected instance)
+    @Setter
+    private Function<Integer, Protocol> protocolLookupOp = Protocols::getProtocol;
```
splatch, line 16: Nit pick - declaring a basic interface such as `ProtocolLookup` (which could still be a functional interface) , gives us a chance to search for usages. With above construction we're forced to look either for `Function` or `Protocol` cause IDEs are not smart enough to lookup `Function` with its type boundaries. ;(

minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/grpc/common/LocationServerInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.function.Supplier;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.opennms.horizon.shared.constants.GrpcConstants;
+
+/**
+ * Location resolver which rely on grpc header.
+ */
+@Slf4j
+public class LocationServerInterceptor implements ServerInterceptor {
```
splatch, line 49: Small thought, maybe its worth moving this class to minion gateway itself to avoid desire to re-use it elsewhere.

splatch (approved):
I hope I have not missed anything since github had problem loading PR at once. I made several comments, however they are rather generic. I see no major issues getting it merged. One note - it would be worth to confirm flows/node mapping which got updated with flows team given limited test coverage we had in this area.

PR #1156, opened 2023-05-23
Basic application tracing support for minion-gateway

deejgregor:
This adds some very basic, but fairly useful, application tracing enhancements to the minion-gateway:

- Tracing: Set the "user" attribute to the tenant ID, if any -- Whenever code calls a method on `TenantIDGrpcServerInterceptor`, the tenantId is added as a span attribute to the current span (if any) with the key `user`. With this, you can do a tag search in a tracing tool to find traces for a specific tenantId, e.g.: `user=opennms-prime`.
- minion-gateway: Start a new span when we call dispatch -- The connections from the minion client to the minion-gateway are long lived and handle many requests, so the lifetime of the trace for the incoming connection is too long to be reasonable to directly contain the traces for all requests. This creates a new top-level trace every time we call `dispatch`, linking it to the connection-level trace, which makes the individual traces more consumable. Note that the span records for the long-lived connection trace won't be emitted until the connection is over, so the if you are looking at recent individual traces, it will often be the case that the parent trace hasn't been seen yet.

One big caveat: due to outstanding OOM issues in the `minion-gateway`, tracing is disabled in the dev environment for now. It is enabled by default in `Tilt`, however.

## Jira link(s)

## Checklist

### DIFF
--- a/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/grpc/common/TenantIDGrpcServerInterceptor.java
+++ b/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/grpc/common/TenantIDGrpcServerInterceptor.java
@@ -38,6 +38,8 @@ import io.grpc.ServerInterceptor;
 import io.grpc.Status;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
 
 import java.time.Duration;
 import java.util.function.Supplier;
@@ -94,6 +96,11 @@ public class TenantIDGrpcServerInterceptor implements ServerInterceptor {
 //----------------------------------------
 
     private String commonReadContextTenantId(Supplier<String> readTenantIdOp) {
-        return readTenantIdOp.get();
+        var tenantId = readTenantIdOp.get();
+        var span = Span.current();
+        if (span.isRecording()) {
+            span.setAttribute("user", tenantId);
+        }
+        return tenantId;
     }
 }
--- a/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/ipc/grpc/server/OpennmsGrpcServer.java
+++ b/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/ipc/grpc/server/OpennmsGrpcServer.java
@@ -35,7 +35,12 @@ import com.google.protobuf.Message;
 import io.grpc.BindableService;
 import io.grpc.Context;
 import io.grpc.stub.StreamObserver;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.trace.Tracer;
 import java.io.IOException;
+import java.lang.reflect.UndeclaredThrowableException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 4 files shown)


minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/ipc/grpc/server/OpennmsGrpcServer.java
```diff
@@ -122,6 +127,8 @@ public class OpennmsGrpcServer extends AbstractMessageConsumerManager implements
     private BiConsumer<RpcRequestProto, StreamObserver<RpcResponseProto>> incomingRpcHandler;
     private OutgoingMessageHandler outgoingMessageHandler;
 
+    private Tracer tracer = GlobalOpenTelemetry.get().getTracer(getClass().getName());
```
splatch, line 130: How does this line will behave with disabled otel?

deejgregor: It will get a [no-op implementation](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L83). If the otel javaagent isn't loaded or if it is disabled, `io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk` won't be on the classpath so `maybeAutoConfigureAndSetGlobal` will [return early](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L224).

minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/grpc/common/TenantIDGrpcServerInterceptor.java
```diff
@@ -94,6 +96,11 @@ public String readContextTenantId(Context context) {
 //----------------------------------------
 
     private String commonReadContextTenantId(Supplier<String> readTenantIdOp) {
-        return readTenantIdOp.get();
+        var tenantId = readTenantIdOp.get();
+        var span = Span.current();
+        if (span.isRecording()) {
```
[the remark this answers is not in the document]
deejgregor: > Does otel track grpc headers out of the box?

When using the javaagent (which is how I have it setup), yes, the [gRPC instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/grpc-1.6) instruments grpc calls and propagates the context in metadata.

> If so, then tenant-id should be recorded earlier, before span is being made.

It depends on *which* span we want it in. Here, we aren't creating a span, we are just adding the attribute to the existing span. I think it happens to put it in the right spot with [the other change that starts a new root span before the message is dispatched in the minion-gateway[(https://github.com/OpenNMS-Cloud/horizon-stream/pull/1156/commits/7d1228fb5279a2b8867c04267e610f0aa8c773b8). See the highlight in the attached image:

image

And, to a certain extent, it doesn't matter all that much because in a tracing tool (at least the two I've used), when you search by attributes, it doesn't care which span matches--it will match on *any* span in the trace (unless you are filtering on a span name, I guess). So as long as you have the attribute you care about on one of the spans in a trace, you'll find it.

> If its not recorded then we probably need also to record location information which comes in same way.

Two things for consideration:
1. Is the location possibly sensitive information? If so, we might not want to include it by default (or hash it or something). This is something we could always add now and change later, or vice-versa--it's only a line or two, anyway, and it's a trivial change.
2. This might be easy to do through configuration, e.g.: 
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/grpc-1.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_6/GrpcSingletons.java#L35 (unless we are picky about the attribute name). I haven't tried this, but it looks like it would do the right thing. There is one downside: this would be on the long-running span that is related to the connection, I think, so if we want it on the per-message traces, we might still need to toss some code at it.

PR #1160, opened 2023-05-23
HS-1519 Remove autoamtic creation of location in inventory

splatch:
## Description
This is an updated version of #1068 made by Benjamin, but applied against current develop. I keep old branch and PR as a reference point in case of issues with this change set.

## Jira link(s)
- https://issues.opennms.org/browse/HS-1519
- 
## Checklist

### DIFF
--- a/inventory/main/src/main/java/org/opennms/horizon/inventory/service/NodeService.java
+++ b/inventory/main/src/main/java/org/opennms/horizon/inventory/service/NodeService.java
@@ -152,8 +152,7 @@ public class NodeService {
     }
 
     private MonitoringLocation findMonitoringLocation(NodeCreateDTO request, String tenantId) throws LocationNotFoundException {
-        String location = locationNameOrDefault(request);
-        Optional<MonitoringLocation> found = monitoringLocationRepository.findByLocationAndTenantId(location, tenantId);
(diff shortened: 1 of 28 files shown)


inventory/main/src/main/java/org/opennms/horizon/inventory/service/NodeService.java
```diff
@@ -318,4 +306,8 @@ public List<NodeDTO> listNodesByTags(String tenantId, List<String> tags) {
             .map(mapper::modelToDTO).toList();
     }
 
+    private static String locationNameOrDefault(NodeCreateDTO request) {
```
cgorantla, line 309: we don't even create default location, so this may not be needed.

cgorantla (commented):
Looks good to me overall.
One comment on using default location

inventory/main/src/main/java/org/opennms/horizon/inventory/service/NodeService.java
splatch: I was asking myself whether it is needed or not, some tests present in inventory failed if this logic was absent. I'll update code to have explicit declaration of location.

cgorantla (approved):

PR #1172, opened 2023-05-30
HS-1669: force plaintext connection for flows

freddy61025:
## Jira link(s)
- https://issues.opennms.org/browse/HS-1669

## Checklist

### DIFF
--- a/charts/opennms/values.yaml
+++ b/charts/opennms/values.yaml
@@ -29,6 +29,7 @@ OpenNMS:
         Memory: 1Gi
     PrivateRepoEnabled: false
     FlowsUrl: flows.onms-dp-dev.dev.nonprod.dataservice.opennms.com:443
+    FlowsTlsEnabled: false
     CertificateManagerUrl: dev
   MetricsProcessor:
     ServiceName: opennms-metrics-processor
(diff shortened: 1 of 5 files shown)


charts/opennms/values.yaml
```diff
@@ -44,6 +45,7 @@ OpenNMS:
         Memory: 256Mi
     PrivateRepoEnabled: false
     FlowIngestorUrl: flows.onms-dp-dev.dev.nonprod.dataservice.opennms.com:443
+    FlowTlsEnabled: false
```
artnaseef, line 48: The default value to TLS enabled is `false` and the port is `443`?  Shouldn't it be default of `true` for TLS enabled?  Since the original code required TLS, we can retain backwards compatibility that way.

If we really want the default to be false, then the default port should be updated to the default port number we use for plain-text connections as well.

splatch (approved):
LGTM!

charts/opennms/values.yaml
```diff
@@ -29,6 +29,7 @@ OpenNMS:
         Memory: 1Gi
     PrivateRepoEnabled: false
     FlowsUrl: flows.onms-dp-dev.dev.nonprod.dataservice.opennms.com:443
+    FlowsTlsEnabled: false
```
artnaseef, line 32: Same comment on TLS enabled as below...

charts/opennms/values.yaml
freddy61025: that url is no longer valid, each cloud platform will have their own url during deployment.

artnaseef (approved):

PR #1190, opened 2023-06-03
Add a new GitHub workflow to test 'tilt ci'

deejgregor:
## Jira link(s)
- https://issues.opennms.org/browse/BTO-509

## Checklist
* [ ] Follows Horizon Stream's [development guidelines.](https://github.com/OpenNMS/lokahi/wiki/Development-Guidelines)
* [ ] Includes [appropriate tests.](https://github.com/OpenNMS/lokahi/wiki/Test-Strategy)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,7 @@
 <!-- Describe this Pull Request, what it changes, and why it's necessary. -->
 
 ## Jira link(s)
-- https://issues.opennms.org/browse/HS-xx
+- https://opennms.atlassian.net/browse/HS-xx
 
 ## Flagged for review
 <!-- Flag things as "needs a close look" for reviewers, if necessary. Include as much detail as possible (line numbers, concerns, and so on). -->
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -21,6 +21,9 @@ on:
         required: false
         type: string
 
+env:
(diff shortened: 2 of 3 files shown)


splatch:
Its good idea to make it part of CI/CD, however second question I have is - whether we do run external-its against tilt setup in order to make sure that everything works?

deejgregor:
> Its good idea to make it part of CI/CD, however second question I have is - whether we do run external-its against tilt setup in order to make sure that everything works?

@splatch it's a good idea! I had the same thought and looked into it a bit, but I ran into two main issues:
1. Tilt (generally?) builds everything itself. I think for running most of our tests, including the external-it tests, we want to run against the docker containers we've built in the `develop.yaml` workflow.
2. The `tilt ci` subcommand does the normal Tilt stuff to bring up the application (including compiling everything), makes sure it is healthy, and then immediately shuts it down. We could possibly do `tilt up` and put it in the background, wait for things to be healthy, run tests, and then `tilt down`, but those subcommands seem to be more suited for an interactive workflow.

I think in this case, it is better to use the tilt CI workflow just to tests the primary Tilt workflow that developers use, and leave the other tests to other workflows. At least for now, until maybe we find a better way to address the issues above. What I **do** hope to do in the somewhat near future is add/improve the probes in our containers, and the status of those probes will be automatically used by everything downstream that looks at the rollout health (Tilt, Argo CD, prometheus alerts, etc.).

PR #1199, opened 2023-06-06
Tilt local env update

jaymberry:
## Jira link(s)

## Checklist
* [ ] Follows Horizon Stream's [development guidelines.](https://github.com/OpenNMS/lokahi/wiki/Development-Guidelines)
* [ ] Includes [appropriate tests.](https://github.com/OpenNMS/lokahi/wiki/Test-Strategy)


Tiltfile
```diff
@@ -410,8 +410,8 @@ k8s_resource(
     'ingress-nginx-controller',
     labels=['0_useful'],
     port_forwards=[
-        port_forward(8123, 80),
-        port_forward(1443, 443),
+        port_forward(8123, 80, host='0.0.0.0'),
```
deejgregor, line 413: For security reasons, we probably don't want to enable this by default, because it exposes externally on all interfaces on the host. It also causes a warning to pop-up every time. I'll see if there's a way to make this optional, or only tie it to internal interfaces.

tests/minion.p12
deejgregor, line 1: It looks like you committed a minion.p12 file.

Tiltfile
deejgregor: I was able to add a command-line option to specify a `host` value, letting us keep this secure by default but making it easy to open up when needed. See: https://github.com/OpenNMS-Cloud/lokahi/pull/1200/commits/0df98e80263e94b46e2a857384fe8657ec1626a6

Interestingly, on my M1 Mac, I **did not** need to open up the host access to be able to test the docker compose minion with tilt.

deejgregor:
See https://github.com/OpenNMS-Cloud/lokahi/pull/1200, and feel free to approve and merge into your branch or cherry-pick as appropriate.

tests/minion.p12
deejgregor: I have some cleanups in https://github.com/OpenNMS-Cloud/lokahi/pull/1200/commits/552c27ee1b64571a8866674f654eb77f4a388c78 that includes deleting the `minion.p12` file and adding `minion.p12` to `.gitignore`.

deejgregor (approved):

deejgregor (approved):

PR #1211, opened 2023-06-09
Tilt: fix adding test location to database

deejgregor:
This is best reviewed commit-by-commit because there's one main change, plus a number of things cleaned-up/improved along the way. This is one ancillary change that I wanted to highlight which is bolded.

The main change: [Tilt: fix adding test location to database](https://github.com/OpenNMS-Cloud/lokahi/pull/1211/commits/bc38390cc4b8bd07c366f1f4d58533144f2746d8) 

Follow the pattern in tests/minion-cert.sh for adding a location
(it it doesn't already exist) in the initContainer for the minion
that sets up certificates.

Added error checking so the init container won't successfully
finish until the location exists and the certificate is created.
In my experience, it takes 2-3 passes until everything is up, but
it does eventually complete.

Ancillary changes:
* Always return server-timing header, even when there are errors 
* Add default-container annotation for ingress-nginx controller pod
* Tilt: minion initContainer: display traceparent if there's an error w… 
* OpenTelemetry workarounds for gRPC and auth service propagation
* **======= LOOK AT THIS ONE: [Tilt: add tenant-id, location, and gRPC status to ingress-nginx access logs](https://github.com/OpenNMS-Cloud/lokahi/pull/1211/commits/09be0125bfeb2ec3a00d6cde1828d10c72676558) =======**
* Remove duplicate ingress annotations
* minion-gateway sink streaming: Reduce exception noise when the client… 
* install-local.sh: fix name of lokahi-minion-gateway image in pull_doc… 

Here is an example with the logging change in Tilt (which will be expanded later if this works well):
```
$ kubectl logs ingress-nginx-controller-5d5ccb976-96hrk| grep TestLocation
10.244.0.59 - - [12/Jun/2023:02:10:37 +0000] "POST /minion.CloudService/MinionToCloudRPC HTTP/2.0" 200 2689 "-" "grpc-java-okhttp/1.54.1" 323 0.968 [default-opennms-minion-gateway-8990] [] 10.244.0.56:8990 2722 0.146 200 da99bab43b128a3d09c12cc15921bcf1 opennms-prime TestLocation - 0
```
- tenant-id: `opennms-prime`
- location: `TestLocation`
- GRPC status codes: `- 0`

Here is the trace for that same transaction, showing the details of the certificate verification (this was made possible with both the workaround for passing the trace context to the auth subrequest and for passing the trace context downstream in gRPC requests from nginx):
image

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-xx

## Checklist

### DIFF
--- a/tilt-helm-values.yaml
+++ b/tilt-helm-values.yaml
@@ -78,7 +78,7 @@ OpenNMS:
           - -c
           - |
             set -euo pipefail
-            trap 's=$?; echo >&2 "$0: Error on line "$LINENO": $BASH_COMMAND"; exit $s' ERR
+            trap 's=$?; echo >&2 "$0: Error on line "$LINENO": $BASH_COMMAND"; if [ -f /tmp/curl-headers.txt ]; then grep server-timing: /tmp/curl-headers.txt; fi; exit $s' ERR
 
             url="https://onmshs.local:1443"
             set -- --connect-to onmshs.local:1443:ingress-nginx-controller.default.svc.cluster.local:443
@@ -87,6 +87,7 @@ OpenNMS:
             graphql="${url}/"
 
             token=$(curl "$@" -sSf -k -X POST -H "Content-Type\\: application/x-www-form-urlencoded" \
+              --dump-header /tmp/curl-headers.txt \
               -d 'username=admin' \
               -d 'password=admin' \
               -d 'grant_type=password' \
@@ -98,6 +99,7 @@ OpenNMS:
             query="{\"query\": \"query { searchLocation(searchTerm: \\\"LOC_NAME\\\") { id, location } }\"}"
             query=${query/LOC_NAME/$locationName}
             result=$(curl "$@" -sSf -k -H "Content-Type: application/json" \
+              --dump-header /tmp/curl-headers.txt \
               -H "Authorization: Bearer $token" \
               --data "${query}" \
               "${graphql}api/graphql")
@@ -115,6 +117,7 @@ OpenNMS:
               query="{\"query\": \"mutation { createLocation(location: {location: \\\"LOC_NAME\\\"}) { id, location } }\"}"
               query=${query/LOC_NAME/$locationName}
               result=$(curl "$@" -sSf -k -H "Content-Type: application/json" \
+                --dump-header /tmp/curl-headers.txt \
                 -H "Authorization: Bearer $token" \
                 --data "${query}" \
                 "${graphql}api/graphql")
--- a/tilt-helm-values.yaml
+++ b/tilt-helm-values.yaml
@@ -13,9 +13,6 @@ Keycloak:
   HostnamePort: 1443
   HostnameAdminUrl: https://onmshs.local:1443/auth
 
-Postgres:
-  addTestLocation: True
-
 Ingress:
   Enabled: false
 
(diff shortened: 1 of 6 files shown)


tilt-helm-values.yaml
```diff
@@ -114,16 +166,21 @@ OpenNMS:
...
+        # could figure out how to get the tenant and minion information in the logs using $sent_http_x_tenant_id
+        # and $sent_http_x_location once these response headers are added. They also seem to need to be in the
+        # location block to work.
+        # Note: The response headers are prefixed with "x-", although the headers from the auth service are not.
+        # Note: I am being a little naughty and depending an internal implementation detail of ingress-nginx by
+        # using the authHeader* variables. These are created by ingress-nginx from the auth-response-headers
+        # annotation. The naming of the authHeader* variables is sensitive to the order of the headers in the
+        # nginx.ingress.kubernetes.io/auth-response-headers annotation.
+        add_header x-tenant-id "$authHeader0" always;
+        add_header x-location "$authHeader1" always;
```
splatch, line 181: I am not very familiar with nginx, beyond all exercises I needed for this project, yet I am not sure if we need output headers for that. The `nginx.ingress.kubernetes.io/auth-response-headers` automatically maps `tenant-id` and `location` under the neath, without the need to use `$authHeaderN`, it also sets up `$upstream_http_tenant_id` variable:
```
			auth_request_set $authHeader0 $upstream_http_tenant_id;
			grpc_set_header 'tenant-id' $authHeader0;
			auth_request_set $authHeader1 $upstream_http_location;
			grpc_set_header 'location' $authHeader1;
```
Can we rely on `$upstream_http_location`? Regardless of this I'd consider getting the `more_clear_header` or `proxy_hide_header` just to avoid people trying to use these headers on client side.

splatch (changes requested):
I'm a bit suspicious on making extra headers flying to clients thus I would like to clarify that point, and ideally remove them from http response to avoid people using these, even if its development level setting.

tilt-helm-values.yaml
deejgregor: > The nginx.ingress.kubernetes.io/auth-response-headers automatically maps tenant-id and location under the neath, without the need to use $authHeaderN, it also sets up $upstream_http_tenant_id variable

The `auth-response-headers` annotation doesn’t setup the `$upstream_http_tenant_id` variable—the [`auth_request_set`](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html#auth_request_set) command **consumes** it to create a new variable, `$authHeader0`.

This is a challenge, I think, because the data we want to log is part of the **auth sub-request response**, and those don’t seem to be available with `$upstream_*` outside of the location that block that handles the sub-request. I couldn’t find a way to access headers from the **proxied request**. And I couldn’t use `$authHeader*` directly in the log format because it’s not available globally—only on authenticated minion requests.

I tried a **bunch** of permutations and this was the only one that I got working. :( And I didn’t find any examples of anyone else trying to do the same type of thing, unfortunately.

There probably is another, better way to do this, but I haven’t found it yet—maybe with a hack with a map to deal with the case when $authHeader* is unset, or there is a way to access proxied request headers, or toss some Lua code at it. But for now, I found this extremely useful, and it’s only in Tilt for now (and we can decide if/when we want to deploy it in our cloud environments).

splatch: Fair, I do understand implications, your reasoning is perfectly fine. Knowing a bit of ingress-nginx I'd rather stick with what you did rather than look for perfect solution which might not exist. 
I do think that long term solution should go towards use of secrets, instead of init containers, however this might be deferred.

splatch (approved):

PR #1215, opened 2023-06-13
HS-1037 Tweak minion gateway defaults so it works with multiple replicas

splatch:
## Description
Small adjustments to keep minion-gatewy working with multiple replicas.

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1037

## Checklist


artnaseef (approved):
LGTM

deejgregor (approved):

charts/lokahi/templates/opennms/minion/minion-gateway-deployment.yaml
```diff
@@ -70,6 +70,12 @@ spec:
                   key: minionGatewayPwd
             - name: SPRING_DATASOURCE_URL
               value: "jdbc:postgresql://{{ .Values.OpenNMS.MinionGateway.DbHost }}:5432/{{ .Values.OpenNMS.MinionGateway.DbName }}?currentSchema={{ .Values.OpenNMS.MinionGateway.DbSchemaName }}"
+            - name: IGNITE_USE_KUBERNETES
```
deejgregor, line 73: We use `USE_KUBERNETES` for what I think is a similar thing in the Minion. I think we should probably try to use similar naming, if we can.

```
minion/assembly/src/main/resources/etc/org.opennms.horizon.minion.ignite.worker.cfg:useKubernetes=${env:USE_KUBERNETES:-true}
```

I don't think we've thought terribly hard about the naming on any of this previously, so I'm kinda wondering if we should change the Minion to follow the naming scheme here. :-) What do you think?

deejgregor (approved):
I had one comment, but I think it's more of something where we might want to change the Minion, not this side.

deejgregor:
Looks good to me. Rebased.

deejgregor (approved):
Adding another review because GitHub is being weird and keeps sending me notifications that I need to review this PR. Calm down, GitHub! :)

charts/lokahi/templates/opennms/minion/minion-gateway-deployment.yaml
splatch: Minion side is a bit different animal, even if it uses same mechanisms. On minion side we do map env vars manually, on minion gateway we match them to spring application properties. Minion vars are intended to be short, to make them easier to override.
On gateway side - spring is able to map env vars to known properties it is possible to apply overrides as needed. However this comes at the cost of being more explicit in how we declare properties/vars. Reason why we have ignite prefix is because spring context is configured through ignite.use_kubernetes property (AFAIR). If we leave use_kubernetes alone, it might mean a lot of different things.

PR #1225, opened 2023-06-16
HS-1631: Add minion instructions to locations page

mikewrosey:
## Description
Adds minion instructions (docker cmd) to locations page. The host, password, minion id, and filename will be dynamically replaced once the user downloads the certificate. User then needs to update only the path to the cert & run.

screenshot-localhost_3009-2023 06 16-12_51_24

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1631

## Checklist


ui/src/composables/useMinionCmd.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const url = computed(() =>
+    location.origin.includes('dev')
+      ? 'minion.onms-fb-dev.dev.nonprod.dataservice.opennms.com'
+      : 'minion.onms-fb-prod.production.prod.dataservice.opennms.com'
+  )
+
+  const minionDockerCmd = computed(
+    () =>
+      `docker run --rm -p 8181:8181 -p 8101:8101 -p 1162:1162/udp -p 8877:8877/udp -p 4729:4729/udp -p 9999:9999/udp -p 162:162/udp -e TZ='America/New_York' -e USE_KUBERNETES="false" -e MINION_GATEWAY_HOST="${url.value}" -e MINION_GATEWAY_PORT=443 -e MINION_GATEWAY_TLS="true" -e GRPC_CLIENT_KEYSTORE='/opt/karaf/minion.p12' -e GRPC_CLIENT_KEYSTORE_PASSWORD='${password.value}'  -e MINION_ID='${minionId.value}'  --mount type=bind,source="/pathToFile/${minionId.value}-certificate.p12",target="/opt/karaf/minion.p12",readonly opennms/lokahi-minion:latest`
```
deejgregor, line 20: The ports on this aren't quite right (in particular, I think `-p 1162:1162/udp` and `-p 162:162/udp` are wrong and it should be `-p 162:1162/udp`--*I think*). I unfortunately don't know what the proper list is. :(

Also, every single environment variable that is static in the list should probably be a default instead in the container. If we are telling **everybody** to always set them by default, we should instead fix the defaults. ;-)

@jaymberry do you think we can set these environment variable defaults for the minion?
- USE_KUBERNETES="false"
- MINION_GATEWAY_PORT=443
- MINION_GATEWAY_TLS="true"
- GRPC_CLIENT_KEYSTORE='/opt/karaf/minion.p12'

Personally, I wouldn't set a TZ by default in the container (except UTC) because using timezones in production environments of any size sucks because you never have everything in one time zone (hence why I always use UTC).

deejgregor (changes requested):
I'm **super** glad we're adding this, but we need to tighten up the example command a bit more.

ui/src/composables/useMinionCmd.ts
```diff
@@ -0,0 +1,26 @@
...
+    minionId.value = 'default'
+  }
+
+  const url = computed(() =>
+    location.origin.includes('dev')
+      ? 'minion.onms-fb-dev.dev.nonprod.dataservice.opennms.com'
+      : 'minion.onms-fb-prod.production.prod.dataservice.opennms.com'
+  )
+
+  const minionDockerCmd = computed(
```
synqotik, line 18: This might be easier to read as an array of strings, then joining on ' '.

synqotik (approved):
Vue parts look good, just one suggestion which can be ignored.

I don't know about the ports and env variables etc. so DJ should approve that part.

ui/src/composables/useMinionCmd.ts
mikewrosey: Thanks DJ, maybe @jbeuree or @jaymberry can weigh in about the ports. It's an easy change, but just want to be sure.

mikewrosey: @deejgregor after asking around & posting in the main channel, it appears nobody seems to know about the ports for sure. It did work when I tested it, but I expect that this cmd will go through several changes in the future. Do you think it would be OK to merge as is, and update as needed? Asking because this composable is to be reused for other features we're working on, like the onboarding.

mikewrosey: @maximbrener tagging you in on this

freddy61025: minion gateway port for cloud is always be 443. Do we need to make it as default during package docker ?

freddy61025: the above command is a bit long but the ports looks correct.

agalue: Even if the port list is correct, here are a few recommendations.
Take a look that the following:
https://github.com/OpenNMS-Cloud/lokahi/blob/develop/inventory/main/src/main/resources/flows-config.json
From the above, you can see that port 9999 has a multi-protocol listener, meaning you can send Netflow 5, Netflow 9, and IPFIX flows through it so that we can remove the dedicated ports (8877 and 4729). Also, if we have 1162 for Traps, what's the reason for 162? That's the official SNMP Trap port (as the standard SNMP port is 161).

mikewrosey: Thanks for all the suggestions. I will trim the ports down for now and get this in to unblock the onboarding page. Will make further changes as necessary.

PR #1232, opened 2023-06-20
HS-1742: add prepareLocationAndCerts.sh script for use with runInPlace.sh

artnaseef:
## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1742

## Checklist
* [x] Includes [appropriate tests.](https://github.com/OpenNMS-Cloud/lokahi/wiki/Test-Strategy)

### DIFF
--- a/minion/assembly/prepareLocationAndCerts.sh
+++ b/minion/assembly/prepareLocationAndCerts.sh
@@ -1,0 +1,407 @@
+#!/bin/bash
+######################################################################################################################
+##
+## DESCRIPTION:
+##	Prepare for running a minion by logging into the cluster, looking up the ID of the location name requested,
+##	and creating the location if is does not yet exist, then downloading the PKCS12 Minion certficiate.
+##
+## PROCESS:
+##	1. Extract the CA certificate from the cluster and save as the trust-store for the Minion (and curl commands)
+##	2. Login to the cluster
+##	3. Lookup the location
+##	4. If the location does not yet exist, create the location
+##	5. Download the location certificate and password
+##
+## EXAMPLE USAGE:
+##	./prepareLocationAndCerts.sh -l LOC001 -u admin -p admin
+##
+######################################################################################################################
+
+
+TEMPLATE_LOOKUP_LOCATION_GQL='
+	query { locationByName(locationName: "%s") { id } }
+	'
+
+TEMPLATE_CREATE_LOCATION_GQL='
+	mutation createLocation { createLocation(location: { location: "%s" }) { id } }
+	'
+
+TEMPLATE_GET_CERTIFICATE_GQL='
+	query { getMinionCertificate(locationId: %d) { certificate, password } }
+'
+
+
+###
+### SCRIPT INPUTS
+###
+
+CERT_ROOTDIR="$(pwd)/target"
+CLIENT_KEYSTORE="${CERT_ROOTDIR}/minion.p12"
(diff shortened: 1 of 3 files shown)


minion/assembly/prepareLocationAndCerts.sh
```diff
@@ -0,0 +1,407 @@
...
+##	4. If the location does not yet exist, create the location
+##	5. Download the location certificate and password
+##
+## EXAMPLE USAGE:
+##	./prepareLocationAndCerts.sh -l LOC001 -u admin -p admin
+##
+######################################################################################################################
+
+
+TEMPLATE_LOOKUP_LOCATION_GQL='
```
splatch, line 21: We do have very similar script in charts/lokahi/scripts/opennms/minion. While it is not as feature rich as this one, it does pretty much same thing.

minion/assembly/prepareLocationAndCerts.sh
```diff
@@ -0,0 +1,407 @@
...
+	query { locationByName(locationName: "%s") { id } }
+	'
+
+TEMPLATE_CREATE_LOCATION_GQL='
+	mutation createLocation { createLocation(location: { location: "%s" }) { id } }
```
splatch, line 26: Nit pick - mutation does not have to specify a name.

minion/assembly/prepareLocationAndCerts.sh
```diff
@@ -0,0 +1,407 @@
...
+	typeset gql_response
+	typeset report_errors_flag
+
+	typeset err_msg
+
+	gql_response="$1"
+	report_errors_flag="${2:-false}"
+
+	# Check for an error message
+	err_msg="$(echo "${gql_response}" | jq 'select(.errors) | .errors')"
```
splatch, line 127: While `jq` is common, it is not standard and might not be available on all developer boxes. If we keep going with it we need to update developer prerequisites for developers to install it.

minion/assembly/prepareLocationAndCerts.sh
```diff
@@ -0,0 +1,407 @@
...
+# STOP on errors
+set -e
+
+parse_command_line "$@"
+
+get_ca_cert_from_k8s
+
+login
+
+if lookup_location "${LOCATION_NAME}"
```
splatch, line 394: Can this condition be simplified to do not have else part?

splatch (approved):
Its good work, however in order to make this script useful it would be good to clarify what to do with other script we have in helm and if this particular script can be integrated with i.e.`runInPlace.sh`.

minion/assembly/prepareLocationAndCerts.sh
artnaseef: Ahh good to know - thanks.

minion/assembly/prepareLocationAndCerts.sh
artnaseef: True.  This script is meant to be helpful and not necessarily solve all cases for all developers, so I will leave it and let others consider how to handle situations where jq might not be viable.

minion/assembly/prepareLocationAndCerts.sh
artnaseef: yeah, but somehow the `!` syntax always rubs me the wrong way.  Probably because some shells have other meaning for `!` which takes precendence.  I juse avoid that character in shell commands - life is just easier that way ;-).

PR #1234, opened 2023-06-21
HS-1627: add certificate revocation when delete location

freddy61025:
## Description
Align with PKI add revoke certificate and validate certificate serial number in certificate manger
BFF delete location will trigger revoke certificate if location successfully removed 

Will followup with https://opennms.atlassian.net/browse/HS-1737

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1626
- https://opennms.atlassian.net/browse/HS-1627

## Checklist


minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
```diff
@@ -0,0 +1,147 @@
...
+ */
+@Component
+public class SerialNumberRepository {
+    private final ObjectMapper mapper = new ObjectMapper();
+    private static final Logger LOG = LoggerFactory.getLogger(SerialNumberRepository.class);
+
+    private final RocksDB db;
+    public SerialNumberRepository(@Value("${grpc.server.db:/app/db}") String rootDir) throws RocksDBException{
+        Objects.requireNonNull(rootDir);
+        LOG.info("Begin to init rocksDB of {}", rootDir);
```
artnaseef, line 60: Grammar nit.  Maybe reword to "Beginning init of rocksDB at {}".

Another thought - I personally use the logging syntax "<message>: <key>=<value>; ..." - goes back to using Splunk.  So in this case, I would personally do something like this - `Beginning init of rocksDB: db-path={}`

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
```diff
@@ -0,0 +1,147 @@
...
+        LOG.info("Begin to close rocketDb");
+        if (db != null) {
+            db.close();
+        }
+        LOG.info("Successfully closed rocketDb");
+    }
+
+    public void addCertificate(String tenantId, String locationId, X509Certificate certificate) throws RocksDBException, IOException {
+        var meta = new CertificateMeta(tenantId, locationId, certificate);
+        db.put(meta.getSerial().getBytes(), mapper.writeValueAsBytes(meta));
```
artnaseef, line 83: Can the key use the tenant-id and location-id so lookups avoid the linear search?  If rocks DB doesn't support a composite key, then we could format something like `TENANT-ID=<tenant>|LOCATION=<location>` - a syntax like that should make conflicting keys almost impossible, and completely infeasible.

Another question - is it possible in rocks DB to have the value be a list?  If not, perhaps we can store the value as JSON text containing a list of the valid certificate serial numbers?

Looking ahead, I see there's a need to be able to lookup the cert just by serial number... I'm going to reach out offline to discuss more.

artnaseef (commented):
Didn't finish; want to catch up offline to discuss...

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
freddy61025: The main target is to make certificate don't contain any custom values in subject. We only use serial number as key.

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * DB structure is key (serial number) value (metadata)
+ */
+@Component
+public class SerialNumberRepository {
+    private final ObjectMapper mapper = new ObjectMapper();
+    private static final Logger LOG = LoggerFactory.getLogger(SerialNumberRepository.class);
+
+    private final RocksDB db;
+    public SerialNumberRepository(@Value("${grpc.server.db:/app/db}") String rootDir) throws RocksDBException{
```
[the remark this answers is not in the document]
freddy61025: I will change it. It just my personal preference to make names more generic. That switching backend will not affect paths.

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
artnaseef: On follow-up, we don't need to optimize this solution, so no changes needed here.

minion-certificate-manager/main/src/test/resources/test-output-dir/client.signed.cert
artnaseef, line 1: Is this file a test input or test output?

If output, then please delete.  I recommend also adding the folder to the `.gitignore` file.  Or, better yet, move the test path to use the `target/` folder instead of `src/test/resources` for test output - tests should never write to any place under `src/`.

rest-server/src/main/java/org/opennms/horizon/server/service/GrpcLocationService.java
```diff
@@ -86,6 +88,11 @@ public Mono<MonitoringLocation> updateLocation(@GraphQLArgument(name = "location
 
     @GraphQLMutation
     public Mono<Boolean> deleteLocation(@GraphQLArgument(name = "id") long id, @GraphQLEnvironment ResolutionEnvironment env) {
-        return Mono.just(client.deleteLocation(id, headerUtil.getAuthHeader(env)));
+        var accessToken = headerUtil.getAuthHeader(env);
+        var status = client.deleteLocation(id, accessToken);
+        if (status) {
```
artnaseef, line 93: It might be good to call revoke even if delete fails, because it is possible the location could get deleted but the certificate doesn't get revoked due to a service crash or other scenario.

artnaseef (changes requested):

minion-certificate-manager/main/src/test/resources/test-output-dir/client.signed.cert
freddy61025: It is a test input. It is for PKCS12GeneratorTest.
PKCS12GeneratorTest itself is mocking the CommandExecutor, openssl is never be called. That said getX509Certificate will never able to find the real cert. The only way i can think of is put a fake file in test output folder. The other way is to spy PKCS12Generator and override getX509Certificate. (However for PKCS12GeneratorTest, I don't want to spy PKCS12Generator)

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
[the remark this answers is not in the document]
freddy61025: done

rest-server/src/main/java/org/opennms/horizon/server/service/GrpcLocationService.java
freddy61025: done

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/SerialNumberRepository.java
freddy61025: done

minion-certificate-manager/docker-it/src/test/java/org/opennms/horizon/minioncertmanager/stepdefs/MinionCertificateManagerTestSteps.java
```diff
@@ -85,36 +97,65 @@ public void createGrpcConnection() {
...
             100,
             timeout,
-            false);
-        assertTrue("Zip file created", success);
+            null);
+        this.serialNumber = serialNumber;
+        assertNotNull("P12 file created", serialNumber);
+    }
+
+    @Then("send isValid with last serial number and timeout {int}ms")
```
artnaseef, line 123: The wording on this step is a little unclear to me.  How about the following instead?

    send "isValid" request with last serial number and timeout {int}ms

minion-certificate-manager/docker-it/src/test/java/org/opennms/horizon/minioncertmanager/stepdefs/MinionCertificateManagerTestSteps.java
```diff
@@ -85,36 +97,65 @@ public void createGrpcConnection() {
...
+    public String checkLastIsValid(long timeout) throws InterruptedException {
+        LOG.info("Checking certificate serial number: {}", serialNumber);
+        Supplier<IsCertificateValidResponse> call = () -> clientUtils.getMinionCertificateManagerStub()
+            .isCertValid(IsCertificateValidRequest.newBuilder().setSerialNumber(serialNumber).build());
+        Supplier<IsCertificateValidResponse> isValid = retryUtils.retry(
+            () -> call,
+            result -> result.get().getIsValid(),
+            100,
+            timeout,
+            null);
```
artnaseef, line 133: Is there some timing here where it might come back as invalid but then later come back as valid?  If not, then the retries can be removed.

minion-certificate-manager/main/src/test/java/org/opennms/horizon/minioncertmanager/certificate/PKCS12GeneratorTest.java
```diff
@@ -71,6 +74,8 @@ void testGenerateP12() throws IOException, InterruptedException {
         // Verify the Results
         //
         Mockito.verify(mockCommandExecutor).executeCommand("openssl genrsa -out client.key.pkcs1 2048", outputDirPath.toFile());
+        Mockito.verify(mockCertificateReader).getX509Certificate(
+            outputDirPath.toFile().getAbsolutePath() + FileSystems.getDefault().getSeparator() + "client.signed.cert");
```
artnaseef, line 78: I like the use of the `mockCertificateReader`.

One thing I recommend here - instead of directly verifying the `getX509Certificate` method was called, use a `Mockito.when(...)` on that call and have it return a value.  Then, capture the return value fro the call to `generate` above, and compare them.

The advantage of that approach is not only does it make sure the right method call is made, it also ensures the result of that method call is returned by the `generate` method itself.  Then the `Mockito.verify(...)` of that method call is no longer needed, because the test already proves it is called.

minion-certificate-manager/main/src/main/java/org/opennms/horizon/minioncertmanager/certificate/CertificateReader.java
```diff
@@ -0,0 +1,42 @@
...
+ *******************************************************************************/
+package org.opennms.horizon.minioncertmanager.certificate;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+
+public class CertificateReader {
```
artnaseef, line 36: This looks good.

artnaseef (changes requested):

minion-certificate-manager/docker-it/src/test/java/org/opennms/horizon/minioncertmanager/stepdefs/MinionCertificateManagerTestSteps.java
freddy61025: i just think using the retry util is a common practices.

minion-certificate-manager/main/src/test/java/org/opennms/horizon/minioncertmanager/certificate/PKCS12GeneratorTest.java
freddy61025: done

artnaseef (approved):

PR #1251, opened 2023-06-27
HS-1737: make certificate verifier rely on serial number

freddy61025:
## Description
It also include HS-1762

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1737

## Checklist


splatch (approved):
No blockers from my side - there is one comment about `location-id` attribute for spans which you need to check with devops.

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
```diff
@@ -68,12 +76,30 @@ public ResponseEntity<Void> validateDebug(HttpServletRequest request) throws Exc
             logger.info("\t{}: {}", header, request.getHeader(header));
         }
 
-        String subjectDn = request.getHeader(headerName);
-        return validate(subjectDn);
+        String cert = request.getHeader("ssl-client-cert");
```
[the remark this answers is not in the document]
freddy61025: good point let me check will nginx forward expired cert to us.

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
```diff
@@ -68,12 +76,30 @@ public ResponseEntity<Void> validateDebug(HttpServletRequest request) throws Exc
...
 
     @GetMapping
-    public ResponseEntity<Void> validate(@RequestHeader("ssl-client-subject-dn") String clientSubjectDn) throws Exception {
+    public ResponseEntity<Void> validate(@RequestHeader("ssl-client-cert") String certificatePem) {
+        var span = Span.current();
+        CertificateParser parser;
+        try {
+            parser = new CertificateParser(certificatePem);
+        } catch (CertificateException ex) {
+            span.setAttribute("message", "Invalid certificate");
```
deejgregor, line 90: What you want to do here is record the exception using [`recordException`](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/exceptions/#recording-an-exception). Note: we don't need to mark this as an "escaped exception" because we don't re-throw the exception.
--


<br class="Apple-interchange-newline">

```suggestion
            span.recordException(ex);
            span.setStatus(StatusCode.ERROR, "invalid certificate: " + ex);
```

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
```diff
@@ -68,12 +76,30 @@ public ResponseEntity<Void> validateDebug(HttpServletRequest request) throws Exc
...
+            span.setStatus(StatusCode.ERROR);
+            return ResponseEntity.status(HttpStatus.FORBIDDEN).build();
+        }
+
+        IsCertificateValidResponse response = minionCertificateManagerClient.isCertValid(parser.getSerialNumber());
+        if (!response.getIsValid()) {
+            span.setAttribute("message", "Invalid certificate. serial number: " + parser.getSerialNumber());
```
deejgregor, line 97:
```suggestion
            span.setStatus(StatusCode.ERROR, "certificate manager reported that certificate is invalid");
```

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
```diff
@@ -68,12 +76,30 @@ public ResponseEntity<Void> validateDebug(HttpServletRequest request) throws Exc
...
+        return validate(cert);
     }
 
     @GetMapping
-    public ResponseEntity<Void> validate(@RequestHeader("ssl-client-subject-dn") String clientSubjectDn) throws Exception {
+    public ResponseEntity<Void> validate(@RequestHeader("ssl-client-cert") String certificatePem) {
+        var span = Span.current();
+        CertificateParser parser;
+        try {
+            parser = new CertificateParser(certificatePem);
```
deejgregor, line 88: If we can successfully parse the serial number, always add it as a span attribute. Then you can always search traces based on serial number and it will be available for debugging.

```suggestion
            parser = new CertificateParser(certificatePem);
            span.setAttribute("serial-number", parser.getSerialNumber());
```

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
```diff
@@ -87,23 +113,25 @@ public ResponseEntity<Void> validate(@RequestHeader("ssl-client-subject-dn") Str
...
-            span.setAttribute("location", location);
+            span.setAttribute(TENANTID_KEY, tenant);
+            span.setAttribute(LOCATIONID_KEY, location);
         }
 
         if (tenant.isBlank() || location.isBlank() || !location.matches("^\\d+$")) {
             span.setStatus(StatusCode.ERROR);
-            return ResponseEntity.notFound().build();
+            return ResponseEntity.status(HttpStatus.UNAUTHORIZED)
+                .header(TENANTID_KEY, tenant)
```
deejgregor, line 126: When the request is unauthorized, I don't think you want to return these headers.

deejgregor (changes requested):
See suggestions for the controller (one or two more coming)

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
deejgregor: If it might be useful, you could also add the problematic certificate itself as an attribute. Actually, the more I think about it, I think it would be a good idea. Then, if you need to debug a parsing issue, you have *everything* you need here to do so.

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
freddy61025: my reason to return it just hoping it can assist troubleshooting. Will that able to trace?

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
```diff
@@ -87,23 +113,25 @@ public ResponseEntity<Void> validate(@RequestHeader("ssl-client-subject-dn") Str
...
         }
 
-        var span = Span.current();
+
         if (span.isRecording()) {
             span.setAttribute("ssl-client-subject-dn", clientSubjectDn);
-            span.setAttribute("user", tenant);
-            span.setAttribute("location", location);
+            span.setAttribute(TENANTID_KEY, tenant);
+            span.setAttribute(LOCATIONID_KEY, location);
```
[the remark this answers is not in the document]
deejgregor: I'd suggest leaving the tenant attribute name as "user". Here's some background on why: https://github.com/OpenNMS-Cloud/lokahi/pull/1167#discussion_r1212585233

If we change either of these names, we should change it in the other places: `git grep -E 'setAttribute\("(user|location)'`

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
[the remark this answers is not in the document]
freddy61025: I have tested. Nginx will not trigger verifier if certificate is invalid or expired.

image

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
freddy61025: done

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
freddy61025: done

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
freddy61025: done

minion-certificate-verifier/main/src/main/java/org/opennms/horizon/minioncertverifier/controller/CertificateController.java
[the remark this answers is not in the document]
freddy61025: done

deejgregor (approved):

PR #1259, opened 2023-06-28
HS-1739: fix metric registry and add jmx reporter

freddy61025:
## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1739

## Checklist

### DIFF
--- a/minion/minion-flows/flows-features/src/main/feature/feature.xml
+++ b/minion/minion-flows/flows-features/src/main/feature/feature.xml
@@ -10,7 +10,8 @@
         <bundle>mvn:io.netty/netty-transport/${netty4.version}</bundle>
         <bundle>mvn:io.netty/netty-resolver/${netty4.version}</bundle>
         <bundle>mvn:io.netty/netty-codec/${netty4.version}</bundle>
-
+        <bundle>mvn:io.dropwizard.metrics/metrics-core/${dropwizard.metrics.version}</bundle>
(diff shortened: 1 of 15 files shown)


agalue:
I tested this via `jconsole`, and seems to be working.

However, I found that Minion offers metrics via `/metrics` on port 8181, but it looks like all are Ignite-related stuff, although related to Dropwizard Metrics. That means there might be a way to expose the JMX metrics through it (as Prometheus metrics), which would be easier to consume and use than `jconsole`.

agalue:
It would be great to have the equivalent metrics at the Minion Gateway level for the Sink Messages (all of them, not just flows). That way, we can compare if what the Gateway receives via Sink matches what the Minion sends through it via gRPC.

freddy61025:
> It would be great to have the equivalent metrics at the Minion Gateway level for the Sink Messages (all of them, not just flows). That way, we can compare if what the Gateway receives via Sink matches what the Minion sends through it via gRPC.

Lets don't make this PR too complex. I have create a separate ticket for it.

agalue:
I never said *this* PR should include the Gateway side, but without it, we're still blind (and won't be able to solve the Flow issues). Also, we need something at the Minion gRPC level (again, not this PR, but still blind without it, as that can also be the source of the problem).

JMX is nice, but not as lovely as Prometheus (yes, it can be another PR; but we should be consistent, I believe).

Finally, when is this going to be merged?

fooker (approved):

PR #1298, opened 2023-07-11
BTO-604: Ensure Keycloak would work in prod-mode.

agalue:
## Description

By default, Keycloak was forced to be deployed in development mode (despite the `Dockerfile` being ready for both development and production mode). I verified the correct way to configure Keycloak behind a proxy for TLS termination when using production mode and found the following:

* We have to use `KC_HOSTNAME_URL` instead of the combination of `KC_HOSTNAME` and `KC_HOSTNAME_PORT`. If we don't do that, Keycloak won't start properly (tested with 19.x and 20.x).
* `KC_PROXY` must be set to `edge` to ensure that restricted access (enforced by production mode) works when having an Nginx Ingress controller terminating TLS (so we wouldn't need to pass the certificates to Keycloak ever).

Due to how the Helm chart was designed, the best approach I found to fix this while maintaining compatibility with how all the different environments are configured is what this PR brings in. I tested the solution with Tilt (Dev Mode) and a local version of our Fabric environment in Prod Mode, and it works as intended.

## Jira link(s)
- https://opennms.atlassian.net/browse/BTO-604

## Checklist

### DIFF
--- a/charts/lokahi/templates/keycloak/keycloak-deployment.yaml
+++ b/charts/lokahi/templates/keycloak/keycloak-deployment.yaml
@@ -60,11 +60,11 @@ spec:
               value: kubernetes
             - name: KC_CACHE
               value: ispn
-            - name: KC_HOSTNAME
-              value: "{{ .Values.Host }}"
+            - name: KC_HOSTNAME_URL # Required in prod-mod to terminate TLS at Ingress
             {{- if .Values.Keycloak.HostnamePort }}
-            - name: KC_HOSTNAME_PORT
-              value: "{{ .Values.Keycloak.HostnamePort }}"
+              value: "{{ .Values.Protocol }}://{{ .Values.Host }}:{{ .Values.Keycloak.HostnamePort }}{{ .Values.Keycloak.Path }}"
+            {{- else }}
+              value: "{{ .Values.Protocol }}://{{ .Values.Host }}{{ .Values.Keycloak.Path }}"
             {{- end }}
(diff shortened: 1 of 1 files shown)


deejgregor (changes requested):
The changes look good but I have a request for one other change: We should make sure we do a production build for the docker image we push for lokahi-keycloak here: https://github.com/OpenNMS-Cloud/lokahi/blob/develop/.github/workflows/feature-keycloak-ui.yml#L62-L69

~~It looks like the github action supports specifying the "target". https://github.com/docker/build-push-action#customizing~~

Or, probably a better way to go since we're trying to have everything production-ready by default: have the Dockerfile build the production target by default, and then have tilt override that to build the development target.

agalue:
The changes in terms of the environment variables for the chart template will work regardless of Keycloak's execution mode. I tested them with Tilt in dev mode and my Minikube environment in prod mode (with the proposed changes here).

If something extra needs to be changed, I'm not aware of it; but any enhancements to this PR would be welcome.

deejgregor:
Alejandro and I looked at it, and we are using the production build everywhere in lokahi **except** in Tilt and with mk-docker-image.sh. The last build in the dockerfile is the normal output unless `--target` is passed to have it stop at an earlier build.

```
$ git grep 'target.*development'
Tiltfile:    target='development',
Tiltfile:    target='development',
build-tools/basic/mk-docker-images.sh:          --target development \
build-tools/basic/mk-docker-images.sh:          --target development \
```

deejgregor (approved):

PR #1301, opened 2023-07-11
Adding 'data-test' to allow easy find of an element for automation

maximbrener:


scthompsonopen (approved):
This is a common component, so these ids should really be passed in as props to the component, or at least be generic/contextual to the component instead of the specific implementation they're currently used for.

I'll approve for now, but we should update in the future.

PR #1326, opened 2023-07-18
BTO-687 - Adding readiness and liveness health checks for all opennms-* services & fixing Keycloak hlth chcks

jaymberry:
## Description

Adding readiness and liveness health checks for all opennms-* services. 

They have been added to all opennms-* services except for minion and ui. Minion is for testing, will ignore, but we need checks for ui.

The internal minion was having issues starting, I removed the startup probe in my local env, it began to work. I did not commit that change.

Keycloak checks are broken, needs work. It may just be the startup probe that is broken.

## Jira link(s)
- https://opennms.atlassian.net/browse/BTO-687

## Checklist


charts/lokahi/templates/opennms/alert/alert-deployment.yaml
```diff
@@ -77,6 +77,26 @@ spec:
           volumeMounts:
             - name: spring-boot-app-config-volume
               mountPath: "/app/config"
+          livenessProbe:
```
cgorantla, line 80: We are not using any Spring boot web components although they seem to be part of build. We should probably do custom health checks depending gRPC server API

jaymberry: I do not follow, we have http defined for the app in the chart, and in alert/src/main/resources/application.yaml the following was defined on the develop branch:
```
management:
  endpoints:
    web:
      base-path: "/"
      exposure:
        include: "*"
```

Since adding the health check, it seems to be working.

geraldhumphries: The health check endpoints are from spring-boot-actuator: https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.monitoring

I think it would be good to keep it this way.

cgorantla: okay, if actuator only depends on `spring-boot-actuator` not on `spring-boot-starter-web` then we should be good

rest-server/src/main/resources/application.yml
```diff
@@ -1,6 +1,23 @@
...
+        enabled: true
+        #add-additional-paths: true
+    web:
+      exposure:
+        include: "*"
+  health:
+    livenessstate:
+      enabled: true
+    readinessstate:
+      enabled: true
```
cgorantla, line 19: why these are added only for rest-server ?

cgorantla (commented):
Some clarifications for my understanding otherwise LGTM!

alert/src/main/resources/application.yaml
```diff
@@ -29,8 +29,14 @@ management:
   endpoints:
     web:
       base-path: "/"
+      path-mapping:
+        health: "healthz"
```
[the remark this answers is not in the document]
jaymberry: Yes, because the base-path was set to "/", I tried removing that and it would break alerts. But, adding it broke the default health path. So, I added a new path.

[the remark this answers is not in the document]
deejgregor: Yeah, let's keep the paths consistent for the Spring Boot services.

charts/lokahi/templates/opennms/minion-certificate-manager/minion-certificate-manager-deployment.yaml
```diff
@@ -73,6 +75,8 @@ spec:
           ports:
             - name: grpc
               containerPort: {{ .Values.OpenNMS.MinionCertificateManager.Port }}
+            - name: http
+              containerPort: {{ .Values.OpenNMS.MinionCertificateManager.PortHealth }}
```
deejgregor, line 79: Let's follow the same port naming scheme that we use for other services, which seems to be GrpcPort and Port. We'll need to rename `OpenNMS.MinionCertificateManager.Port` to `OpenNMS.MinionCertificateManager.GrpcPort`, but I think it's worth it for consistency.

charts/lokahi/values.yaml
```diff
@@ -20,6 +20,8 @@ OpenNMS:
     ImagePullPolicy: IfNotPresent
     Replicas: 1
     Port: 9090
+    #PortHealth: 8080
+    PortHealth: 9090
```
deejgregor, line 24: I don't see this variable used anywhere. It looks like we have port 8080 hard coded at https://github.com/OpenNMS-Cloud/lokahi/pull/1326/files#diff-1cea732e1a57c6b5b7ec5570fac2ac2649abbd117e85f63257a4b1b8154eac3fR90

charts/lokahi/values.yaml
```diff
@@ -237,6 +240,7 @@ OpenNMS:
     ImagePullPolicy: IfNotPresent
     Replicas: 1
     Port: 8990
+    PortHealth: 8080
```
deejgregor, line 243: I suggest we rename Port and PortHealth to GrpcPort and Port to follow the existing pattern used by other gRPC services.

charts/lokahi/values.yaml
agalue: Please be careful; as mentioned in another comment, the Prometheus Metrics port uses the same HTTP interface for everything managed via the actuator, AFAIK. If that's not the case, ensure the Prometheus Metrics work (which are currently working on port 8080 for several services already; see the Prometheus Annotations).

charts/lokahi/templates/opennms/alert/alert-deployment.yaml
```diff
@@ -77,6 +77,26 @@ spec:
...
+            httpGet:
+              path: "/healthz/liveness"
```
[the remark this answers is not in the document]
jaymberry: On services that the base-path is not affected, I do not see any data coming to /actuator/prometheus. Is there currently data being scraped from these services for prometheus?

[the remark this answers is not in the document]
agalue: Yes, that's why I'm insisting on verifying that it works. Because of that, many things would need changes and refactorization (for instance, all the work I've done to verify flows would only be valuable if Prometheus is reachable at that location).

charts/lokahi/values.yaml
jaymberry: I was going to leave it hardcoded, seeing that Port is used and not sure if there is a reason to have the user have the ability to change the port for health. Thoughts?

charts/lokahi/values.yaml
agalue: It depends. Several services assume 8080 for Monitoring via Prometheus (which I believe is the default port for the management endpoint via actuator):
https://github.com/OpenNMS-Cloud/lokahi/blob/develop/charts/lokahi/templates/opennms/metricsprocessor/metricsprocessor-deployment.yaml#L20
https://github.com/OpenNMS-Cloud/lokahi/blob/develop/charts/lokahi/templates/opennms/metricsprocessor/metricsprocessor-deployment.yaml#L70-L71

agalue: Back to what @deejgregor suggested, the Minion Gateway is a better example:
https://github.com/OpenNMS-Cloud/lokahi/blob/develop/charts/lokahi/templates/opennms/minion-gateway/minion-gateway-service.yaml#L11
https://github.com/OpenNMS-Cloud/lokahi/blob/develop/charts/lokahi/templates/opennms/minion-gateway/minion-gateway-deployment.yaml#L22
https://github.com/OpenNMS-Cloud/lokahi/blob/develop/charts/lokahi/templates/opennms/minion-gateway/minion-gateway-deployment.yaml#L92-L93

charts/lokahi/templates/opennms/alert/alert-deployment.yaml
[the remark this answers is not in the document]
jaymberry: No overlap, this is good.

charts/lokahi/templates/opennms/minion-certificate-manager/minion-certificate-manager-deployment.yaml
jaymberry: updated

charts/lokahi/values.yaml
jaymberry: I am not sure who changed the port on API, but it was there when I got this ticket and it breaks when I try to change it back to default.

charts/lokahi/values.yaml
jaymberry: I have changed this to @deejgregor's suggestion. This is not used in the cloud so should not affect prometheus.

charts/lokahi/templates/opennms/minion-certificate-manager/minion-certificate-manager-deployment.yaml
agalue: The thing is, are we telling the actual application to use different ports rather than the defaults? It looks like if you change the ports via the Values file, Spring Boot will ignore that and use its defaults. The ability to change them should also be implemented in the application itself. So, if that's hardcoded in Java, it should be hardcoded in the Helm Templates (not configurable via the Values file).

agalue:
We should homologate the configuration. It looks like the following is common to all the services except for alerts:
```yaml
management:
  endpoints:
    health:
      probes:
        enabled: true
    web:
      exposure:
        include: "*"
```
That should make the probes available at `/actuator/health/*`, and those services with Prometheus metrics should still work (available at `/actuator/prometheus`).
That said, I have the impression the following doesn't seem necessary for alerts:
```yaml
      base-path: "/"
      path-mapping:
        health: "actuator/health"
        # With the basepath set to /, need to specify the path-mapping.
```
Also, the following is assumed everywhere (based on what's running in Fabric) but only explicitly specified in a couple of places (the `minion-certificate-verifier`, and `rest-server`, but that's available for all), and if it is hardcoded in `application.yaml` or inferred by its default, it should *NOT* be configurable via Helm chart, as it is ignored anyway):
```yaml
management:
  server:
    port: 8080
```

jaymberry:
> We should homologate the configuration. It looks like the following is common to all the services except for alerts:
> 
> ```yaml
> management:
>   endpoints:
>     health:
>       probes:
>         enabled: true
>     web:
>       exposure:
>         include: "*"
> ```
I am not sure I follow, you mean make config used by all services?

> That should make the probes available at `/actuator/health/*`, and those services with Prometheus metrics should still work (available at `/actuator/prometheus`). That said, I have the impression the following doesn't seem necessary for alerts:
> 
> ```yaml
>       base-path: "/"
>       path-mapping:
>         health: "actuator/health"
>         # With the basepath set to /, need to specify the path-mapping.
> ```
I am not sure who added the base-path, it was there before I took the ticket. But, when I remove this and it breaks the application.

> Also, the following is assumed everywhere (based on what's running in Fabric) but only explicitly specified in a couple of places (the `minion-certificate-verifier`, and `rest-server`, but that's available for all), and if it is hardcoded in `application.yaml` or inferred by its default, it should _NOT_ be configurable via Helm chart, as it is ignored anyway):
> 
> ```yaml
> management:
>   server:
>     port: 8080
> ```
So, we should not allow this to be configurable from the helm chart and just hard code it in the templates?

jaymberry:
Have removed the following, alerts still work, not sure why my earlier tests failed, must have been unrelated. 
```
      base-path: "/"
      path-mapping:
        health: "actuator/health"
        # With the basepath set to /, need to specify the path-mapping. 
```

deejgregor (approved):

PR #1337, opened 2023-07-20
HS-1833: Fix steps for application series query

mikewrosey:
## Description
Step property was hardcoded to 2000000, which creates an interval of about 33 mins.

Grafana makes use of the `maxDataPoints` property which is tied to pixel width: _"The default value is the width (or number of pixels) of the graph, because you can only visualize as many data points as the graph panel has room to display."_

This PR grabs the width of the chart using the Chart.js chartArea property, then calculates the step with the formula `Math.floor((endTime - startTime) / maxDataPoints)`. Time is in ms here.  This formula is the same used in the grafana plugin.

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1833

## Checklist

### DIFF
--- a/ui/src/components/Flows/LineChart.vue
+++ b/ui/src/components/Flows/LineChart.vue
@@ -50,6 +50,10 @@ const props = defineProps({
   format: {
     required: true,
     type: Function as PropType<(val: string) => string>
+  },
+  getChartAreaWidthForDataPoints: {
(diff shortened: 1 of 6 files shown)


freddy61025 (approved):
I am not good at javascript. Step calculation logic looks good to me.

PR #1339, opened 2023-07-20
HS-1485: Refactor Alert Service Cucumber Tests

vserrago:
## Description
This PR is intended to precede the actual changes for HS-1485; I figure that it might be easier to review & merge this PR separately. 

The scope of this PR is the alert service cucumber tests. The major changes are that:
- Many steps have been updated remove explicit references to the `tenantId`.
- Step classes have been streamlined by removing dead and duplicated code.
- Some scenario and step descriptions have been reworded to reduce verbosity.

No tests have been added or removed. This refactoring is necessary with the upcoming changes. Previously, on each scenario many duplicates of all of a tenant's alert conditions were created, for each policy specified in a background step. This means that an event would generate way too many alerts for duplicate policies that probably should not be allowed to exist together regardless.

In my prior PR I had attempted to get around this with tenantId aliases, to which I received feedback that it was hard to understand. In this altered approach we no longer have to care about the tenant in each step, having a reference to it via the `TenantSteps` class.

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1485

## Checklist

### DIFF
--- a/alert/src/test/java/org/opennms/horizon/alertservice/stepdefs/AlertTestSteps.java
+++ b/alert/src/test/java/org/opennms/horizon/alertservice/stepdefs/AlertTestSteps.java
@@ -40,6 +40,7 @@ import io.restassured.config.RestAssuredConfig;
 import io.restassured.path.json.JsonPath;
 import io.restassured.response.Response;
 import io.restassured.specification.RequestSpecification;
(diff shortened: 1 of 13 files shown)


cgorantla (approved):
LGTM!

PR #1341, opened 2023-07-21
HS-1785: Fix passive discovery node detection

cgorantla:
## Description
Passive discovery node creation was broken after changes to locationId
Add test to cover passive discovery node creation scenario

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1785

## Checklist

### DIFF
--- a/inventory/docker-it/src/test/resources/org/opennms/horizon/inventory/passive-discovery.feature
+++ b/inventory/docker-it/src/test/resources/org/opennms/horizon/inventory/passive-discovery.feature
@@ -35,4 +35,15 @@ Feature: Passive Discovery
     And A GRPC request to get passive discovery list
     Then The passive discovery toggle should be false
 
+  Scenario: Send new suspect event and create node
+    Given Passive Discovery fields to persist for location named "Testing Passive Discovery"
+    When A GRPC request to upsert a passive discovery
+    And A GRPC request to get passive discovery list
+    Then The upserted and the get of passive discovery should be the same
+    Given A GRPC request to enable a passive discovery
+    When A new suspect event sent to Inventory for location named "Testing Passive Discovery" and IP Address "192.168.5.1"
+    Then A new node should be created with location "Testing Passive Discovery" and IP Address "192.168.5.1"
+
+
+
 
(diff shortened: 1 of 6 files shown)


inventory/docker-it/src/test/resources/org/opennms/horizon/inventory/passive-discovery.feature
```diff
@@ -35,4 +35,15 @@ Feature: Passive Discovery
...
     Then The passive discovery toggle should be false
 
+  Scenario: Send new suspect event and create node
+    Given Passive Discovery fields to persist for location named "Testing Passive Discovery"
+    When A GRPC request to upsert a passive discovery
+    And A GRPC request to get passive discovery list
+    Then The upserted and the get of passive discovery should be the same
+    Given A GRPC request to enable a passive discovery
+    When A new suspect event sent to Inventory for location named "Testing Passive Discovery" and IP Address "192.168.5.1"
+    Then A new node should be created with location "Testing Passive Discovery" and IP Address "192.168.5.1"
```
vserrago, line 45: From how this scenario is written out it seems like there are two separate tests here:

```suggestion
  Scenario: Upserted and Get requests return the same value
    Given Passive Discovery fields to persist for location named "Testing Passive Discovery"
    When A GRPC request to upsert a passive discovery
    And A GRPC request to get passive discovery list
    Then The upserted and the get of passive discovery should be the same

  Scenario: A suspect event should create a new node if it doesn't exist
    Given A GRPC request to enable a passive discovery
    When A new suspect event sent to Inventory for location named "Testing Passive Discovery" and IP Address "192.168.5.1"
    Then A new node should be created with location "Testing Passive Discovery" and IP Address "192.168.5.1"
```

Is that the case? If not, I think a more descriptive scenario would help clarify.

cgorantla: Each scenario is independent.
We need to have passive discovery to enable toggling and subsequent new suspect handling. So that's why it combines two steps into one.

vserrago: Looking at the other scenarios, the all seem to have the same setup steps. Maybe as a nit, they can be refactored into the background steps.

vserrago (approved):

PR #1347, opened 2023-07-24
HS-1858: Validate locations before getting certificate

cgorantla:
Check for tenantId, location validity before getting certificate

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1858

## Checklist


rest-server/src/main/java/org/opennms/horizon/server/service/GrpcMinionCertificateManager.java
```diff
@@ -20,14 +21,22 @@ public class GrpcMinionCertificateManager {
...
     public Mono<CertificateResponse> getMinionCertificate(Long locationId, @GraphQLEnvironment ResolutionEnvironment env) {
         String tenantId = headerUtil.extractTenant(env);
+        String authHeader = headerUtil.getAuthHeader(env);
+        try {
+            var monitoringLocation = inventoryClient.getLocationById(locationId, authHeader);
+            var location = monitoringLocation.getId();
+            CertificateResponse minionCert = mapper.protoToCertificateResponse(client.getMinionCert(tenantId, location, authHeader));
+            return Mono.just(minionCert);
+        } catch (Exception e) {
+            return Mono.error(e);
```
geraldhumphries, line 37: We should not be returning exceptions through the API. Maybe throw a new custom exception for "location does not exist" when there's a 404, catch it here, and return a more simple message?

geraldhumphries (changes requested):

rest-server/src/main/java/org/opennms/horizon/server/service/GrpcMinionCertificateManager.java
cgorantla: Mono doesn't have any other way to return errors other than returning whole exception.

It consumes the exception and returns an error message as shown below

```
{"query":"query {\n  getMinionCertificate(locationId: 404){\n    certificate\n    password\n  }\n}"}

< 200 OK OK
< Date: [Tue, 25 Jul 2023 15:47:21 GMT]
< Vary: [Origin, Access-Control-Request-Method, Access-Control-Request-Headers]
< Content-Type: [application/json]
< content-length: [309]

{"errors":[{"message":"Exception while fetching data (/getMinionCertificate) : io.grpc.StatusException: NOT_FOUND: Given location doesn't exist.","locations":[{"line":2,"column":3}],"path":["getMinionCertificate"],"extensions":{"classification":"DataFetchingException"}}],"data":{"getMinionCertificate":null}}

```

agalue: The problem is that it is returning an HTTP 200. The consumer of the Certificate Verifier is the Nginx Ingress Controller. Returning a 200 means everything went well (it doesn't matter if the content shows an error; that'd be ignored). If there is an error, it should return something different than 200.

https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#external-authentication

cgorantla: This API is GraphQL query.  Not sure if we are using this anywhere other than UI.
If we need to include proper status codes, we may have to change API to return Mono<ResponseEntity> and UI need to handle updated API.  Guess that's beyond scope of this bug.

freddy61025: I would suggest to fix this issue for now. And create another task to standardize common error message in BFF.

rest-server/src/main/java/org/opennms/horizon/server/service/GrpcMinionCertificateManager.java
```diff
@@ -20,14 +24,25 @@ public class GrpcMinionCertificateManager {
...
+            var monitoringLocation = inventoryClient.getLocationById(locationId, authHeader);
+            var location = monitoringLocation.getId();
+            CertificateResponse minionCert = mapper.protoToCertificateResponse(client.getMinionCert(tenantId, location, authHeader));
+            return Mono.just(minionCert);
+        } catch (StatusRuntimeException e) {
+            if (e.getStatus().getCode().equals(Status.Code.NOT_FOUND)) {
+                return Mono.error(new LocationNotFoundException("Invalid location id " + locationId));
+            }
+            // fallback to generic exception
+            return Mono.error(new IllegalArgumentException("Exception while fetching Minion certificate for location {}" + locationId));
```
geraldhumphries, line 44: This might be a small bug, `("Exception while fetching Minion certificate for location {}" + locationId)` won't fill that placeholder.

geraldhumphries (approved):
Small bug to fix but it shouldn't block the PR, LGTM!

rest-server/src/main/java/org/opennms/horizon/server/service/GrpcMinionCertificateManager.java
vserrago:
```suggestion
            return Mono.error(new IllegalArgumentException("Exception while fetching Minion certificate for location " + locationId));
```

freddy61025 (approved):

vserrago (approved):

PR #1350, opened 2023-07-25
HS-1654: Fix extra URL fragments caused by an incompatibility between Keycloak and VueJS router

geraldhumphries:
## Description

Keycloak uses URL fragments to keep track of items in localStorage. VueJS's router is incompatible with how Keycloak does this, which causes the fragments to stick around and accumulate. When the URL gets long enough it causes the nginx ingress to return 502 errors.

This is a workaround that fixes the issue. Keycloak needs to be fully initialized before the `@/router` file is imported anywhere - in main.ts the import was moved to the onReady callback, but it also needed to be changed in welcomeStore since it's imported in App.vue. 

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1654

## Checklist

### DIFF
--- a/ui/src/components/Inventory/InventoryTextAnchorList.vue
+++ b/ui/src/components/Inventory/InventoryTextAnchorList.vue
@@ -33,7 +33,8 @@
 
 <script lang="ts" setup>
 import { Anchor } from '@/types/inventory'
-import router from '@/router'
+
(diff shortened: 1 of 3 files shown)


scthompsonopen (approved):
Makes sense to me. I haven't tested it directly, but everything in this makes sense.

PR #1384, opened 2023-08-03
BTO-773: Fix Ignite Cluster Discovery.

agalue:
## Description
I found errors on the readiness probe when the Minion Gateway starts due to Ignite issues while trying to identify and negotiate with cluster members. This doesn't happen all the time, but when Deployment restarts (due to an upgrade, for instance), there is more than one Gateway running (the one that was already there and the new one that is trying to start), which confuses Ignite due to how the discovery service was defined.

With this patch, I found no errors on the initial boot with multiple replicas, after roll-restart, and after scaling up and down, so I'm confident about not having those readiness errors anymore.

There are a few warnings, but I'd need to study Ignite to see how we can improve things there, but we can do that on another PR.

## Jira link(s)
- https://opennms.atlassian.net/browse/BTO-773

## Checklist


bborovan (approved):
Looks good

PR #1390, opened 2023-08-04
Tilt: ssl-check.sh at startup and improve context/namespace handling

deejgregor:
- Run tools/ssl-check.sh in Tilt
- ssl-check.sh: remove effective no-op of '2>&1' before '> /dev/null'
- Allow specifying arbitrary kubectl options in SSL tools
- Tilt/ssl-check.sh: Run HTTP checks separately after ingress is up
- ssl-check.sh: also retry minion check
- tilt ci: Make sure necessary names are in /etc/hosts
- ssl-check.sh: connect to 127.0.0.1 to reach minion

## Jira link(s)
- https://opennms.atlassian.net/browse/BTO-784

## Checklist

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -191,6 +191,15 @@ def load_certificate_authority(secret_name, name, key_file_name, cert_file_name)
 def generate_certificate(secret_name, domain, ca_key_file_name, ca_cert_file_name):
     local('./install-local/generate-and-sign-certificate.sh "default" {} {} {} {}'.format(domain, secret_name, ca_key_file_name, ca_cert_file_name));
 
+def ssl_check(domain, port):
+    local([
+        './tools/ssl-check.sh',
+        '-t',
+        '5',
+        domain,
+        str(port),
+    ])
+
 # If you don't specify a resource, the button will be added to the global nav (location.NAV).
 def create_devmode_toggle_btn(devmode_key, resource=None):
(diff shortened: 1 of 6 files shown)


Tiltfile
```diff
@@ -49,8 +49,10 @@ load('ext://uibutton', 'cmd_button', 'location')
 load('ext://helm_remote', 'helm_remote') # for simple charts like jaeger and cert-manager
 load('ext://helm_resource', 'helm_resource', 'helm_repo') # for charts that we want to have resources for
 
+context = k8s_context()
```
geraldhumphries, line 52: nit: Slight inconsistency here, the `reload-certificates` button uses this variable, but other places like `load_certificate_authority` call the `k8s_context()` function directly. `k8s_namespace()` is similar but doesn't have a variable. I suggest changing these to be consistent - either always use variables or always call functions.

geraldhumphries (approved):
One small nitpick that won't block this PR, otherwise looks good.

Just to open the discussion, kubectl options being passed into the commands like this makes the interface for the scripts wide open. This is good and bad IMO. It's flexible, but could lead to a maintenance headache if someone uses it to do something a little too fancy.

Right now only the k8s context and namespace are used. Maybe we should keep them explicit? I see that we aren't always passing in the context (`install-local` sets the context in the `create_cluster` method).

PR #1395, opened 2023-08-08
LOK-1889: Updates to the Inventory Page

scthompsonopen:
## Description
Amanda provided some suggested updates for the Inventory page, specifically a redesign of the inventory card, and a slight reworking of how you can modify an individual node. The only updates intended for this sprint were the visual updates to the card, but I went ahead and made the functionality updates as well, and also tried to improve the performance of the page by tweaking some graphql queries to be closer to a true BFF style so we can get everything we need to load the Inventory Page with as few queries as possible. 

Prior to this update in order to load the inventory page we needed 3 calls to load each type of node (monitored, unmonitored and detected) then we needed to gather all the nodeIds manually and load the tags in an additional query, and then for each monitored node, we needed to do an individual metric query for each node. This meant that having upwards of say 15 monitored nodes, meant we needed to make 19 individual graphQL calls in order to load the inventory page. I've reduced that number down to 2. One for all the nodes, and their tags, and another for all the metrics. I could not determine how to properly gather the latency metrics and the nodes and combine them in Java. 

Old Card Design
image

New Card Design
image

Note, I made one visual change from the Figma design to the final product, I used an unstyled button instead of a chip to represent the `tags: #` as that was a clickable entity. I thought it might encourage the user a bit more to click on what looks like a button.

Old Tag Deletion Modal
image

New Tag Deletion Modal
image

Minor Updates to Tag Manager
Since we have plans to fix dark mode post EAR, I thought I would fix an issue that I noticed in the overlay which was defaulting to dark mode. I've fixed things so that they use the feather variables which will react to dark mode being properly enabled.

image

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-1889

## Flagged for review
I've made some tweaks to some BE files in order to support this page better. If my tweaks are no good, let's have a chat to see how we can create an endpoint to properly support this page.

## Checklist
* [X] Follows Lōkahi's [development guidelines.](https://github.com/OpenNMS-Cloud/lokahi/wiki/Development-Guidelines)


geraldhumphries (approved):
BE changes look good. FE changes look good and should significantly increase maintainability (especially inventoryQueries). LTGM!

mikewrosey (approved):
Looks great. I did notice that the search by labels/tags is not working. It makes a call but doesn't send any search variables. I know there's a draft PR for some fixes around that. Do we want to hide the search until that's merged? That's the only thing I found, so all in all it's awesome stuff here.

PR #1410, opened 2023-08-15
LOK-1917: Alerts from Internal Events are missing Descriptions

freddy61025:
## Description
image
image

## Jira link(s)
- https://opennms.atlassian.net/browse/HS-1917

## Checklist


vserrago:
It looks like to me that the message being displayed will either be "timeout" or an exception message. These might be reasonable for the internal events to report, however I don't think they should be exposed to the end user. Exception messages can leak information, and "timeout" isn't very descriptive.

Perhaps in the alert service we should provide a more informative description, maybe via resource bundle, or something similar? That seems like a story or more in itself, plus the fact that we have to know what language to render in. Maybe what we can do is capture that work in Jira, and then just make sure that we don't leak exception messages to the end user.

vserrago (approved):

PR #1426, opened 2023-08-23
BTO-783: minion-gateway: flesh out traces for all connection types and streaming messages including identity attributes

deejgregor:
- BTO-783: Add identity attributes to MinionToCloudMessage streaming and message spans
- BTO-783: Add tracing details for CloudToMinionRPC dispatch calls
- BTO-783: Add tracing for CloudToMinionMessage
- BTO-783: Add tracing for MinionToCloudRPC
- Tilt: include minion-gateway messages in tracing spans by default
- Don't use GlobalOpenTelemetry.get().getTracer anymore
- remove unused import

## Description

We should now have tracing spans created for **all** Minion<->Minion-gateway communications:
* CloudToMinionMessage
* MinionToCloudMessage
* CloudToMinionRPC
* MinionToCloudRPC

Streaming gRPC calls (the first two in the list above) are a little interesting, because by default, the OpenTelemetry gRPC instrumentation only adds spans at the *connection* level. These are both long-lived streams, and that can result in massive traces with a wide variety of messages intermingled. These changes now create a new trace for each message in a gRPC stream, and that new trace is linked to the connection-level trace (note: the top-level connection-level span won't be recorded *until the connection is closed*, so even though there is a linkage, you might not see the connection-level trace immediately).

This adds a few key attributes to the spans that seemed useful based on perusing the messages. Also, useful for development debugging: it can record the gRPC message payloads in the span. This is turned off by default, but can be enabled by setting environment variables, which are enabled by default in Tilt:
* DEBUG_SPAN_CONTENT -- show the decoded gRPC inner content for the message (in whatever form makes sense for the message)
* DEBUG_SPAN_FULL_MESSAGE -- show the full gRPC message (including metadata that is included in other attributes--like the identity of the minion, message ID, RPC moduleId, twin consumerKey, etc.)

Here is an example of a heartbeat trace (interestingly, you'll see about 2/3 of the way down, the *incoming* heartbeat from the Minion ends up triggering an echo RPC back to the Minion):
image

Here are the attributes on the incoming heartbeat message:
image

And here's the same for the RPC echo call:
image
image
...
image

## Jira link(s)
- https://opennms.atlassian.net/browse/BTO-783

## Flagged for review

This will unfortunately fail the SonarCloud coverage check because there is no test coverage on the new code. I started doing some refactoring to write some tests to exercise the tracing code I added, however I discovered that the classes I touched are completely untested today (at least from SonarCloud's perspective).

## Checklist

### DIFF
--- a/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/grpc/common/TenantIDGrpcServerInterceptor.java
+++ b/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/grpc/common/TenantIDGrpcServerInterceptor.java
@@ -38,7 +38,6 @@ import io.grpc.ServerInterceptor;
 import io.grpc.Status;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.trace.Span;
 
 import java.time.Duration;
--- a/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/ipc/grpc/server/OpennmsGrpcServer.java
+++ b/minion-gateway/ipc-grpc-server/src/main/java/org/opennms/horizon/shared/ipc/grpc/server/OpennmsGrpcServer.java
@@ -40,6 +40,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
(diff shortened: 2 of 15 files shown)


deejgregor:
@cgorantla FYI: this PR is mostly ready, except I need to switch to a better way of doing this with Spring Boot application properties:
```
        private static final boolean DEBUG_SPAN_FULL_MESSAGE = Boolean.parseBoolean(System.getenv().getOrDefault("DEBUG_SPAN_FULL_MESSAGE", "false"));
        private static final boolean DEBUG_SPAN_CONTENT = Boolean.parseBoolean(System.getenv().getOrDefault("DEBUG_SPAN_CONTENT", "false"));
```

If you want to review in the meantime, go for it. This *might* help with your Minion gateway connectivity work, so let me know if you want me to try to wrap it up soon to assist with that.

cgorantla (approved):
LGTM!

PR #1449, opened 2023-09-07
LOK-1891: BFF API Concerns

vserrago:
## Description

This is not the complete change-set for this ticket; more will come in the future.

- Disables graphql `GET` requests. Currently, mutations can be made via a `GET` request, which is a CSRF issue. It's generally suggested that if you allow graphql over `GET`, only queries should be allowed. We currently don't use `GET` graphql requests at all in our front-end so I don't see the value in supporting them at present.
- Exposes configuration to enable depth and complexity limiting for graphql requests. My plan is to provide default limits in a later PR, at which point, they will be configured to run.
- Adds web exception handling to prevent stacktrace exposure and unexpected exception messages from being returned in a response.
- Adds graphql exception handling to prevent unexpected exception messages from being returned in a response.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-1891

## Checklist


bouff (commented):
LGTM.  Will give a chance for others to provide feedback - otherwise will approve tomorrow afternoon.

rest-server/src/main/java/org/opennms/horizon/server/config/GraphqlConfig.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+
+/**
+ * Provides fine-tuned configuration for GraphQL.
+ *
+ * @see io.leangen.graphql.spqr.spring.autoconfigure.BaseAutoConfiguration
+ */
+@Configuration
+@Slf4j
+public class GraphqlConfig {
```
freddy61025, line 55: Is there any missing config update? I cannot find any of those config in PR. Suggest to put those config in charts/lokahi/values.yaml.

vserrago: As mentioned in the PR description I'll be adding those in a later PR.

PR #1465, opened 2023-09-18
LOK-1930: Minion cleanup

cgorantla:
## Description
1. Reconnect after Sink Stream errors out
2. Add a workaround to send trap config  to avoid timeout errors
3. Shutdown Minion properly when non-recoverable error happens
4. Update logs on Minion grpc client ( don't show exceptions unless debug is set)
5. Update logs in plugins ( don't show exceptions unless debug is set)

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-1930

## Checklist


freddy61025 (approved):
LGTM

minion/plugins/snmp-plugin/src/main/java/org/opennms/horizon/minion/snmp/SnmpMonitor.java
```diff
@@ -159,21 +159,27 @@ public CompletableFuture<ServiceMonitorResponse> poll(MonitoredService svc, Any
 
             return future;
         } catch (NumberFormatException e) {
-            LOG.debug("Number operator used in a non-number evaluation", e);
+            if (LOG.isDebugEnabled()) {
```
bouff, line 162: Nit: is the conditional needed in this case (if debug then write a debug message)?

There isn't any intense computation for the log message itself, so I'd recommend against having these..

minion/minion-grpc/grpc-client/src/test/java/org/opennms/horizon/minion/grpc/channel/SSLChannelFactoryTest.java
```diff
@@ -109,9 +110,12 @@ void testEmptyTrustStoreWithException() throws Exception {
         channelFactory.setTrustStoreType("pkcs12");
         channelFactory.setTrustStorePassword("changeit");
         when(keyStoreFactory.createKeyStore("pkcs12", trustStore, "changeit")).thenThrow(new GeneralSecurityException(""));
-        when(channelBuilderFactory.create(eq("baz"), eq(443), isNull(), any(TlsChannelCredentials.class))).thenReturn(managedChannelBuilder);
 
-        channelFactory.create("baz", 443, null);
+        try {
+            channelFactory.create("baz", 443, null);
+        } catch (Exception e) {
+
```
bouff, line 117: Since we are swallowing the exception here... and this seems to be unexpected - why don't we have the test fail if an exception is thrown on create?

minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
```diff
@@ -289,7 +288,9 @@ private void handleDisconnect() {
             try {
                 rpcStream.onCompleted();
             } catch (Exception exc) {
-                LOG.info("Error on cleanup of existing rpc stream", exc);
+                if (LOG.isDebugEnabled()) {
+                    LOG.error("Error on cleanup of existing rpc stream", exc);
```
bouff, line 292: Nit: I find it confusing that the lower level `DEBUG` will result in showing/hiding a (higher level) `ERROR` log.

Why isn't the level kept the same?  (E.g. `DEBUG` is on, one will see `DEBUG` logs)

inventory/main/src/main/java/org/opennms/horizon/inventory/service/TrapConfigService.java
```diff
@@ -60,7 +65,17 @@ public class TrapConfigService {
...
 
+    private final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+        .setNameFormat("trap-config-update-scheduler-%d")
+        .build();
+    private final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+
     @EventListener(ApplicationReadyEvent.class)
+    public void scheduleConfigUpdate() {
+        // This is work around for Minion not to have timeout for CloudToMinion Stream.
+        // Keep sending Trap config every 15 mins.
```
bouff, line 76: Please raise a tech-debt task (and include the Jira ID here, in the comment) with the context for this approach.

minion/minion-grpc/grpc-client/src/test/java/org/opennms/horizon/minion/grpc/channel/SSLChannelFactoryTest.java
cgorantla: Updated test.

minion/minion-grpc/grpc-client/src/main/java/org/opennms/horizon/minion/grpc/MinionGrpcClient.java
cgorantla: I kept this to ERROR to make it easier to search and visibility purpose on the console. We can go either way.

vserrago: I also don't think it's a good idea to gate higher severity log statements behind lower severity logs being enabled. As mentioned, the `logger#is<level>Enabled()` methods are for preventing computationally-heavy log statements from getting invoked. To have it gate higher-severity messages would not be intuitive to someone like an operator unless they were looking directly at this source file.

I think it would be better to log them at the debug level, or alternatively, at the warn level since they are an error we can otherwise suppress/recover from.

vserrago (changes requested):

bouff (approved):

vserrago (approved):

PR #1479, opened 2023-09-22
LOK-2079: Re-add ability to make CORS requests

vserrago:
## Description
Prior to https://github.com/OpenNMS-Cloud/lokahi/pull/1449, the graphql endpoint would allow CORS requests. When it was replaced, this behaviour was not carried over, causing an interruption in frontend development. Generally it is better to not allow CORS by default given that only our UI is making requests, and via the same origin. My fix here is to provide the configuration option to expose this option via configuration.

This is how you can provide helm chart values without having git tracking them:

In `tilt-config.json`:

```json
{
  "devmode": [],
  "values": ["./tilt-helm-values-local.yaml"]
}
```

In `tilt-helm-values-local.yaml`:

```yaml
OpenNMS:
  API:
    CorsAllowed: true
```

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2079

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 # For local dev and sample envs, a place to put required dependencies, secrets,
 # or test data that we do not want to push to repo.
 tilt_config.json
+tilt-helm-values-local.yaml
 dev/tmp
 local-sample/tmp
 
--- a/charts/lokahi/templates/opennms/api/api-deployment.yaml
+++ b/charts/lokahi/templates/opennms/api/api-deployment.yaml
@@ -86,12 +86,16 @@ spec:
               value: "{{ .Values.OpenNMS.API.PackagedMinionFile }}"
             - name: MINION_ENDPOINT
               value: "{{ .Values.OpenNMS.API.MinionEndpoint }}"
+            {{- with .Values.OpenNMS.API.CorsAllowed }}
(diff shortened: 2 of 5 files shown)


mikewrosey:
I might be missing something but I checked out this branch and looks like the issue is still present.
```
Access to fetch at 'https://onmshs.local:1443/api/graphql' from origin 'http://localhost:8080' has been blocked by CORS policy
```

.gitignore
```diff
@@ -9,6 +9,7 @@
 # For local dev and sample envs, a place to put required dependencies, secrets,
 # or test data that we do not want to push to repo.
 tilt_config.json
+tilt-helm-values-local.yaml
```
cgorantla, line 12: Since tilt is only used for local development, why can't we directly add this to repo ?

vserrago: I added this so that a developer can specify helm values locally that they would not necessarily want to commit. Similar to a git-ignored `.env.development.local` in the UI, or `application-local.properties` in Spring.

cgorantla: I understand that but why can't we add  this setting into tilt by default.

```
OpenNMS:
  API:
    CorsAllowed: true
```

cgorantla (approved):
lgtm!

PR #1492, opened 2023-10-03
LOK-2112: Add shutdown hook, add delay for shutdown 

cgorantla:
## Description
Add shutdown hook (to display the error message at the end)
Add configurable shutdown delay with `MINION_DELAY_BEFORE_SHUTDOWN` env.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2112

## Checklist

### DIFF
--- a/minion/minion-ipc/ipc-heartbeat/src/main/java/org/opennms/horizon/minion/ipc/heartbeat/internal/HeartbeatProducer.java
+++ b/minion/minion-ipc/ipc-heartbeat/src/main/java/org/opennms/horizon/minion/ipc/heartbeat/internal/HeartbeatProducer.java
@@ -30,9 +30,6 @@ package org.opennms.horizon.minion.ipc.heartbeat.internal;
 
 
 import com.google.protobuf.Timestamp;
-import java.util.Timer;
-import java.util.TimerTask;
-
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.opennms.cloud.grpc.minion.Identity;
@@ -40,6 +37,9 @@ import org.opennms.horizon.grpc.heartbeat.contract.HeartbeatMessage;
 import org.opennms.horizon.shared.ipc.rpc.IpcIdentity;
 import org.opennms.horizon.shared.ipc.sink.api.MessageDispatcherFactory;
 import org.opennms.horizon.shared.ipc.sink.api.SyncDispatcher;
+
+import java.util.Timer;
+import java.util.TimerTask;
 
 @Slf4j
 @RequiredArgsConstructor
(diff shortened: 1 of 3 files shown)


minion/minion-ipc/ipc-heartbeat/src/main/java/org/opennms/horizon/minion/ipc/heartbeat/internal/HeartbeatProducer.java
```diff
@@ -70,7 +70,11 @@ public void run() {
                         .build();
                     dispatcher.send(heartbeatMessage);
                 } catch (Throwable t) {
-                    log.error("An error occurred while sending the heartbeat. Will try again in {} ms", PERIOD_MS, t);
+                    if (log.isDebugEnabled()) {
+                        log.debug("An error occurred while sending the heartbeat. Will try again in {} ms", PERIOD_MS, t);
+                    } else {
+                        log.error("An error occurred while sending the heartbeat. Will try again in {} ms", PERIOD_MS);
```
jbeuree, line 76: Why are we changing the log severity of this based on if debug logging is enabled?

deejgregor: If there are one or two common exceptions that we expect, we might want to explicitly check for those (and log at lower level if that's the right thing to do) and let everything else fall through to the error log and log the throwable.

Or, if we want to keep this structure, we definitely want to log as an error in both cases, and when we decide not to include the Throwable in the log, we need to include the exception as a string (or ideally: both the type and message)--otherwise, we won't have any information to debug issues.

cgorantla: I missed appending exception message in the log  but we had this discussion on log levels before https://github.com/OpenNMS-Cloud/lokahi/pull/1465#discussion_r1329255485

jbeuree: I do agree with some of the comments that if there are specific instances where the exception/failure can be considered an error, we should always log it as such, and then let the others go through at a lower severity. I suspect this whole conversation may be tangental to getting the shutdown hook/timer in though.
I also wonder if it would be 'normal' for there to be occasional heartbeat failures and they'd only be considered errors after a period of time once we determine comms are down and cannot be re-established.

cgorantla: yeah, I saw this error and wanted to take care of it. we can move this to debug and then take care of rate controlled logging for all of these in another issue.

jbeuree (approved):

PR #1536, opened 2023-10-31
LOK-2055: ignore node enrichment if ip lookup fail

freddy61025:
## Description
This PR will skip node enrichment if anything happen during grpc call to inventory rather than lost that flow record.
It also make sure findByIpAddressAndLocationIdAndTenantId only return single record. Priority return snmpPrimary first then first record.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2055

## Checklist

### DIFF
--- a/inventory/main/src/main/java/org/opennms/horizon/inventory/service/IpInterfaceService.java
+++ b/inventory/main/src/main/java/org/opennms/horizon/inventory/service/IpInterfaceService.java
@@ -3,6 +3,7 @@ package org.opennms.horizon.inventory.service;
 import lombok.RequiredArgsConstructor;
 import org.opennms.horizon.azure.api.AzureScanNetworkInterfaceItem;
 import org.opennms.horizon.inventory.dto.IpInterfaceDTO;
+import org.opennms.horizon.inventory.exception.InventoryRuntimeException;
 import org.opennms.horizon.inventory.mapper.IpInterfaceMapper;
 import org.opennms.horizon.inventory.model.AzureInterface;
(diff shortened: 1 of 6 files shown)


inventory/main/src/main/java/org/opennms/horizon/inventory/service/IpInterfaceService.java
```diff
@@ -75,6 +83,12 @@ public void createOrUpdateFromScanResult(String tenantId, Node node, IpInterface
                 modelRepo.save(ipInterface);
             }, () -> {
                 IpInterface ipInterface = mapper.fromScanResult(result);
+                modelRepo.findByIpAddressAndLocationIdAndTenantId(ipInterface.getIpAddress(), node.getMonitoringLocationId(), tenantId)
```
cgorantla, line 86: Note that this is also used in update ( scan can run multiple times).  Also throwing exception will prevent adding other interfaces.

cgorantla (changes requested):

inventory/main/src/main/java/org/opennms/horizon/inventory/service/IpInterfaceService.java
freddy61025: as per discussion, replace duplication checking by make sure findByIpAddressAndLocationIdAndTenantId only return single record. Priority return snmpPrimary first then first record.

vserrago (approved):

cgorantla (approved):
LGTM!

PR #1538, opened 2023-11-01
LOK-2171: Fix discovery page call for tags when no discovery available

mikewrosey:
## Description
Right now if we have only passive type discoveries or only active type discoveries, the frontend will make a tags call for both types. This fixes things so that it will only make the tags call for the type which has available discoveries.

Also fixes the selected/highlighted issue when both a passive and active discovery have the same id.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2171

## Checklist

### DIFF
--- a/ui/src/components/Discovery/DiscoveryListCard.vue
+++ b/ui/src/components/Discovery/DiscoveryListCard.vue
@@ -9,7 +9,7 @@
         v-for="(item, index) in list"
         :key="index"
         class="discovery-name pointer"
-        :class="{ selected: selectedId == item.id }"
+        :class="{ selected: selectedId == item.id && item.type === selectedType }"
       >
         <div
           class="name pointer"
@@ -56,6 +56,7 @@ defineProps<{
   toggleDiscovery?: (discovery: NewOrUpdatedDiscovery) => void
   passive?: boolean
   selectedId?: number
+  selectedType?: string
 }>()
 </script>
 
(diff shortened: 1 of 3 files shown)


ui/src/store/Queries/discoveryQueries.ts
```diff
@@ -63,7 +64,8 @@ export const useDiscoveryQueries = defineStore('discoveryQueries', () => {
   const { data: tagsByPassiveDiscoveryIdData, execute: tagsByPassiveDiscoveryIdExecute } = useQuery({
     query: TagsByPassiveDiscoveryIdDocument,
     cachePolicy: 'network-only',
-    variables: discoveryId
+    variables: discoveryId,
+    paused: true
```
mikewrosey, line 68: The reason why it calls for both passive & active every time is because these queries share the same variables, which automatically trigger the calls when changed. Adding paused here makes it so that they will only call when explicitly asked.

ui/src/components/Discovery/DiscoveryListCard.vue
```diff
@@ -9,7 +9,7 @@
         v-for="(item, index) in list"
         :key="index"
         class="discovery-name pointer"
-        :class="{ selected: selectedId == item.id }"
+        :class="{ selected: selectedId == item.id && item.type === selectedType }"
```
mikewrosey, line 12: A passive discovery and active discovery might both have the same id, which means both will be highlighted when selected. This makes sure to factor in the type.

freddy61025 (approved):

PR #1539, opened 2023-11-01
LOK-2169 - Switch to citus

jbeuree:
## Description
Switching the DB from postgres to citus/postgres. Switch some sample services (event and notification) to use distributed tables.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2169

## Flagged for review
There's conditional SQL in liquibase for distributing the tables. If the citus extension is not installed, then this isn't run as the functions are specific to the extension and will fail otherwise.
Added some init container dependencies on the DB. Takes longer for citus to start up and be ready, and many containers were spinning and going into backoff. Waiting in the init container may not be the ideal approach, but it does let the test deployments come up much quicker since there's fewer failing containers.

## Checklist


jbeuree:
PR is still in draft, but wanted some folks to start looking it over for comments. In general:
- Tables will need to have the 'distribution' column for citus included in the primary key. The event and notification service have been updated to use a composite key with the tenant id.

deejgregor:
This is going to be really nice to have for development and I'm super looking forward to it!

On a quick browse through, I have a few comments. I'll look more closely later.
1. Can we move citus out of the lokahi Helm chart (make another chart at the same level of lokahi in the charts directory). I'm trying to keep the lokahi chart to be just lokahi and not include dependencies. I'm thinking that it will eventually make sense to introduce a parent chart for testing that can include lokahi, grafana, postgres/citus, etc. for testing.
2. I saw a node port in the citus template -- NodePorts are best avoided ([link](https://oteemo.com/think-nodeport-kubernetes/)) and we should get rid of that completely if we can.
3. Can we keep `postgres` as the `secretKeyRef` `name` instead of changing it to `citus`, or make it configurable? We create that secret in our cloud environment, and it would be nice to not have the name change. Or if we need to make the change, we'll need to coordinate changes before/after this is merged.
4. A few things on your `wait-for-db` `initContainer`: (1) ~~we should deduplicate this and pull it in with a template~~ actually, if we pull this out into a separate Helm chart and install it (with `--wait`) before installing the Lokahi chart, we should be fine on startup... if we do that, we don't need this check or these other comments... (2) I'd also **strongly** suggest not doing `> /dev/null 2>&1` on your `psql` command--if we ever get an odd error, it is impossible to debug without changing the chart or doing things out-of-band. (3) It looks like you are hard-coding the name of the citus pod with `-h citus`. 
5. I see that you added cert-manager -- what needs it? (I see we have it in Tilt already, and I'm kinda surprised it isn't in install-local already... hmmm...)

jbeuree:
> 2\. link

I'll open a separate ticket for (1) and most of (4). I'll keep this one for just switching to citus and the new ticket for the additional refactoring to move charts around.

freddy61025 (approved):
LGTM. Just one minor concert about we didn't change JPA's ID.
It is not related to this PR but I did some simple searching. There is three places using findById. (There is also some delete by Id function calls)
AlertServiceImpl > alertRepository.findById
MonitoringSystemService > systemRepository.findById
NodeService > nodeRepository.findById

cgorantla (approved):
LGTM!
Validated that we can receive events and display them

jbeuree:
> LGTM. Just one minor concert about we didn't change JPA's ID. It is not related to this PR but I did some simple searching. There is three places using findById. (There is also some delete by Id function calls) AlertServiceImpl > alertRepository.findById MonitoringSystemService > systemRepository.findById NodeService > nodeRepository.findById

Yeah, I'll make sure there's a separate bug opened for fixing any cases where queries happen without the tenant ID. Once we enable the db policies, these queries won't work.

PR #1560, opened 2023-11-16
LOK-2206: Move to reactive way of returning result

cgorantla:
## Description
This seems to fix unstability issues in BFF

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2206

## Checklist

### DIFF
--- a/rest-server/src/main/java/org/opennms/horizon/server/service/GrpcNodeService.java
+++ b/rest-server/src/main/java/org/opennms/horizon/server/service/GrpcNodeService.java
@@ -53,7 +53,6 @@ import reactor.core.publisher.Mono;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.stream.Collectors;
 
 @RequiredArgsConstructor
 @GraphQLApi
@@ -120,9 +119,7 @@ public class GrpcNodeService {
 
     @GraphQLQuery
     public Flux<TopNNode> getTopNNode(@GraphQLEnvironment ResolutionEnvironment env, Integer timeRange, TimeRangeUnit timeRangeUnit) {
-        var nodes = client.listNodes(headerUtil.getAuthHeader(env));
-        var topNNodes = nodes.stream().map(nodeDTO -> nodeStatusService.getTopNNode(nodeDTO, timeRange, timeRangeUnit, env))
-            .collect(Collectors.toList());
-        return Flux.fromIterable(topNNodes).flatMap(mono -> mono);
+        return Flux.fromIterable(client.listNodes(headerUtil.getAuthHeader(env)))
+            .flatMap(nodeDTO -> nodeStatusService.getTopNNode(nodeDTO, timeRange, timeRangeUnit, env));
     }
 }
--- a/rest-server/src/main/java/org/opennms/horizon/server/service/GrpcNodeService.java
+++ b/rest-server/src/main/java/org/opennms/horizon/server/service/GrpcNodeService.java
@@ -51,9 +51,9 @@ import org.springframework.stereotype.Service;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 @RequiredArgsConstructor
 @GraphQLApi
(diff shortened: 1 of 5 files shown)


rest-server/src/main/java/org/opennms/horizon/server/service/GrpcNodeService.java
```diff
@@ -120,7 +120,9 @@ public Mono<Boolean> discoveryByNodeIds(List<Long> ids, @GraphQLEnvironment Reso
 
     @GraphQLQuery
     public Flux<TopNNode> getTopNNode(@GraphQLEnvironment ResolutionEnvironment env, Integer timeRange, TimeRangeUnit timeRangeUnit) {
-        // see LOK-2206
-        return Flux.fromIterable(new ArrayList<>());
+        var nodes = client.listNodes(headerUtil.getAuthHeader(env));
+        var topNNodes = nodes.stream().map(nodeDTO -> nodeStatusService.getTopNNode(nodeDTO, timeRange, timeRangeUnit, env))
+            .collect(Collectors.toList());
+        return Flux.fromIterable(topNNodes).flatMap(mono -> mono);
```
vserrago, line 126: This can be simplified to skip using a stream entirely:

```suggestion
        return Flux.fromIterable(client.listNodes(headerUtil.getAuthHeader(env)))
            .flatMap(nodeDTO -> nodeStatusService.getTopNNode(nodeDTO, timeRange, timeRangeUnit, env));
```

rest-server/src/main/java/org/opennms/horizon/server/service/NodeStatusService.java
```diff
@@ -196,46 +202,48 @@ private NodeReachability transformToNodeReachability(long id, TimeSeriesQueryRes
...
-         var topNNode = new TopNNode();
-         topNNode.setNodeLabel(nodeDTO.getNodeLabel());
-         topNNode.setLocation(nodeDTO.getLocation());
-         topNNode.setReachability(nodeReachability.getReachability());
-         topNNode.setAvgResponseTime(nodeResponseTime.getResponseTime());
-         return topNNode;
+    public Mono<TopNNode> getTopNNode(NodeDTO nodeDTO, Integer timeRange, TimeRangeUnit timeRangeUnit, ResolutionEnvironment env) {
+        var nodeReachability = getNodeReachability(nodeDTO, timeRange, timeRangeUnit, env);
+        var nodeResponseTime = getNodeAvgResponseTime(nodeDTO, timeRange, timeRangeUnit, env);
+        var result = nodeReachability.zipWith(nodeResponseTime);
```
vserrago, line 228: Because these items return a container type rather than the underlying type, I think it would make sense to use the explicit type here:

```suggestion
        Mono<NodeReachability> nodeReachability = getNodeReachability(nodeDTO, timeRange, timeRangeUnit, env);
        Mono<NodeResponseTime> nodeResponseTime = getNodeAvgResponseTime(nodeDTO, timeRange, timeRangeUnit, env);
        Mono<Tuple2<NodeReachability, NodeResponseTime>> result = nodeReachability.zipWith(nodeResponseTime);
```

Otherwise it's not immediately clear that we're working with `Mono`s, or even the `Mono<Tuple2<>>`

freddy61025 (approved):
codewise LGTM but i don't see any major changes for the tsdb client, will it really solve the problem ?

vserrago (approved):
I have 2 minor suggestions, but overal LGTM.

I wonder if it would make sense to refactor the code so that instead of querying the TDSB once-per-node, we do it once-per-query. If this change works as expected, maybe we can capture the once-per-query work as a tech-debt ticket?

vserrago (approved):

cgorantla:
> codewise LGTM but i don't see any major changes for the tsdb client, will it really solve the problem ?

This is resolved by removing the blocking (remove future.join) when returning the result

cgorantla:
> I have 2 minor suggestions, but overal LGTM.
> 
> I wonder if it would make sense to refactor the code so that instead of querying the TDSB once-per-node, we do it once-per-query. If this change works as expected, maybe we can capture the once-per-query work as a tech-debt ticket?

Yes, this was also suggested by @deejgregor , created issue https://opennms.atlassian.net/browse/LOK-2216

vserrago (approved):

PR #1599, opened 2023-12-21
LOK-2254: WIP: Add trace ID to graphQL errors shown in the UI

deejgregor:
This is a work-in-progress and is hacky and definitely needs cleaned up.

## Description

Add trace ID (if available) to errors shown in the UI.

Our web server adds a `Server-Timing` header on all responses (including errors) with the traceparent header. This pulls out the trace ID portion of that and adds it to errors shown in the UI as a snackbar (the red popup at the bottom center of the screen).

Here's what the header looks lke, BTW:
```
$ curl --dump-header /dev/stdout -sSf https://auth.onms-fb-dev.dev.nonprod.dataservice.opennms.com/ | grep -i Server-timing
curl: (22) The requested URL returned error: 404
server-timing: traceparent;desc="00-ee7c5ef3834410c3f7a74ebf2146f18c-830c100f808d18f6-01"
```

There a few issues I ran into while doing this (I'm not sure if any of these need to be addressed as part of this PR, I just wanted to mention them as an FYI because they definitely impacted my attempts to do this, in case anyone else is playing around):
1. In the UI development mode, the `Server-Timing` header is never visible in the JavaScript code, even with `Timing-Allow-Origin:  *`. I've found some references to browsers not making this visible when there is a mix of HTTP and HTTPS.
2. Some of the responses don't have the `response` object. I haven't dug into why that happens in villus.
3. When we get multiple errors in short succession, I think each one of them will get their own snackbar and they'll cover each other up, only showing the most recent. (the combination of this and previous item confused the heck out of me for awhile)

Super nice-to-have: if it is possible to add links to the content in the snackbar, we could link directly to the trace in Jaeger in local dev environments (making this feature optional and only enabled where we want it), which would make debugging exceptionally easy.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2254

## Checklist

### DIFF
--- a/ui/src/services/gqlService.ts
+++ b/ui/src/services/gqlService.ts
@@ -21,7 +21,24 @@ const getGqlClient = (kc: Keycloak) => {
 // error notification plugin for specific status codes
 const errorNotificationPlugin = definePlugin(({ afterQuery }) => {
   afterQuery(({ error }, { response }) => {
-    if (!error) return
+    if (!error || !response) {
+      return
+    }
+
+    const serverTiming = response.headers.get('server-timing')
+
+    let traceDetails = ''
+
+    if (serverTiming) {
+      // This is added by our ingress-nginx config. We only care about the trace ID.
+      // traceparent format: https://www.w3.org/TR/trace-context/#traceparent-header
+      const pattern = /traceparent;desc="([0-9a-f]{2})-([0-9a-f]{32})-([0-9a-f]{16})-([0-9a-f]{2})"/
+      const result = pattern.exec(serverTiming)
+
+      if (result && result.length > 1) {
+        traceDetails = ` (Trace ID: ${result[2]})`
+      }
+    }
 
     const errorsMsgs = [
       {
@@ -50,11 +67,11 @@ const errorNotificationPlugin = definePlugin(({ afterQuery }) => {
       },
       {
         err: 409,
-        msg: 'Conflicting entry has occured.'
+        msg: 'Conflicting entry has occurred.'
       },
       {
         err: 500,
-        msg: 'Server error has occured.'
+        msg: 'Server error has occurred.'
       }
     ]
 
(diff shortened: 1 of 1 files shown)


ui/src/services/gqlService.ts
```diff
@@ -21,7 +21,30 @@ const getGqlClient = (kc: Keycloak) => {
 // error notification plugin for specific status codes
 const errorNotificationPlugin = definePlugin(({ afterQuery }) => {
   afterQuery(({ error }, { response }) => {
+    let traceId = ""
```
synqotik, line 24: Consider replacing here through `errorsMsgs` with:

```javascript
if (!error || !response) {
  return
}

const serverTiming = response.headers.get('server-timing')

let traceId = ''

if (serverTiming) {
  const pattern = /traceparent;desc="(..)-([^"-]*)-([^"-]*)-(..)"/
  const result = pattern.exec(serverTiming)

  if (result && result.length > 1) {
    traceId = `Trace ID: ${result[2]}`
  }
}
```

ui/src/services/gqlService.ts
```diff
@@ -21,7 +21,30 @@ const getGqlClient = (kc: Keycloak) => {
...
+//      console.log('token: ', response?.headers.get('token'))
+//      for (const pair of response.headers.entries()) {
+//        console.log(`${pair[0]}: ${pair[1]}`);
+//      }
+    }
     if (!error) return
+    if (!response) return
 
     const errorsMsgs = [
       {
```
synqotik, line 50: Fix `occured` -> `occurred` in a couple cases below.

ui/src/services/gqlService.ts
```diff
@@ -65,17 +88,18 @@ const errorNotificationPlugin = definePlugin(({ afterQuery }) => {
     }
 
     const notificationMsg = () => {
```
synqotik, line 90: Consider replacing this with:

```javascript
const notificationMsg = () => {
  const errorObj = errorsMsgs.filter({ err }) => hasError(err))?.[0]

  return (errorObj?.msg || 'An unknown error has occurred') + ' ' + traceId
}
```

ui/src/services/gqlService.ts
deejgregor: Done!

ui/src/services/gqlService.ts
deejgregor: Ooh, much nicer. Thank you! Done!

ui/src/services/gqlService.ts
deejgregor: Done. Thanks for catching this.

synqotik (approved):
Looks good after changes!

PR #1617, opened 2024-02-05
Auto-formatting

fooker:
## Description
Adds auto-formating using spotless maven integration.
This also adds a check for github-actions and https://pre-commit.com/ integration.

This is split in two commits-groups, first is setting up the formatter, second is applying it to all the current code.


cgorantla (commented):
Awesome. This looks great !

Do we need .mvn/.gitkeep and .vscode/settings.json
Also does License needs to be updated to 2024 ?

cgorantla (approved):
LGTM!

PR #1636, opened 2024-02-13
LOK-2307: Move event conf to shared-lib

cgorantla:
## Description
This will allow us to expose event conf to both alarms and events service.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2307


synqotik (approved):
LGTM

PR #1671, opened 2024-02-27
LOK-2241: Fix Alerts page pagination, fix Spinner reactivity

synqotik:
## Description

Provides a more complete fix to LOK-2241. Alerts page pagination was not working correctly, this should fix pagination and also selection.

`alertsStore` contains alert data for the current page; Alerts page table data now stays in sync with this.

There is an issue where if a user pages quickly, the store and pagination display may be out of sync with what has been fetched by graphql. Added a spinner to alert the user that current data fetch is still in progress.

This PR changes selection (single select or else select/deselect all) to only work per visible items on the current page. This prevents cases where a user may click Clear or Acknowledge, not realizing there are selected items on other pages. We may revisit this behavior.

Also fixed an issue with the Spinner, the `{ isActive }` from the composable was not reactive and therefore the Spinner was never updating its visibility; added `toRefs()`.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2241

### DIFF
--- a/ui/src/components/Common/Spinner.vue
+++ b/ui/src/components/Common/Spinner.vue
@@ -8,7 +8,7 @@
 
 <script setup lang="ts">
 import useSpinner from '@/composables/useSpinner'
-const { isActive } = useSpinner()
+const { isActive } = toRefs(useSpinner())
 </script>
 
 <style scoped lang="scss">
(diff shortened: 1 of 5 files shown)


cgorantla (approved):
LGTM!

Syed-Bilal-Haider-Dataq (commented):
Looking good.

PR #1673, opened 2024-02-27
LOK-2343: Allow user to override minion endpoint port using env

mershad-manesh:
## Description
We want to be able to override the minion endpoint port using an environment variable
 
## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2343


rest-server/src/main/java/org/opennms/horizon/server/utils/MinionDockerZipPackager.java
```diff
@@ -93,13 +93,19 @@ private static byte[] loadDockerCompose(String minionName, String password) thro
             minionEndpoint = "";
         }
 
+        String minionEndpointPort = System.getenv("MINION_ENDPOINT_PORT");
+        if (minionEndpointPort == null) {
+            minionEndpointPort = "443";
```
cgorantla, line 98: may be we can use `1443` to avoid conflicts.

cgorantla (approved):
LGTM!

PR #1677, opened 2024-02-29
LOK-2305: Friendly error message for duplicate discovery name validation

synqotik:
## Description

When a user enters a duplicate Discovery name on the Add Discovery screen, a graphQL server side call is being made which results in a snackbar error message including a trace ID, which is not the best user experience.

Unfortunately, errors from graphQL are handled via a `villus` plugin we added; it's difficult to change that.

However, we can do a client-side check for duplicate name (any names we have in the `store`) and display an error message next to the Discovery Name field, without performing a server side call, as if it were a regular `yup` validation error. Could be a pattern for other similar validation situations.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2305


Syed-Bilal-Haider-Dataq (commented):
I tested the functionality, but I noticed an common issue. When a user updates a discovery name and adds a duplicate name, an error should be thrown. But when I change the tab error message is not set empty. 
DiscoveryListCard.vue. Can you implement it?
 @click="() => { discoveryStore.validationErrors.name && (discoveryStore.validationErrors.name = ''); selectDiscovery(item); }"

synqotik:
> I tested the functionality, but I noticed an common issue. When a user updates a discovery name and adds a duplicate name, an error should be thrown. But when I change the tab error message is not set empty. DiscoveryListCard.vue. Can you implement it? @click="() => { discoveryStore.validationErrors.name && (discoveryStore.validationErrors.name = ''); selectDiscovery(item); }"

Good catch. I fixed it another way - make sure it doesn't check against its own name when doing validation.

synqotik:
> if it is not too much effort, can we extend this for monitoring policy and alert rules

I'm going to open a separate ticket for those, this PR is already getting involved. Can use a similar approach.

cgorantla (approved):
LGTM!

PR #1685, opened 2024-03-04
LOK-2333: Suppress calling flows API until back end is implemented

synqotik:
## Description

Flows is not yet fully implemented. The Node Status / Node Details pages were calling `findExporters` API, causing a snackbar error message to pop up all the time.

This disables that call for now.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2333


cgorantla (approved):
LGTM!

PR #1686, opened 2024-03-04
LOK-2351: Friendly error messages for new Monitoring Policy and Rules

synqotik:
## Description

Add validation for Monitoring Policy and Rule names to prevent blank or duplicate names.

Does a client-side check, if that fails, displays a friendly error message next to then input field and does not send an API request.

Should prevent "snackbar" errors from displaying (they occur when graphql API calls fail).

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2351

### DIFF
--- a/ui/src/components/MonitoringPolicies/MonitoringPoliciesPolicyForm.vue
+++ b/ui/src/components/MonitoringPolicies/MonitoringPoliciesPolicyForm.vue
@@ -45,6 +45,7 @@
           label="New Policy Name"
           v-focus
           data-test="policy-name-input"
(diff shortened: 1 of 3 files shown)


cgorantla (approved):
lgtm!

synqotik:
> Looking good. When I try adding a monitoring policy with a duplicate name, it works well. However, if I use a unique name, the error doesn't go away, and the error message isn't clear at runtime. The same situation happens with rules. Lastly, there's a similar problem with tabs, just like we faced in the Discovery section.

Thanks, I may need to go back and clean that up later.

PR #1698, opened 2024-03-09
DRAFT: Skip UI date tests that fail near DST change

synqotik:
## Description
Skip some UI unit tests for date-related functionality that may fail right before/during a daylight savings time change.

Eventually should make these more robust.

## Jira link(s)
N/A


synqotik:
Closing this out without mergin - only encounter these once or twice a year. We probably need a more robust solution.

PR #1710, opened 2024-03-13
LOK-2376: Monitoring Policies UI, split into new and legacy components

synqotik:
## Description

We are revamping the Monitoring Policies UI with a new design. This moves and renames the existing components to `Legacy` components, but keeps the existing route of `monitoring-policies`.

It adds a new empty `container` which will host the new UI; currently the route is `monitoring-policies-new`. Once the new page has all the existing functionality, we can route to it and remove the Legacy components.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2376


cgorantla (approved):
LGTM!

PR #1712, opened 2024-03-13
LOK-2347: Fix nodeStatus test, vue-router async issue

synqotik:
## Description

`vue-router` 4 is `async`, need to account for this when using a router in unit tests.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2347


cgorantla (approved):
Looks like this passes test cases.

PR #1737, opened 2024-03-20
LOK : 2272 : Prevent duplicate nodes for the same IPAddress for a given location

Junaid-Khan-Nant:
## Description
- When we have multiple Minions, sometimes I see that duplicate nodes being created. This should be prevented.
- Node is always associated with a location so it should have foreign key with monitoring location and a node can have multiple Ip Interfaces.
- We should prevent duplicate IpAddress ( IpInterface) for a given location.
- This can be achieved with unique constraint between IpAddress and location.
- Add a test case to prove that this will work.

## Jira link(s)
https://opennms.atlassian.net/browse/LOK-2272

## Checklist


inventory/main/src/main/resources/db/changelog/hs-0.1.0/tables/ip-interface-add-new-column.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    </changeSet>
+
+
+    <changeSet id="0.1.1-ip_interface" author="Junaid-Khan-Nant">
+        <preConditions onFail="HALT">
+            <changeSetExecuted id="0.1.0-add_column_to_ipinterface" author="Junaid-Khan-Nant" changeLogFile="db/changelog/hs-0.1.0/tables/ip-interface-add-new-column.xml"/>
+        </preConditions>
+        <sql>
+            UPDATE inventory.ip_interface SET location_id = (select distinct(n.monitoring_location_id)  from inventory.node n JOIN inventory.monitoring_location m On n.monitoring_location_id=m.id JOIN inventory.ip_interface i ON i.node_id=n.id);
```
cgorantla, line 29: This won't work, there is no need for schema reference `inventory` here.

We need to test this with and without these changes to determine if this can be applied properly, please do that

inventory/main/src/main/java/org/opennms/horizon/inventory/service/IpInterfaceService.java
```diff
@@ -107,15 +113,21 @@ public void createFromAzureScanResult(
...
+            ipInterface.setNode(node);
+            ipInterface.setTenantId(tenantId);
+            ipInterface.setSnmpPrimary(networkInterfaceItem.getIsPrimary());
+            ipInterface.setIpAddress(InetAddressUtils.getInetAddress(networkInterfaceItem.getIpAddress()));
+            ipInterface.setAzureInterface(azureInterface);
+            ipInterface.setLocation(node.getMonitoringLocation());
+            modelRepo.save(ipInterface);
+        } catch (DataIntegrityViolationException e) {
+            LOG.error("Ip address already exists for a given location :", e.getMessage());
+            throw new UniqueConstraintsException("Ip address already exists for a given location :" + e.getMessage());
```
cgorantla, line 129: As I already updated before, all DataIntegrityVoilationException are not Unique constraint exceptions.

smunir-onms (commented):
LGTM, discussed changes verified.

cgorantla (approved):
LGTM!

PR #1744, opened 2024-03-24
NMS-2412: Upgrade to JNA 5.14.0

synqotik:
## Description

Java Native Access (JNA) added support for Apple Silicon (M1 etc. chips) in 5.6. Latest version is 5.14.0. We are on 4.4.0.

This appears to mostly just affect ICMP on Minion. Running with JNA 4.4.0 on an M1 chip results in no ICMP being sent, but it's fixed with 5.6.0 or higher. This also gets us up to the latest version of JNA.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2412


cgorantla (commented):
LGTM!

PR #1756, opened 2024-03-27
Implement Thread Model for Monitor Execution 

Mohammad-Junid:
## Description
Current monitor execution is undefined and current implementation is unclear. Current API uses futures to delegate problem to implementation.

Implement a thread pool to be shared by all monitors and use this threadpool to execute the monitoring. Therefore alter the monitor interface to not return a future.

Implement a separate thread pool used to process and hand-off monitor results from minion to gateway.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2360

### DIFF
--- a/minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
+++ b/minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
@@ -23,7 +23,6 @@ package org.opennms.horizon.minion.taskset.worker.impl;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
@@ -139,16 +138,8 @@ public class TaskExecutorLocalMonitorServiceImpl implements TaskExecutorLocalSer
             if (monitor != null) {
                 // TBD888: populate host, or stop?
                 MonitoredService monitoredService = configureMonitoredService(taskDefinition);
-                CompletableFuture<ServiceMonitorResponse> future = CompletableFuture.supplyAsync(
-                        () -> {
-                            try {
-                                return monitor.poll(monitoredService, taskDefinition.getConfiguration())
-                                        .get();
-                            } catch (InterruptedException | ExecutionException e) {
-                                throw new RuntimeException(e);
-                            }
-                        },
-                        executor);
+                CompletableFuture<ServiceMonitorResponse> future =
+                        monitor.poll(monitoredService, taskDefinition.getConfiguration());
                 future.whenCompleteAsync(this::handleExecutionComplete, executor);
 
             } else {
--- a/minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
+++ b/minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
@@ -22,8 +22,12 @@
 package org.opennms.horizon.minion.taskset.worker.impl;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.opennms.horizon.minion.plugin.api.MonitoredService;
 import org.opennms.horizon.minion.plugin.api.ServiceMonitor;
 import org.opennms.horizon.minion.plugin.api.ServiceMonitorManager;
@@ -43,7 +47,7 @@ import org.slf4j.LoggerFactory;
  *  problems due to serialization/deserialization.
  */
 public class TaskExecutorLocalMonitorServiceImpl implements TaskExecutorLocalService {
-
+    private ExecutorService executor;
     private static final Logger DEFAULT_LOGGER = LoggerFactory.getLogger(TaskExecutorLocalMonitorServiceImpl.class);
 
     private Logger log = DEFAULT_LOGGER;
@@ -66,6 +70,9 @@ public class TaskExecutorLocalMonitorServiceImpl implements TaskExecutorLocalSer
         this.scheduler = scheduler;
         this.resultProcessor = resultProcessor;
         this.monitorRegistry = monitorRegistry;
+        this.executor = Executors.newCachedThreadPool(new ThreadFactoryBuilder()
+            .setNameFormat("monitor-service-response-handler")
+            .build());
     }
 
     // ========================================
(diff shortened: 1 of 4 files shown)


minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
```diff
@@ -132,9 +138,12 @@ private void executeIteration() {
...
                 MonitoredService monitoredService = configureMonitoredService(taskDefinition);
-                CompletableFuture<ServiceMonitorResponse> future =
-                        monitor.poll(monitoredService, taskDefinition.getConfiguration());
-                future.whenComplete(this::handleExecutionComplete);
+                CompletableFuture.runAsync(
+                        () -> {
+                            monitor.poll(monitoredService, taskDefinition.getConfiguration())
+                                    .whenCompleteAsync(this::handleExecutionComplete, executor);
+                        },
+                        executor);
```
fooker, line 146: I don't think this is sufficient for thread management here.

fooker (changes requested):

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
fooker: I would expect a method returning a `Future` not to block for a long time as part of the contract (similar how methods returning `Optional` should never return `null`). This rises the question on how we do this in blocking impls. We can not have a thread pool per monitor.

I see how this looks counter-intuitive for async monitor impls like the ICMP monitor. But this impl can just block on the future filled in the callback which will just park the task thread.

Outside of the API, almost everything can stay the same but just the call to `poll` has to be wrapped by the thread pool which will transform the blocking method into a `CompletionStage` and we can do the hand-off in the already existing async way.

Mohammad-Junid: Done

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
```diff
@@ -132,9 +139,18 @@ private void executeIteration() {
             if (monitor != null) {
                 // TBD888: populate host, or stop?
                 MonitoredService monitoredService = configureMonitoredService(taskDefinition);
-                CompletableFuture<ServiceMonitorResponse> future =
-                        monitor.poll(monitoredService, taskDefinition.getConfiguration());
-                future.whenComplete(this::handleExecutionComplete);
+                CompletableFuture<ServiceMonitorResponse> future = CompletableFuture.supplyAsync(
+                        () -> {
+                            try {
+                                return monitor.poll(monitoredService, taskDefinition.getConfiguration())
```
cgorantla, line 145: you can simply use `future.whenCompleteAsync` instead of wrapping in another `CompletableFuture.supplyAsync`

cgorantla (changes requested):

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
Mohammad-Junid: Done

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
kashifsardar181, line 1: Its look good based on previous review suggestion

minion/plugins/icmp-plugin/src/test/java/org/opennms/horizon/minion/icmp/IcmpMonitorTest.java
```diff
@@ -82,6 +89,18 @@ public void poll() throws Exception {
         assertTrue(serviceMonitorResponse.getResponseTime() > 0.0);
     }
 
+    @Test
+    public void testPollThroughThreadPoll() {
+        icmpMonitor = getIcmpMonitor(false, false);
+        CompletableFuture<ServiceMonitorResponse> future = icmpMonitor.poll(monitoredService, testConfig);
+        future.whenCompleteAsync(
```
cgorantla, line 96: Not sure why we have the tests in async mode. Tests will always pass since the execution will be sequential here.

Mohammad-Junid: we  have discussed in team by removing async then there is no need of this test case as its already exists "**testError**".

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalMonitorServiceImpl.java
```diff
@@ -66,6 +69,9 @@ public TaskExecutorLocalMonitorServiceImpl(
         this.scheduler = scheduler;
         this.resultProcessor = resultProcessor;
         this.monitorRegistry = monitorRegistry;
+        this.executor = Executors.newCachedThreadPool(new ThreadFactoryBuilder()
```
cgorantla, line 72: Since `TaskExecutorLocalMonitorServiceImpl` is only invoked when a new task is created, this executor should be moved to `TaskExecutorLocalServiceFactoryImpl` and also handle executor shutdown

Mohammad-Junid: The executor moved to  TaskExecutorLocalServiceFactoryImpl

minion/minion-taskset/taskset-worker/src/main/java/org/opennms/horizon/minion/taskset/worker/impl/TaskExecutorLocalServiceFactoryImpl.java
```diff
@@ -102,4 +111,36 @@ public TaskExecutorLocalService create(TaskDefinition taskDefinition) {
...
         }
     }
+
+    private TaskExecutorLocalService createAndGetMonitorServiceInThreadPool(TaskDefinition taskDefinition) {
+        // Create CompletableFuture for asynchronous task creation
+        TaskExecutorLocalService taskExecutorLocalService = null;
+        CompletableFuture<TaskExecutorLocalService> future = CompletableFuture.supplyAsync(
+                () -> {
+                    try {
+                        // Create the service instance
```
cgorantla, line 121: Just pass the executor to `TaskExecutorLocalMonitorServiceImpl` , and keep whatever you had previously in [/TaskExecutorLocalMonitorServiceImpl.java](https://github.com/OpenNMS-Cloud/lokahi/pull/1756/files#diff-94787fcc9391d92e961a5070e1d378b7f3ec5cff44dc7d947f8dbd47efb491ef)

Mohammad-Junid: The changes are done

cgorantla (approved):
LGTM!

fooker (approved):

PR #1782, opened 2024-04-05
LOK-2266   Improve Minion runInPlace.sh to fetch password and certificate without explicitly specifying it

mershad-manesh:


cgorantla (approved):
Validated this. LGTM!

PR #1789, opened 2024-04-10
LOK-2443: Fix issues with Node Status Alerts table

synqotik:
## Description

Fix some issues with Node Status, Recent Alerts table:

- use clickable icon instead of `FeatherExpansionPanel`
- only use `cursor: pointer` for expand icon, not entire row
- fix some alignment and padding
- have expand caret match actual expanded state; retain expanded state unless Refresh clicked

`components.d.ts` was also auto-updated to include some components we added on Monitoring Policies page.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2443


cgorantla (approved):
LGTM!

PR #1831, opened 2024-04-30
FE: Monitoring Policies - Alert Rules - Select Vendor

Shahbaz-dataq:
## Description
For Alert Rules when SNMP Trap is selected, a list of vendors should be displayed and the user must choose a vendor before continuing. The vendor will be used to select event types for the vendor.

This can be implemented as a FeatherSelect or a FeatherAutocomplete; Select is probably ok for now since there are maybe 50 vendors.

Make a graphQL call to listVendors which returns a list. Before display, the list should be converted as follows:

All items should have the first letter capitalized for display. However you need to save the vendor name as-is for the graphQL call

Generic should be the first item displayed

If a Select is used, make sure it scrolls properly, maybe display 6-8 items at a time

Store the selected vendor in the store, or perhaps just in a local variable in the component, it will be used for the next graphQL call

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2472

## Checklist


ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -28,6 +30,9 @@ type TState = {
   numOfAlertsForRule: number
   validationErrors: any,
   alertRuleDrawer: boolean
+  vendors?: string[],
+  formatedVendor?: string[]
```
synqotik, line 34: should be `formattedVendors`

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -28,6 +30,9 @@ type TState = {
...
+  eventDefinations?: AlertEventDefinition[]
```
synqotik, line 35: Should be `eventDefinitions`

synqotik (changes requested):
Looks good! Just a couple small changes.

We will probably need to cache the alert definitions by vendor in a map, but can do that in a subsequent PR.

synqotik (approved):
LGTM!

PR #1833, opened 2024-05-01
FE: Monitoring Policies: Implement number of affected nodes

Shahbaz-dataq:
## Description
Create an item in the monitoringPoliciesStore, a Map from policy id to affected node count.

graphQL call is: getNodesCountByMonitoringPolicy . Unfortunately this only get node count per policy, instead of getting all node counts for all policies.

When Monitoring Policy page loads, after calling getMonitoringPolicies , launch calls (async, in the background) to get count for each policy and store in that Map.

Possible code: 

onMounted(() => store.getMonitoringPolicies().then(() => store.getAllAffectedNodeCounts()))
Monitoring policy table, etc., will query that Map to get the count per policy. Note, this may be “eventually consistent”.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2447

## Checklist


synqotik (changes requested):
Looks good, just a couple of small fixes.

I think `async/await` with `forEach` is ok here. We may or may not need to change to use `Promise.all` + `map` instead, we'll see.

synqotik (approved):
LGTM!

PR #1837, opened 2024-05-06
FE: Monitoring Policies - Alert Rules - Get Trigger and Clear Events

Shahbaz-dataq:
## Description
In Monitoring Policy, Alert Rules panel (for create or edit), after the user has chosen a Vendor for SNMP Trap event type, make a call to alertEventDefsByVendor passing the Vendor and EventType.SNMP_TRAP to get the event definitions. These will populate the Trigger Event and Clear Event dropdowns.

make call to alertEventDefsByVendor. May want to cache the responses in the monitoringPolicyStore, by vendor

Use the name field to populate the Trigger Event dropdown

The Clear Event should be an input text field. When a user selects a Trigger Event, the corresponding Clear Event should be populated.

The Clear Event will be displayed based on the clearKey for the selected Trigger Event, see below on how to parse it. Some clearKeys are empty, so display an empty string in the Clear Event text box.

Save the user selection in the Alert Rule

The clearKey, if present, will usually be in the form:

uei.opennms.org/vendor/cisco/bsnAPCoverageProfileFailed:%dpname%:%nodeid%:%interface%:%parm[#1]%:%parm[#2]%:%parm[#3]%

We want to get the last part of the UEI and display that. Can probably split the string on : to get the UEI, then split by / and pick the last part.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2473

## Checklist


ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -90,7 +81,7 @@ async function getDefaultEventCondition(): Promise<AlertCondition> {
   }
 }
 
-async function getDefaultRule(): Promise<PolicyRule> {
+export async function getDefaultRule(): Promise<PolicyRule> {
```
synqotik, line 84: I think this can be left as non-`export` - we are probably going to remove it or else have it just return `cloneDeep(defaultRule)`. It makes an API using `getDefaultEventCondition()` which we don't need if we just want an empty rule.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -102,10 +93,10 @@ async function getDefaultRule(): Promise<PolicyRule> {
   }
 }
 
-export const  useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore', {
-  state: (): TState =>  ({
-    selectedPolicy: cloneDeep(defaultPolicy),
-    selectedRule: cloneDeep(defaultRule),
+export const useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore', {
+  state: (): TState => ({
+    selectedPolicy: undefined,
```
synqotik, line 98: We changed these to be `cloneDeep(defaultPolicy)` and `cloneDeep(defaultRule)` here and everywhere else, instead of setting to `undefined`, so they are always defined, otherwise we get errors if they a data-bound in Vue components. You should revert back to the `cloneDeep` versions unless you're sure it's ok to set these to `undefined` without any errors.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -130,8 +122,8 @@ export const  useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore'
           this.affectedNodesByMonitoringPolicyCount?.set(policy.id, count ?? 0)
         })
       })
-    
-      this.selectedRule = cloneDeep(defaultRule)
+
+      this.selectedRule = undefined
```
synqotik, line 126: See my comment above, here and anywhere else that `selectedPolicy` or `selectedRule` are set to `undefined`

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -185,7 +178,8 @@ export const  useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore'
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       this.selectedRule!.alertConditions?.map((currentCondition: AlertCondition) => {
         if (currentCondition.id === id) {
-          return { ...currentCondition, ...condition }
+          // this.refactorClearEvent(currentCondition.triggerEvent ?? {})
```
synqotik, line 181: Remove commented-out line.

This method `updateCondition` doesn't end up doing anything, we may eventually remove it or fix it.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -277,23 +284,24 @@ export const  useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore'
...
       await addMonitoringPolicy({ policy })
 
       if (!error.value) {
-        this.selectedPolicy = cloneDeep(defaultPolicy)
-        this.selectedRule = cloneDeep(defaultRule)
+        this.selectedPolicy = undefined
+        this.selectedRule = undefined
         this.validationErrors = {}
-        this.getMonitoringPolicies()
+        await this.getMonitoringPolicies()
```
synqotik, line 304: I don't think we need `await` here, it can run in the background and not block here.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -329,7 +337,7 @@ export const  useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore'
         this.selectedPolicy!.rules?.splice(ruleIndex, 1)
       }
 
-      this.getMonitoringPolicies()
+      await this.getMonitoringPolicies()
```
synqotik, line 340: Don't think we need `await` here.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -340,8 +348,9 @@ export const  useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore'
     async removePolicy() {
       const { deleteMonitoringPolicy } = useMonitoringPoliciesMutations()
       await deleteMonitoringPolicy({ id: this.selectedPolicy?.id })
+      this.monitoringPolicies = []
       this.getMonitoringPolicies()
-      this.selectedRule = cloneDeep(defaultRule)
```
synqotik, line 344: Keep `cloneDeep()` version here and elsewhere.

ui/tests/MonitoringPoliciesLegacy/monitoringPoliciesLegacy.test.ts
```diff
@@ -100,7 +99,7 @@ describe('Monitoring Policies Legacy', () => {
     const store = useMonitoringPoliciesStore()
     const newPolicyBtn = wrapper.get('[data-test="new-policy-btn"]')
 
-    expect(store.selectedPolicy).toStrictEqual(defaultPolicy)
+    expect(store.selectedPolicy).toBeUndefined()
```
synqotik, line 102: Leave this as it was if you revert back to the `cloneDeep()` versions.

ui/src/components/MonitoringPolicies/DetectionMethodComponents/SNMPTrapAlertConditions.vue
```diff
@@ -107,6 +81,12 @@ onMounted(async () => {
   }
   if (monitoringPoliciesStore.selectedRule?.alertConditions && monitoringPoliciesStore.selectedRule?.alertConditions[0]) {
     condition.value = monitoringPoliciesStore.selectedRule?.alertConditions[0]
+    if (monitoringPoliciesStore.selectedRule?.alertConditions[0].triggerEvent?.clearKey) {
+      clearEvent.value = monitoringPoliciesStore.refactorClearEvent(monitoringPoliciesStore.selectedRule?.alertConditions[0].triggerEvent.clearKey)
+    }
+    if (monitoringPoliciesStore.selectedRule?.alertConditions[0].triggerEvent?.uei) {
```
synqotik, line 87: I think remove this and change to 

```typescript
if (monitoringPoliciesStore.selectedRule?.alertConditions[0].triggerEvent?.clearKey) {
  clearEvent.value = monitoringPoliciesStore.refactorClearEvent(monitoringPoliciesStore.selectedRule?.alertConditions[0].triggerEvent.clearKey)
} else {
  clearEvent.value = ''
}
```

In other words, if the `triggerEvent` has a `clearKey`, we parse that and use that as the `clearEvent`, otherwise there's no clear event. At least let's go with this for now.

synqotik (changes requested):
Added comments.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -47,16 +48,6 @@ const defaultPolicy: Policy = {
   rules: []
 }
 
-const defaultRule: PolicyRule = {
```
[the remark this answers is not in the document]
Shahbaz-dataq: As we discussed in meeting. We don't need `defaultRule`

ui/src/store/Views/monitoringPoliciesStore.ts
Shahbaz-dataq: We are using this function. See Line 55, 68 and 138 in `SNMPTrapAlertConditions`

ui/tests/MonitoringPoliciesLegacy/monitoringPoliciesLegacy.test.ts
Shahbaz-dataq: We need this if we want the policy in store to be `undefined`

synqotik (approved):

PR #1878, opened 2024-05-21
FE: Monitoring Policies - Copy Rule

Shahbaz-dataq:
## Description
Implement copy a rule for a Monitoring Policy, when a user clicks on the “copy” icon in the Alert Rules table.

Copy the rule, give it the name Copy of <current rule name>, everything else should be the same.

This is added to the store, user still has to click Save Policy for it to be saved. However it should immediately display in the alert rules table.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2507

## Checklist


faizanjanjua (approved):
Seems Good to me.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -358,6 +358,16 @@ export const useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore',
       delete copiedPolicy.name
       this.displayPolicyForm(copiedPolicy)
     },
+    copyRule(rule: PolicyRule) {
+      const copiedRule = cloneDeep(rule)
+      copiedRule.name = `Copy of ${rule.name}`
+      copiedRule.id = new Date().getTime()
```
synqotik, line 364: Does this work when actually saving? What are the new rule and condition ids after saving?

May want to try this. For new rules for an existing policy (whether added or copied), set `rule.id` and all `condition.id` to `-1`. Then in `saveRule` check for `-1` and call `delete rule.id` or `delete condition.id` before sending to server.

synqotik (changes requested):
Added a comment.

synqotik (approved):
Changes look good, thanks!

PR #1881, opened 2024-05-22
FE: Monitoring Policies - cache number of rules affected by rule id

Shahbaz-dataq:
## Description
When a user tries to delete a rule, it seems there's a long delay while CountAlertByRuleId is called.

Keep a cache of rule id -> number of rules affected in the store.

When user navigates to the Edit policy screen, get all rule ids from that monitoring policy and make CountAlertByRuleId requests for any we don't already have and cache them. Do this in the background, don't block.

Make sure to check the cache when user goes to delete a rule.

Also, when user clicks delete a rule, put up a spinner if any graphQL call has to be made, until the Delete Modal appears.

We may need to revisit this, for example have a graphQL request to get values for multiple rules, but we can try this for now.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2511

## Checklist


ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -358,6 +360,17 @@ export const useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore',
       delete copiedPolicy.name
       this.displayPolicyForm(copiedPolicy)
     },
+    copyRule(rule: PolicyRule) {
+      const copiedRule = cloneDeep(rule)
```
synqotik, line 364: See my comments in #1878 

However, I think the caching part looks good in this PR. We can merge this, then you can rebase #1878 off of `develop` and make sure that copy rule is working correctly.

synqotik (approved):
LGTM - let's merge this, then you can use #1878 to make sure Copy Rule is working.

PR #1894, opened 2024-05-28
FE: Monitoring Policies: When creating a new monitoring policy and adding new alert rule each alert rule replaces the last.

Shahbaz-dataq:
…ding new alert rule each alert rule replaces the last.

## Description
When creating new monitoring policy after saving first alert rule. If you try to add new alert rule and save it will replace the last one. Same thing happening in edit monitoring policy

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2542

## Checklist


ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -301,17 +303,17 @@ export const useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore',
 
       policy.rules = policy.rules?.map((rule) => {
         rule.alertConditions = rule.alertConditions?.map((condition) => {
-          if (!policy.id) delete condition.id // don't send generated ids
-          if (condition.id === -1) delete condition.id
+          if (condition.isNew === true) {
+            delete condition.id // don't send generated ids
+            delete condition.isNew
```
synqotik, line 308: We should `delete condition.isNew` here whether or not it is set to `true`, so it is not sent to the back end at all in any case.

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -301,17 +303,17 @@ export const useMonitoringPoliciesStore = defineStore('monitoringPoliciesStore',
...
           delete condition.clearEvent
           return condition
         })
-        if (!policy.id) {
+        if (rule.isNew === true) {
           delete rule.id // don't send generated ids
-        }
-        if (rule.id === -1) {
-          delete rule.id
+          delete rule.isNew
```
synqotik, line 316: We should `delete rule.isNew` here whether or not it is set to `true` (same reason as above).

ui/src/store/Views/monitoringPoliciesStore.ts
```diff
@@ -68,11 +68,12 @@ function getDefaultThresholdCondition(): ThresholdCondition {
 function getDefaultEventCondition(alertDefs: Array<AlertEventDefinition>): AlertCondition {
   if (alertDefs?.length) {
     return {
-      id: -1,
+      id: new Date().getTime(),
```
synqotik, line 71: Let's create a function in `utils.ts` called something like `createId()`, which just returns `new Date().getTime()` and use that instead. That way we can swap it with a different implementation, plus it's clearer what we are doing.

synqotik (changes requested):
Added a couple comments.

synqotik (approved):
Looks good!

PR #1900, opened 2024-05-31
LOK-2499: Assign Monitored Entity ID to Monitored Service in GRPC interface

smunir-onms:
## Description
Add a field monitoredEntityId to MonitoredServiceDTO in monitoredService.proto containing the monitored entity ID for the particular service. Alter all creations of MonitoredServiceDTO to fill the field.

To determine the correct value, create a static helper method that creates a reproducible monitored entity ID and use it to fill the property and also create the IDs in the MonitoredEntityProvider implementation.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2499

## Checklist


aurang-zeb313 (commented):
LGTM.

inventory/main/src/main/java/org/opennms/horizon/inventory/service/taskset/response/ScannerResponseService.java
```diff
@@ -356,8 +357,11 @@ private MonitoredService createMonitoredService(ServiceResult serviceResult, IpI
                         .setTenantId(tenantId)
                         .build());
 
-        MonitoredServiceDTO newMonitoredService =
-                MonitoredServiceDTO.newBuilder().setTenantId(tenantId).build();
+        MonitoredServiceDTO newMonitoredService = MonitoredServiceDTO.newBuilder()
+                .setTenantId(tenantId)
+                .setMonitoredEntityId(DiscoveryMonitoredEntityProvider.createMonitoredEntityId(
+                        tenantId, ipInterface.getLocation().getId(), serviceResult.getService()))
+                .build();
```
fooker, line 364: I don't think we need to touch this code. It's just a very strange way to transport the tenant into the create method.

inventory/main/src/main/java/org/opennms/horizon/inventory/model/DiscoveryMonitoredEntityProvider.java
```diff
@@ -93,4 +97,9 @@ public List<MonitoredEntity> getMonitoredEntities(final String tenantId, final l
                         .build())
                 .toList();
     }
+
+    public static String createMonitoredEntityId(
+            final String tenantId, final long locationId, final String serviceName) {
+        return Joiner.on(":").join("entity", tenantId, locationId, serviceName);
+    }
```
fooker, line 104: This in not sufficient for multiple reasons:
- The monitored entity ID we need in the DTO must be fully qualified (it must include the provider id prefix - see `MonitoredEntity#getId()`)
- The triple of (tenantId, locationId and serviceName) is not unique: There can be multiple nodes with the same service name, there can be multiple interfaces on the same node with the same service name and in future, there can be multiple services with the name on the same interface - using the service id is sufficient here.

smunir-onms:
> The `monitored_entity_id` field is not set in the mappers and therefore missing in almost all query responses. You have to update the mapper or find a clever way to inject the value somehow (i.e. you can try and add a method to the monitored service entity)

Hi @fooker, the monitoredEntityId is added in the MonitoredService Entity class, and MonitoredServiceMapper Impl is updated accordingly.

inventory/main/src/main/java/org/opennms/horizon/inventory/model/DiscoveryMonitoredEntityProvider.java
smunir-onms: OK, now serviceId is used, so MonitoredEntity#getId() would be like "discovery:entity:{serviceId} is that fine?

inventory/main/src/main/java/org/opennms/horizon/inventory/model/DiscoveryMonitoredEntityProvider.java
```diff
@@ -90,4 +91,8 @@ public List<MonitoredEntity> getMonitoredEntities(final String tenantId, final l
                         .build())
                 .toList();
     }
+
+    public static String createMonitoredEntityId(long serviceId) {
+        return Joiner.on(":").join("entity", serviceId);
+    }
```
fooker, line 97: Why is this required?

inventory/main/src/main/java/org/opennms/horizon/inventory/model/MonitoredService.java
```diff
@@ -56,4 +57,7 @@ public class MonitoredService {
 
     @Column(name = "ip_interface_id", insertable = false, updatable = false)
     private long ipInterfaceId;
+
+    @Transient
+    private String monitoredEntityId;
```
fooker, line 62: Where is this set? I think this needs to be a `getMonitoredEntityId` method which calls the provider to get the return value.

inventory/main/src/main/java/org/opennms/horizon/inventory/service/MonitoredServiceService.java
```diff
@@ -42,18 +42,17 @@ public class MonitoredServiceService {
     private final MonitoredServiceMapper mapper;
     private final IpInterfaceRepository ipInterfaceRepository;
 
-    public MonitoredService createSingle(MonitoredServiceDTO newMonitoredService, IpInterface ipInterface) {
-
-        String tenantId = newMonitoredService.getTenantId();
+    public MonitoredService createSingle(String monitorType, IpInterface ipInterface) {
```
fooker, line 45: Changing the signature is required, but not part of this issue. Is this a git-fuckup somehow?

fooker (changes requested):

inventory/main/src/main/java/org/opennms/horizon/inventory/service/MonitoredServiceService.java
smunir-onms: No, as I added a commit for LOK-2574,  it could be in the new branch/PR whereas I accommodated the changes here.

inventory/main/src/main/java/org/opennms/horizon/inventory/model/DiscoveryMonitoredEntityProvider.java
smunir-onms: This is not required, as a suggestion I asked in the last [comment](https://github.com/OpenNMS-Cloud/lokahi/pull/1900#discussion_r1627570669). Ok, I turn it to just return Long.toString(serviceId);

inventory/main/src/main/java/org/opennms/horizon/inventory/model/MonitoredService.java
smunir-onms: This is the confusion, when I started, changes as per the task description, I was not sure where this would be used, I think something missed to be addressed. Need clarification.

inventory/main/src/main/java/org/opennms/horizon/inventory/monitoring/MonitoredEntity.java
```diff
@@ -71,6 +71,11 @@ public class MonitoredEntity {
      * @return the ID
      */
     public String getId() {
-        return this.source.getProviderId() + ":" + this.entityId;
+
```
fooker, line 74: Remove newline

inventory/main/src/main/java/org/opennms/horizon/inventory/service/MonitoredServiceService.java
fooker: Then at least switch the parameters

inventory/main/src/main/java/org/opennms/horizon/inventory/service/MonitoredServiceService.java
```diff
@@ -42,18 +42,17 @@ public class MonitoredServiceService {
...
 
         Optional<MonitoredService> monitoredServiceOpt =
-                modelRepo.findByTenantIdTypeAndIpInterface(tenantId, newMonitoredService.getMonitorType(), ipInterface);
+                modelRepo.findByTenantIdTypeAndIpInterface(ipInterface.getTenantId(), monitorType, ipInterface);
 
         if (monitoredServiceOpt.isEmpty()) {
 
-            MonitoredService monitoredService = mapper.dtoToModel(newMonitoredService);
```
fooker, line 54: `dtoToModel` can be removed then?

fooker (changes requested):

inventory/main/src/main/java/org/opennms/horizon/inventory/service/MonitoredServiceService.java
smunir-onms: Yes, previously the mapper was transforming just tenantId;
image

fooker: I mean, can it be removed from the Mapper class?

inventory/main/src/main/java/org/opennms/horizon/inventory/monitoring/MonitoredEntity.java
smunir-onms: newline removed.

inventory/main/src/main/java/org/opennms/horizon/inventory/service/MonitoredServiceService.java
smunir-onms: You are right, it can be, so removed.

fooker (approved):

PR #1933, opened 2024-06-11
FE: Node Details: Graphs for Node CPU and Memory

Shahbaz-dataq:
## Description
Implement line graphs for CPU and Memory utilization for the Node Status page.

Use the Graphs/LineGraph.vue component. Will need to determine what the metric names (from the back end / Cortex labels) are for these 2 items. See the Graphs button in the SNMP Interfaces table on Node Status page for example.

Implement the Download and Refresh icons. Download doesn’t need to work yet; Refresh can refresh the graphQL call for the metric data.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2289


ui/src/components/NodeStatus/NodeGraph.vue
```diff
@@ -2,19 +2,58 @@
...
+    monitor: 'SNMP',
+    nodeId: route.params.id as string,
+    instance: '',
+    timeRange: 10,
+    timeRangeUnit: TimeRangeUnit.Minute,
+    ifName: ''
+  }
+})
+const displayEmptyMsgIfNoData = (hasData: boolean) => {
+  if (hasMetricData.value) return
```
Syed-Bilal-Haider-Dataq, line 53: Add an "if" block.
 ```
if (hasMetricData.value) {
return
}
```

File: ui/src/components/NodeStatus/NodeGraph.vue [hunk 2]
```diff
@@ -2,19 +2,58 @@
...
+        <FeatherTooltip title="Refresh" v-slot="{ attrs, on }">
+          <FeatherButton v-bind="attrs" v-on="on" icon="Refresh" class="refresh-icon" @click="graphRef.loadMetricDataAndRenderGraph()">
+            <FeatherIcon :icon="Refresh" />
+          </FeatherButton>
+        </FeatherTooltip>
+      </div>
     </section>
     <section class="node-component-content">
-      <div>Graph goes here</div>
+      <LineGraphForUtilizations ref="graphRef" :graph="graphProps" :type="$props.type" @has-data="displayEmptyMsgIfNoData" />
```
Syed-Bilal-Haider-Dataq, line 19: Break into multiple lines for clear understanding.
 <LineGraphForUtilizations 
ref="graphRef" :graph="graphProps" 
:type="$props.type" 
@has-data="displayEmptyMsgIfNoData" />

ui/src/components/NodeStatus/NodeGraph.vue
```diff
@@ -2,19 +2,58 @@
...
     <section class="node-component-header">
       <h3 data-test="heading" class="node-label">{{ title }}</h3>
+      <div class="actions">
+        <FeatherTooltip title="Download to PDF" v-slot="{ attrs, on }">
+          <FeatherButton v-bind="attrs" v-on="on" icon="Download" class="download-icon" @click="graphRef.onDownload()">
+            <FeatherIcon :icon="DownloadFile" />
+          </FeatherButton>
+        </FeatherTooltip>
+        <FeatherTooltip title="Refresh" v-slot="{ attrs, on }">
+          <FeatherButton v-bind="attrs" v-on="on" icon="Refresh" class="refresh-icon" @click="graphRef.loadMetricDataAndRenderGraph()">
```
Syed-Bilal-Haider-Dataq, line 12: Break into multiple lines for clear understanding.
  <FeatherButton
   v-bind="attrs" v-on="on" 
   icon="Refresh" class="refresh-icon" 
   @click="graphRef.loadMetricDataAndRenderGraph()"
>

ui/src/components/Graphs/LineGraphForUtilizations.vue
```diff
@@ -0,0 +1,227 @@
...
+import { ChartData, ChartOptions, TitleOptions } from 'chart.js';
+import Chart from 'chart.js/auto';
+import { PropType } from 'vue';
+import { getChartGridColor, getColorFromFeatherVar } from '../utils';
+import { downloadCanvas } from './utils';
+
+const emits = defineEmits(['has-data'])
+const graphs = useGraphs()
+const { onThemeChange, isDark } = useTheme()
+let chart: any = {}
```
Syed-Bilal-Haider-Dataq, line 22: Use a chart with the "ref" attribute.
const chart: any = ref({})

ui/src/components/NodeStatus/NodeGraph.vue
Shahbaz-dataq: Done

ui/src/components/NodeStatus/NodeGraph.vue
```diff
@@ -2,19 +2,58 @@
...
+      <LineGraphForUtilizations ref="graphRef" :graph="graphProps" :type="$props.type" @has-data="displayEmptyMsgIfNoData" />
     </section>
   </div>
 </template>
 
 <script lang="ts" setup>
+import { TimeRangeUnit } from '@/types/graphql';
+import DownloadFile from '@featherds/icon/action/DownloadFile'
+import Refresh from '@featherds/icon/navigation/Refresh'
+import { GraphProps } from '@/types/graphs';
```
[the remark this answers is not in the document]
Shahbaz-dataq: Done

ui/src/components/NodeStatus/NodeGraph.vue
Shahbaz-dataq: Done

synqotik (changes requested):
I've thought about this a bit more. Let's use `LineGraph` and make updates to it if needed. I'd like to just standardize using one component for all these type of graphs if possible.

Let's use a real metric instead of dummy data. It does seem like adding dummy data is more involved. If the metrics for CPU and Memory utilization don't work or don't exist yet, then use another one and put a TODO comment to fix it.

I think `LineGraph` already has a download button, so we don't need that in `NodeGraph`. 

You could add that `loadMetricDataAndRenderGraph` to `LineGraph` and either add a refresh button to `LineGraph` or to `NodeGraph` to cal that function.

In `NodeGraph`, add a prop like `metricName`, this will be the metric to query, then pass it in the `graphProps.metrics` to `LineGraph`

`NodeGraph` `type` prop should be same as `LineGraph` `prop.type`, it's `'bytes'` or `'percentage'`.

We can always update CSS styles in `LineGraph` if needed, to be conditional based on props passed in (i.e. whether it's in a modal or not). I think you can set `width` in `NodeStatusTableContent.vue` in the `div` enclosing the `NodeGraph`. It's ok if styling isn't perfect for now.

synqotik (approved):
LGTM! Thanks for changes.

PR #1960, opened 2024-06-27
FE: User Management - add store data

Shahbaz-dataq:
## Description
Data format for a User should be as follows (this will eventually change). Add new file types/users.d.ts.

export interface User {
  id: number
  firstName: string
  lastName: string
  password: string
  email: string
  role: string
  enabled: boolean
}
Add a new store, store/Queries/userStore.ts which contains a list of User objects plus methods to get the list, add a user, update a user by id, delete a user by id.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2620


synqotik (approved):
LGTM!

This will be updated as BFF is completed, but looks good for now.

PR #1957, opened 2024-06-26
LOK-2631:  Add UserManagement 

cgorantla:
## Description

Supports user management for ( add, update, get, resetPassword)

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2631

## Checklist
* [ ] Follows Lōkahi's [development guidelines.](https://github.com/OpenNMS-Cloud/lokahi/wiki/Development-Guidelines)
* [ ] Appropriate reviewer(s) have been selected.
* [ ] Jira issue(s) have been updated to "In Review".
* [ ] Includes [appropriate tests.](https://github.com/OpenNMS-Cloud/lokahi/wiki/Test-Strategy)
* [ ] Documentation has been updated as necessary.
* [ ] Notify devops of changes to the Charts.
* [ ] Notify documentation team of any changes to names of screens or features (affects URLs).

PR #1974, opened 2024-07-03
LOK-2650: FE: Get user list from BFF

Shahbaz-dataq:
## Description
BFF for User Management is being added in LOK-2631: BFF: User Management - Add User
DONE
, see [PR](https://github.com/OpenNMS-Cloud/lokahi/pull/1957)

Add GraphQL to get this user list.

Query name: getAllUsers

query getAllUsers {
  getAllUsers {
    username
    id
    firstName
    lastName
    enabled
    email
    createdTimestamp
  }
}
 

Changes needed to User in users.d.ts:

export interface User {
  id: number
  firstName: string
  lastName: string
  password: string
  email: string
  roles: string[]
  enabled: boolean
  username: string
  createdTimestamp: number
}
 

There should be a mapper from the GraphQL datatype to our internal User interface, they will be similar but not quite the same. In a subsequent issue, we will handle mapping the other way to save/update a user.

Regarding password, this is actually represented in the GraphQL as credentials, but we are only supporting the ‘password' type of credential for now. For this getAllUsers call, we don’t need or want to get or store the passwords, we will do that only when a user is being edited. So password field of User will just be an empty string when returned from getAllUsers.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2650


ui/src/mappers/users.mapper.ts
```diff
@@ -0,0 +1,16 @@
+import { User as ServerUser } from '@/types/graphql'
+import { User as ClientUser } from '@/types/users'
+
+export const mapUsersFromServer = (user: ServerUser): ClientUser => {
```
synqotik, line 4: Nit: rename to `mapUserFromServer` (User should be singular)

ui/src/store/Views/userStore.ts
```diff
@@ -73,9 +79,13 @@ export const useUserStore = defineStore('userStore', {
 
       // On Successfully saving user
 
+      this.clearSelectedUsers()
       this.userEditMode = CreateEditMode.None
 
     },
+    clearSelectedUsers() {
```
synqotik, line 86: Rename to `clearSelectedUser`

synqotik (approved):
LGTM - just a couple small comments

PR #1984, opened 2024-07-09
FE: Service Inventory page - table

atif1019:
Now that the new Service Inventory page has been created, add the Search and Table components as below.

Initially, just use mock data. When BFF is done, we will wire it up with real data.

Columns will be:

Type

Service

Node (this is node alias or node label)

Description

Reachability

Uptime

Latency

Actions

Tabs should be: All Services, Windows Services, Hosted Services, Standalone Services

Add this mock data for all 4 tabs. Ideally use slightly different data (rows) per tab. The Windows, Hosted and Standalone Services tabs table data can just be 2 rows each

No CPU Usage column

Add Search Type, Search input and Filter, like on Node Status page. They don’t need to work right now

Actions can have 1 icon that links user to Node Status page; can use the Info icon like on Network Inventory cards. Note, the mock data should have a node id as well as node name. We don’t display the node id, but will need it for this link.

Sorry about the “K” items in circles, those are comments in the prototype, ignore them


ui/src/types/ServiceInventory.d.ts
```diff
@@ -0,0 +1,14 @@
+export interface ServiceInventoryItem {
+  id: number;
+  type: string;
+  service: string;
+  Node: number;
```
synqotik, line 5: Should be `node` (all lowercase), here and everywhere else it is used.

synqotik (changes requested):
Added a few comments.

synqotik (approved):
LGTM. Thanks for making the changes.

PR #2020, opened 2024-07-30
FE: Allow wildcards for last octet in Discovery IP Address ranges

Shahbaz-dataq:
## Description
Allow more ranges and wildcards for IP addresses in Discoveries. Not quite sure whether the BE fully supports them, but we can implement the following in the UI.

In front end UI, on Discovery page, when setting any IP address range for any of the discovery types, allow wildcards and range for the last octet of an IPv4 address.

However, convert this to a full IP range expression before sending to back end.

If user types 192.168.1.1-254, detect it and convert that to 192.168.1.1-192.168.1.254 before sending to BFF.

If user types 192.168.1.*, detect it and convert to 192.168.1.0-192.168.1.255 before sending to BFF.

This is only valid for the last octet, not any other ones. Note that when we get updated Discovery data from BFF, it will not have the wildcard any longer.

See LOK-2681: We need to allow discovering a whole subnet.  It does allow me to discover 192.168.86.0/24. But it doesn't allow me to discover 192.168.86.1-254.
To Do
 for the original ticket. Opening this separately since there could be a BFF or BE component.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2703


ui/tests/store/Views/discoveryStore.test.ts
```diff
@@ -287,6 +287,17 @@ describe('Discovery Store', () => {
     expect(discoveryStore.saveSelectedDiscovery).toBeCalled()
   })
 
+  it('Test for validateIpRange', () => {
```
synqotik, line 290: Can just call `discoveryStore.validateIpRange()` directly and do an `expect` against the result.

Need to add more test cases here, for example checking for `0` in last or other positions, checking if any octet is greater than 255 that it fails, etc.

synqotik (changes requested):
Requesting some more unit tests.

This is ok for now, if it more or less works; I don't want to hold things up. However, it relies on large, complex regexes that probably still don't fully handle all cases.

We should open a ticket to:

- move this parsing functionality to a common utilities folder, so it can be used anywhere, not just in Discovery
- have a parser that doesn't try to do everything using large regexes, but breaks it down into simpler parts first. Should first detect whether it's a single IP address, a subnet, an IP address with the last octet having a range, or pair of IP addresses (full range) Then check that each octet is valid, etc.,  then return the parsed and reformatted result.
- see if we can find any libraries that do this, or at least parts of it

ui/src/components/Discovery/discovery.constants.ts
```diff
@@ -43,6 +43,10 @@ export const REGEX_EXPRESSIONS = {
   IP: [
     /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(?:-(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))?$/,
     /^(?:\d{1,3}\.){3}\d{1,3}(?:\/(?:[1-2]?[0-9]|3[0-2]))?$|^((?:\d{1,3}\.){3}(?:\d{1,3}-){1,2}\d{1,3})$/,
-    /^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4})|(([0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{0,4})?::(([0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{0,4})?|(^[0-9a-fA-F]{1,4}:){1,7}[0-9a-fA-F]{1,4}\/(1[0-1][0-9]|12[0-8]|[0-9]|[1-9][0-9])$/
+    /^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4})|(([0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{0,4})?::(([0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{0,4})?|(^[0-9a-fA-F]{1,4}:){1,7}[0-9a-fA-F]{1,4}\/(1[0-1][0-9]|12[0-8]|[0-9]|[1-9][0-9])$/,
+    // Regex for whole subnet i.e., 255.255.255.1-254
```
synqotik, line 47: Should be `Regex for partial subnet`

ui/tests/store/Views/discoveryStore.test.ts
```diff
@@ -287,6 +315,119 @@ describe('Discovery Store', () => {
     expect(discoveryStore.saveSelectedDiscovery).toBeCalled()
   })
 
+  it('Test for validateIpRange', async () => {
```
synqotik, line 318: We can refactor this in the future using `it.each`.

synqotik (approved):
LGTM!

PR #2024, opened 2024-07-31
jira/LOK-2707 fix total count of monitored nodes

atif1019:
Description
Initially we didn’t had pagination for network inventory so we were displaying length of nodes list. 
After pagination it is displaying length of nodes per page not the total number of nodes. 

It should display total count from  grapgQL call `BuildNetworkInventory`.

Jira link(s)
(https://opennms.atlassian.net/browse/LOK-2707)


synqotik (approved):
LGTM

PR #2039, opened 2024-08-07
FE: Monitoring Policies, do not display Clear Event field if there is none

atif1019:
## Description
FE: Monitoring Policies, do not display Clear Event field if there is none
 Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2709


synqotik (approved):
LGTM

PR #2040, opened 2024-08-07
FE: On Monitoring policy after enable/disable policy table does not restore to full width

atif1019:
## Description
FE: On Monitoring policy after enable/disable policy table does not restore to full width
## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2725


Shahbaz-dataq (approved):
LGTM

PR #2065, opened 2024-08-21
LOK-2752 : Bring tags and monitored entity ids into alert service when monitored services target is detected

aurang-zeb313:
…monitored entity ids for detected targets of simple discovery service. Also use tags while creating alert in the alert service.

## Description
When simple active discovery targets is detected successfully, need to send those targets tags and monitored entity ids into the alert service.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2752

## Checklist


inventory/main/src/main/java/org/opennms/horizon/inventory/component/TaskSetResultsConsumer.java
```diff
@@ -88,6 +93,19 @@ public void receiveMessage(@Payload byte[] data) {
...
+            List<String> monitoredEntityIds = message.getResultsList().stream()
+                    .filter(TaskResult::hasMonitorResponse)
+                    .map(obj -> obj.getMonitorResponse().getMonitoredEntityId())
+                    .toList();
+
+            if (!monitoredEntityIds.isEmpty()) {
+                simpleMonitorDiscoveryService.handleSimpleActiveDiscoveryTags(monitoredEntityIds, tenantId);
+            } else {
+                log.error("Monitored entity ids do not exist.");
+            }
```
cgorantla, line 107: I think we should push tag updates when we persist discovery and tags not from task set results consumer.  I understand that  monitored_entity_state may not be available at that point.  We should probably try to create monitored_entity_state with down status along with simple_monitored_entity

PR #2071, opened 2024-08-26
LOK-2749 Bring monitored entity id and host into alert service on target detection.

Mohammad-Junid:
## Description
Just like node info , we need monitored entities info in alert service. Need to add monitored entities publisher in  service inventory, when monitored entity info is added, updated or deleted then need to reflect those changes  in the alert service. For implementation see NodeKafkaProducer in the inventory service. Create table with name of  simple_monitored_entity in the alert service, this table will have two columns monitored entity id and host.

## Jira link(s)
- https://opennms.atlassian.net/browse/LOK-2749

## Checklist


alert/src/main/java/org/opennms/horizon/alertservice/grpc/AlertGrpcService.java
```diff
@@ -461,6 +465,10 @@ private void getFilter(
                 for (Node node : nodes) {
                     nodeIds.add(String.valueOf(node.getId()));
```
aurang-zeb313, line 466: here nodeIds is used for both node and monitored entityId?

aurang-zeb313 (changes requested):

alert/src/main/java/org/opennms/horizon/alertservice/grpc/AlertGrpcService.java
Mohammad-Junid: Yes . Its just prepare the list of string ids and then filter the column "managedObjectInstance"  in which we have the node id as well as entity id  in string format.

aurang-zeb313 (approved):

shared-lib/alert/src/main/proto/alerts.proto
```diff
@@ -45,6 +45,7 @@ enum ManagedObjectType {
   SNMP_INTERFACE = 3;
   SNMP_INTERFACE_LINK = 4;
   INTERFACE = 5;
+  MONITERED_ENTIRY = 6;
```
cgorantla, line 48: Do we really need to update ManagedObjectType for this ?  We probably want to get rid of this.

cgorantla (commented):
Please see my comment, lgtm otherwise

